PR #3, opened 2021-10-13
Prep api-client-core for React bindings

thegedge:
There are two major things in here:

1. Added the `Select` type, which can reduce a schema based on a selection. This helps us (generically) construct the result types for our React hooks based on the selection.
2. Factored out all of the operation building bits of the operation runners, which we use to help us build a query to pass to urql.

There's some other small stuff in here, like `s/ModelManager/AnyModelManager/g` and exporting a function to construct a Gadget-typed error for an error payload.


packages/api-client-core/spec/types.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+type _TestDefaultSelectionIsValidSelection = AssertTrue<Has<typeof TestDefaultSelection, TestSchemaSelection>>;
+
+type _DefaultSelectionWithSelection = AssertTrue<
+  IsExact<DefaultSelection<TestSchemaSelection, { select: { num: true } }, typeof TestDefaultSelection>, { num: true }>
+>;
+
+type _DefaultSelectionWithoutSelection = AssertTrue<
+  IsExact<DefaultSelection<TestSchemaSelection, Selectable<TestSchemaSelection>, typeof TestDefaultSelection>, typeof TestDefaultSelection>
+>;
```
airhorns, line 42: 👏

packages/api-client-core/src/types.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export type Select<Schema, Selection extends FieldSelection | null | undefined> = Selection extends null | undefined
+  ? never
+  : FilterNever<{
+      [Key in keyof Selection & keyof Schema]: Selection[Key] extends true
+        ? Schema[Key]
+        : Selection[Key] extends FieldSelection
+        ? Select<Schema[Key], Selection[Key]>
+        : never;
+    }>;
```
airhorns, line 64: These types are so much terser than what we're doing now, 🤘

packages/api-client-core/src/GadgetRecordList.ts
```diff
@@ -2,8 +2,8 @@
 /* eslint-disable @typescript-eslint/require-await */
 import { GadgetRecord } from "./GadgetRecord";
 import { InternalModelManager } from "./InternalModelManager";
-import { ModelManager } from "./ModelManager";
-import type { PaginationOptions } from "./operationRunners";
```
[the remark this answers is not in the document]
thegedge: 👍 will change.

I think auto import always assumes you want to use something as a value. Weird, since I don't think `PaginationOptions` ever could, but I guess there's always the possibility of an import having side effects 😬

packages/api-client-core/src/support.ts
```diff
@@ -112,56 +113,57 @@ export const assertOperationSuccess = (response: OperationResult<any>, dataPath:
...
+    return new GadgetNonUniqueDataError(error.message);
+  } else {
+    return new GadgetOperationError(error.message, error.code);
+  }
+};
+
 export const assertMutationSuccess = (response: OperationResult<any>, dataPath: string[]) => {
   const operationResponse = assertOperationSuccess(response, dataPath);
 
-  if (!operationResponse.success) {
```
thegedge, line 118: Here was my big oopsie that would have broke things.

thegedge:
We're green in gadget 🎉

PR #4, opened 2021-10-18
Export everything from GadgetFunctions in api-client-core

thegedge:
I forgot to do this in https://github.com/gadget-inc/js-clients/pull/3. I planned on moving these from my react branch to `api-client-core`. I'll keep my copy in the react client for now, and come back later to remove it.


thegedge:
Nope! There's no urgency. I'm just going to include a copy in the React client for now (it originated from there anyways) and I'll come back later to remove it, once we deploy a new version of the core.

PR #5, opened 2021-10-18
Add React hooks to work with a Gadget API.

thegedge:
A lot of the complexity in here feeds a few requirements:

- Should be as easy to use as possible
- Any `select` option should be restricted to the schema of the relevant model
- Any `select` option should also limit the type of the returned record, such that one should not be able to select fields that were not part of `select`.

With that in place, we unfortunately took on the burden of higher TypeScript complexity. Some tricks are in place to ensure things work just like the regular imperative client we export for every app.

## For reviewers

Two things left to do, but I wanted to get the main part of this up for review. Those two things are:
1. Add a little README.md to the react client
2. Figure out a nice way to test this

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.1.2",
+  "version": "0.2.0",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",


packages/react/src/GadgetFunctions.ts
```diff
@@ -0,0 +1,73 @@
+import { GadgetRecord, GadgetRecordList, LimitToKnownKeys, VariableOptions } from "@gadgetinc/api-client-core";
```
thegedge, line 1: This file was added to api-client-core, but not properly exported from its index. I'm just going to put it in here for now, and come back later to remove it, once we release a version of api-client-core that properly exports its own copy of this file.

packages/react/src/useAction.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * React hook to fetch a Gadget record using the `findByXYZ` method of a given manager.
+ *
+ * @param action `findByXYZ` function from a Gadget manager that will be used
+ * @param params input params to the Gadget action
+ * @param options options for selecting the fields in the result
+ */
+export const useAction = <
+  OptionsT extends OptionsType, // currently necessary for Options to be a narrow type (e.g., `true` instead of `boolean`)
```
thegedge, line 25: Comment says it all. Without this, we don't get narrowed types for the options. In other words:

```
useFindOne(client.widget, { select: { name: true } });
```

Will give us `name: boolean` instead of `name: true`. I don't completely understand the why behind this, but it appears to work! Hoping one day we can get rid of it.

packages/react/src/useFindBy.ts
```diff
@@ -0,0 +1,68 @@
...
+export const useFindBy = <
+  OptionsT extends OptionsType, // currently necessary for Options to be a narrow type (e.g., `true` instead of `boolean`)
+  F extends FindOneFunction<OptionsT, any, any, any>,
+  Options extends F["optionsType"]
+>(
+  finder: F,
+  value: string,
+  options?: LimitToKnownKeys<Options, F["optionsType"]>
+): UseQueryResponse<GadgetRecord<Select<Exclude<F["schemaType"], null | undefined>, Options["select"]>>> => {
+  const memoizedOptions = useStructuralMemo(options);
```
thegedge, line 32: Without this, and given how options are typically written inline as an argument, we often would have objects with new identities every time. This would mean the deps of the `useMemo` directly below would be different, creating a new plan. Similarly, the `plan.variables` object would likely be different, causing us to run a new `useQuery`.

packages/react/src/useFindMany.ts
```diff
@@ -0,0 +1,54 @@
...
+      manager.findMany.operationName,
+      manager.findMany.defaultSelection,
+      manager.findMany.modelApiIdentifier,
+      memoizedOptions
+    );
+  }, [manager, memoizedOptions]);
+
+  const context = useMemo<Partial<OperationContext>>(
+    () => ({
+      additionalTypenames: [capitalize(manager.findMany.modelApiIdentifier)],
```
thegedge, line 34: I decided to just go this route, but I could also include the GraphQL type name as another piece of metadata.

packages/react/src/useFindOne.ts
```diff
@@ -0,0 +1,49 @@
...
+ * React hook to fetch a Gadget record using the `findOne` method of a given manager.
+ *
+ * @param manager Gadget model manager to use
+ * @param id id of the record to fetch
+ * @param options options for selecting the fields in the result
+ */
+export const useFindOne = <
+  OptionsT extends OptionsType, // currently necessary for Options to be a narrow type (e.g., `true` instead of `boolean`)
+  F extends FindOneFunction<OptionsT, any, any, any>,
+  Options extends F["optionsType"]
```
thegedge, line 18: As you can see, I'm actually not exposing any urql options right now. I think we'd want to do that, eh? Or should we expose our own layer on top of it? I'm inclined to do the former, as I think we're pretty committed to the urql ecosystem.

packages/react/README.md
```diff
@@ -0,0 +1,72 @@
+<div align="center">
+  <p>
+    <img alt="Gadget logo" src="https://raw.githubusercontent.com/gadget-inc/js-clients/main/docs/assets/gadget-logo.png" />
```
thegedge, line 3: This won't work until we merge. I didn't want to do a relative path (`../../`), but maybe that'd be okay?

packages/react/src/useFindOne.ts
airhorns: Yeah, I think we should proxy through any urql options the developer wants to set so we don't take away power unnecessarily, so `Options extends F["optionsType"] & Omit<UseQueryArgs<never, never>, 'query' | 'variables'>` or something like that. There's not actually that many other options (see https://github.com/FormidableLabs/urql/blob/main/packages/react-urql/src/hooks/useQuery.ts#L22-L28) so we could even just inline those types here if that omit business seems too complicated.

packages/react/README.md
```diff
@@ -0,0 +1,72 @@
...
+the same [features](https://github.com/FormidableLabs/urql/blob/main/README.md#-features) as urql.
+
+## Installation
+
+```
+yarn add @gadgetinc/react react
+# or
+npm install --save @gadgetinc/react react
+```
+
```
airhorns, line 35: I think it'd be good to explicitly call out that this is a companion package to your specific api client package, you know? Maybe something like

## Installation

This package is a companion package to the JavaScript client package generated for your Gadget app, so you must install the JS client for your app, and then install this package. 

To install the JS client for your app, you must set up the Gadget NPM registry, and then install the client:

```bash
npm config set @gadget-client:registry https://registry.gadget.dev/npm
yarn add @gadget-client/my-app-slug
# or
npm install @gadget-client/product-tagger-8000
```

Full installation instructions can be found in the [Gadget docs](https://docs.gadget.dev/) at `https://docs.gadget.dev/api/<my-app-slug>/installing`.

Once you have your JS client installed, you can install the React hooks library with yarn or npm:

```bash
yarn add @gadgetinc/react react
# or
npm install --save @gadgetinc/react react
```

packages/react/README.md
```diff
@@ -0,0 +1,72 @@
...
+## Example usage
+
+```typescript
+import { useAction, useFindMany } from "@gadgetinc/react";
+import React from "react";
+import { Client } from "@gadget-client/my-gadget-app";
+
+const [_, deleteWidget] = useAction(Client.widget.delete);
```
airhorns, line 43: I think you need an instance of the api, no? Or can you really just use the prototype like this? Also, I feel like it might freak people out to see hooks used at the top level of a module as opposed to inside a React component, can I suggest adding a wrapper component?

packages/react/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+// if id or _all weren't fields on a widget, this would be a type error
+const [result, refresh] = useFindMany(Client.widget, {
```
airhorns, line 46: Can you document how you'd set up the urql context with the right auth and whatnot?

packages/react/src/useStructuralMemo.ts
```diff
@@ -0,0 +1,18 @@
+import deepEqual from "deep-equal";
+import { useRef } from "react";
+
+/**
+ * Memoize and ensure a stable identity on a given value as long as it remains the same, structurally.
+ */
+export const useStructuralMemo = <T>(value: T): T | undefined => {
```
airhorns, line 7: Any chance we could import this or something similar from urql? I feel like they probably need to so something similar internally such that we could avoid shipping extra JS kbs to the client

DEVELOPMENT.md
```diff
@@ -0,0 +1,39 @@
...
+```json
+{
+  "dependencies": {
+    "@gadgetinc/react": "0.1.2"
+  },
+  "resolutions": {
+    "@gadgetinc/react": "https://codeload.github.com/gadget-inc/js-clients/tar.gz/@gadgetinc/react-v0.1.2-gitpkg-0123456"
+  }
+}
+```
```
airhorns, line 39: 🙏

packages/react/src/useAction.ts
thegedge: Yep, happy to change the name to something more useful like that 🚀

DEVELOPMENT.md
kira-bruneau: 🙏

kira-bruneau (approved):
Other than what @airhorns mentioned, this LGTM 🙌

thegedge:
Put this back to a draft 😅

PR #6, opened 2021-10-21
Update runners and function types for actions.

thegedge:
Function type definitions are updated to add bulk action types, namespacing, and a new piece of metadata: `modelSelectionField`.

The model selection field allows us to distinguish between the model's API identifier, and the field in GraphQL that applies to the `select` option given to our operations. With bulk actions, for example, the model API identifier may be "widget", but the GraphQL response is served under a field named "widgets". The model API identifier is still useful for bulk actions, because we can fetch hydrations from `gadgetMeta` with it.

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.2.2",
+  "version": "0.3.0",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",


packages/api-client-core/src/operationRunners.ts
```diff
@@ -82,36 +83,44 @@ export interface ActionRunner {
...
   const response = await modelManager.connection.currentClient.mutation(plan.query, plan.variables).toPromise();
 
   // pass bulk responses through without any assertions since we can have a success: false response but still want
   // to process it in a similar fashion since some of the records could have been processed
   const dataPath = namespace ? [namespace, operation] : [operation];
   const mutationResult = isBulkAction ? get(response.data, dataPath) : assertMutationSuccess(response, dataPath);
 
+  // Currently, delete actions have a null selection. We do an early return for this because `hydrateRecordArray` will fail
+  // if there's nothing at `mutationResult[modelSelectionField]`, but the caller isn't expecting a return (void).
+  if (defaultSelection == null) {
```
thegedge, line 115: Not a fan of this, but it gets the job done. Tested and verified on a branch in the gadget repo.

packages/api-client-core/src/operationBuilders.ts
```diff
@@ -115,7 +116,7 @@ export const actionOperation = (
 
   const selection = options?.select || defaultSelection;
   if (selection) {
-    actionOperation.fields!.push({ [modelApiIdentifier]: fieldSelectionToGQLBuilderFields(selection) });
+    actionOperation.fields!.push({ [modelSelectionField]: fieldSelectionToGQLBuilderFields(selection) });
```
thegedge, line 119: Important bit is here for bulk actions: as pointed out in the PR description, the field would be "widgets", but we need to get hydrations (below) for "widget".

packages/api-client-core/src/index.ts
```diff
@@ -13,5 +14,3 @@ export * from "./operationBuilders";
 export * from "./operationRunners";
 export * from "./support";
 export * from "./types";
-
-export type { AsyncRecord, AsyncRecordList, FindOneFunction, FindManyFunction, ActionFunction } from "./GadgetFunctions";
```
airhorns, line 17: I guess this still builds ok eh? I thought I had to do this to get compiling with `isolatedModules` on, but if CI is 🍏 I think we're good! 👍

packages/api-client-core/src/operationRunners.ts
airhorns: Makes sense! We'll probably have to evolve this over time but I dunno how exactly so send it.

packages/api-client-core/src/operationRunners.ts
```diff
@@ -82,36 +83,44 @@ export interface ActionRunner {
...
-  ): Promise<Shape extends void ? void : GadgetRecordList<Shape>>;
+  ): Promise<Shape extends void ? void : GadgetRecord<Shape>[]>;
 }
 
 export const actionRunner: ActionRunner = async <Shape = any>(
   modelManager: { connection: GadgetConnection },
   operation: string,
   defaultSelection: FieldSelection | null,
   modelApiIdentifier: string,
+  modelSelectionField: string,
```
airhorns, line 99: Is this signature change gonna require a rebuild of all the built JS clients when we ship this?

packages/api-client-core/src/operationRunners.ts
```diff
@@ -82,36 +83,44 @@ export interface ActionRunner {
     operation: string,
     defaultSelection: FieldSelection | null,
     modelApiIdentifier: string,
+    modelSelectionField: string,
     isBulkAction: true,
     variables: VariableOptions,
     options?: SelectionOptions | null,
     namespace?: string | null
-  ): Promise<Shape extends void ? void : GadgetRecordList<Shape>>;
+  ): Promise<Shape extends void ? void : GadgetRecord<Shape>[]>;
```
airhorns, line 91: Whoopth! 👼

packages/api-client-core/src/index.ts
thegedge: Hmm, yeah, that _could_ be problematic, but I haven't had any issues. Maybe `export *` is fine?

PR #10, opened 2021-10-29
Bump api-client-core dep in react to ^0.4

thegedge:
The new version of `api-client-core` includes improvements to the `Select` type, which makes the result types for all of our hooks be properly typed when selecting an array/nullable type. This unfortunate led to TS complaining about:

> Type instantiation is excessively deep and possibly infinite

on the return types. Adding an additional generic parameter for the schema seems to have resolved this issue. I can't explain why, unfortunately.

One question: should we consider a different version specifier for now, that is a little more loose? I was thinking `~0.3` (>=0.3, <1) instead of `^0.3` (>= 0.3, <0.4).

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -18,7 +18,7 @@
     "prerelease": "gitpkg publish"
   },
   "dependencies": {
-    "@gadgetinc/api-client-core": "^0.3.0",
+    "@gadgetinc/api-client-core": "^0.4",
     "deep-equal": "^2.0.5",
     "urql": "^2.0.5"
   },
--- a/packages/react/src/useAction.ts
+++ b/packages/react/src/useAction.ts
@@ -48,7 +48,8 @@ import { useStructuralMemo } from "./useStructuralMemo";
  */
 export const useAction = <
   GivenOptions extends OptionsType, // currently necessary for Options to be a narrow type (e.g., `true` instead of `boolean`)
-  F extends ActionFunction<GivenOptions, any, any, any, any>,
+  SchemaT,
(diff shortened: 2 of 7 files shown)


thegedge:
So I've tested it out on a query, and we'll see the type instantiation depth error if our `select` is too deep. For example, this one gives the error in the kitchen sink test app:

```js
  const [result, _refresh] = useFindMany(apiKeyClient.widget, {
    select: {
      id: true,
      name: true,
      gizmos: {
        edges: {
          node: {
            id: true,
            name: true,
            widget: {
              section: {
                title: true,
                _all: true,
                sprocket: {
                  _all: true
                }
              }
            }
          },
        },
      },
    },
  });
```

kira-bruneau (approved):
> So I've tested it out on a query, and we'll see the type instantiation depth error if our select is too deep. For example, this one gives the error in the kitchen sink test app:

Ah ok. It seems like it would be pretty common to have a query with that many layers, but I think it would be ok to ship this, and continue working on a better fix in another PR.

> One question: should we consider a different version specifier for now, that is a little more loose? I was thinking ~0.3 (>=0.3, <1) instead of ^0.3 (>= 0.3, <0.4).

While it's still in initial development with 0.x, couldn't it introduce breaking changes between minor versions?

thegedge:
> While it's still in initial development with 0.x, couldn't it introduce breaking changes between minor versions?

That's how some people try to do it, but <1.0.0 is unstable across the board. Semantic versioning [says](https://semver.org/#spec-item-4):

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Given that, I think it would be a little easier for us to work with anything <1, embrace the potential for breakage while we get it all right. Once we're happy with the end result, we move to 1.0.0, at which point semantic versioning should kick in.

airhorns (approved):
Makes sense to me, let's see if we can figure out a way to allow much deeper types though! Maybe that is why Prisma did the named type generation stuff that inspired our original approach?

thegedge:
>  Makes sense to me, let's see if we can figure out a way to allow much deeper types though! Maybe that is why Prisma did the named type generation stuff that inspired our original approach?

Maybe!

We can always recommend `// @ts-ignore` for now, since things will be fine at runtime, and hopefully get all of these types nice and happy down the road. I definitely know how to redo `Select` to get us some more depth, but at the end of the day I think we may always hit this limit if we go the "single, recursive type" route :(

PR #11, opened 2021-11-10
[react] Typed default selection support

airhorns:
When working on some example apps I went to use a plain ole `useFindMany(api.widget)` without passing a `select` option and found there to be no typed return value! The implementation supports this just fine, but we needed one little extra step to default the `select` option at type time as well as at runtime. This adds that, as well as a bunch of type-time tests that assert it works. To type test in this package I added some example clients that we yarn install from production Gadget apps which feels a little weird but I think buys us a lot of confidence so I think we aught to do it.

### DIFF
--- a/packages/react/spec/useBulkAction.spec.ts
+++ b/packages/react/spec/useBulkAction.spec.ts
@@ -1,0 +1,48 @@
+import { GadgetRecord } from "@gadgetinc/api-client-core";
+import { CombinedError } from "@urql/core";
+import { assert, IsExact } from "conditional-type-checks";
+import { useBulkAction } from "../src";
+import { bulkExampleApi } from "./apis";
+
+// these functions are typechecked but never run to avoid actually making API calls
+const TestUseBulkActionCanRunActionsWithVariables = () => {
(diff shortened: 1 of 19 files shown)


packages/react/spec/useBulkAction.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { GadgetRecord } from "@gadgetinc/api-client-core";
+import { CombinedError } from "@urql/core";
+import { assert, IsExact } from "conditional-type-checks";
+import { useBulkAction } from "../src";
+import { bulkExampleApi } from "./apis";
+
+// these functions are typechecked but never run to avoid actually making API calls
+const TestUseBulkActionCanRunActionsWithVariables = () => {
```
thegedge, line 8: Thanks for writing all of these tests. It was on my backlog to tackle this month, alongside routes work 🙇

thegedge (approved):
Nice catch! This definitely crossed my mind a couple of times while working on the React bindings, but I think I kept forgetting about it 🙈

packages/react/spec/useBulkAction.spec.ts
airhorns: No problemo!

PR #14, opened 2021-11-21
Switch to the multipart fetch exchange for urql to support file uploads

airhorns:
This switches us over to using the built in bits of urql that support file uploads over graphql. We're gonna use https://github.com/jaydenseric/graphql-upload on the backend which `@urql/exchange-multipart-fetch` works with well! This exchange is a drop in replacement for the normal fetch exchange if you're not uploading files also, so we can roll it out ahead of time. 

All the integration testing of this thing is gonna happen Gadget side.

See https://formidable.com/open-source/urql/docs/api/multipart-fetch-exchange/ for more info on it.

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -1,4 +1,5 @@
-import { Client, dedupExchange, fetchExchange, subscriptionExchange } from "@urql/core";
+import { Client, dedupExchange, subscriptionExchange } from "@urql/core";
+import { multipartFetchExchange } from "@urql/exchange-multipart-fetch";
 import fetch from "cross-fetch";
 import { ExecutionResult } from "graphql";
 import {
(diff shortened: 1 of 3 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -257,7 +258,7 @@ export class GadgetConnection {
       fetch: this.fetch,
       exchanges: [
         dedupExchange,
-        fetchExchange,
+        multipartFetchExchange,
```
thegedge, line 261: Any performance issues to using this over `fetchExchange`? If yes, perhaps down the road we could provide a set of default exchanges from the generated API clients, based on whether or not there are file fields.

thegedge (approved):
Well that was easy! 🎉

packages/api-client-core/src/GadgetConnection.ts
airhorns: The multipart fetch exchange does exactly the same thing as the fetchExchange when there's no files involved, it only does multipart stuff if there are files, so I don't think so! Woop woop!

PR #15, opened 2021-12-03
Better handle connected states

thegedge:
Previously, there was a chance that if the connection was established very quickly, we may have missed the "connected" event from the websocket before we created the listener in `connectionOpen`. This would result in a false timeout from our own `setTimeout`, which always rejects.

To handle this scenario, we ensure a connected listener is given in the construction of the subscription client that tracks the status.

**TODO:**
- [ ] Verify this scenario can occur, and try to reproduce it

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.4.6",
+  "version": "0.4.7",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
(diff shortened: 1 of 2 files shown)


airhorns (approved):
Makes sense to me! Not sure if this will fix the timeout issue we're seeing but good to eliminate this as a potential issue regardless. 👍

thegedge:
In some local testing, I'm finding this doesn't really seem to help much. I'm going to see if there's something a bit better we can do.

thegedge:
☝️ We've decided to listen for the error message when connecting, but no need to check for the connected state.

airhorns:
LGTM!

PR #16, opened 2022-01-11
Add a configurable ws timeout

angelini:
Set the WS connection timeout - the time we wait for the `onConnect` callback to run at 10s. (2x the previous value).

And allow users of the client to pass a different timeout in to avoid needing to change the client when we update this value in the future.


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -18,6 +18,7 @@ import { traceFunction } from "./support";
 export type TransactionRun<T> = (transaction: GadgetTransaction) => Promise<T>;
 export interface GadgetSubscriptionClientOptions extends Partial<SubscriptionClientOptions> {
   urlParams?: Record<string, string | null | undefined>;
+  connectionTimeout?: number;
```
thegedge, line 21: I think we should suffix with `Ms`, like you did with the timeout arg below, since this is "external" facing (in other words, far more important to be clear about unit here).

thegedge (approved):

airhorns (approved):

PR #17, opened 2022-01-20
Use urql's cache exchange when running in the browser by default

airhorns:
urql, the graphql react client library we use, has a light weight caching feature that reloads queries shown on screen when any mutation that might have changed the data runs. See https://formidable.com/open-source/urql/docs/basics/document-caching/ . You get it out of the box when using urql normally, but right now when using Gadget's react hooks, it is off by accident. This is because we ask users to pass in our highly-configured urql client, and we don't include the exchange. I didn't include it initially because it doesn't make much sense in a server-to-server communication context where you expect and API client to make API calls and not cache data that happened to have been fetched recently, but in the browser, urql has chosen this as default behaviour and I agree it makes sense.

Fixes sc-3190

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -1,4 +1,4 @@
-import { Client, dedupExchange, subscriptionExchange } from "@urql/core";
+import { cacheExchange, Client, dedupExchange, subscriptionExchange } from "@urql/core";
 import { multipartFetchExchange } from "@urql/exchange-multipart-fetch";
 import fetch from "cross-fetch";
 import { ExecutionResult } from "graphql";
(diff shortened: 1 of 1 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -258,26 +258,30 @@ export class GadgetConnection {
...
-            };
-          },
-        }),
-      ],
-    });
+    const exchanges = [dedupExchange];
+
+    // apply urql's default caching behaviour when client side (but skip it server side)
+    if (typeof window != "undefined") {
+      exchanges.push(cacheExchange);
```
thegedge, line 265: So reading up on this a while back, I think we need to update everything to always add `__typename` and `id` to the query to make this valuable?

airhorns: We are already woop woop! https://github.com/gadget-inc/js-clients/blob/2afa56e477231acdd1c5eb96e94cb978e9f7563b/packages/api-client-core/src/FieldSelection.ts#L20 is a thing which the client sets most of the time.

thegedge: Ah yes, I see it defaults to `false`, but we generally use `true`. How about the id? Or do you think it's a "use id to make it fast and good, but your choice to not select it?" kind of dealio?

thegedge (approved):

packages/api-client-core/src/GadgetConnection.ts
airhorns: I think it would be wise to fetch it, but, document caching doesn't actually need it so I don't think we should force it. If we add graph-cache support then it'd be a thing at that point for sure.

thegedge: Oh right, my bad. I was thinking this was graph cache 😅

PR #18, opened 2022-02-02
Retry WS on connection failure

angelini:
As we continue to investigate the WS connection issue, here is what our understanding of the most common occurrence looks like:

- A `subscriptionClient` is created within the sandbox
- It attempts to make a connection to the API
- After 10 seconds a userland timeout within the client is triggered - this tears down the client but does not affect the ongoing request
- After 12 seconds nothing else happens - meaning NGINX's `proxy-connect-timeout` is not triggered
- After 60 seconds the request is killed, likely caused by `proxy-read-timeout` in NGINX
- Given the failure more - NGINX attempts to reconnect with another host - this request goes through
- After this successful connection, everything is dropped because the client in was torn down after the 10s mark

What we would have wanted to happen is to realize that the first connection was unsuccessful much quicker and to try another host before the 10 second userland timeout is triggered (after which point we can't recover).

In theory we could lower the `proxy-read-timeout` within our NGINX ingress, to allow this connection to try multiple hosts within 10s. However, this timeout is shared by all incoming requests and lowering it from 60s to 3s would have many large unintended impacts elsewhere.

So instead of doing that, we set an ACK timeout within the `graphql-ws` connection and retry based on that failure. This way we can attempt multiple hosts without our 10s userland timeout without affecting other parts of Gadget.

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -5,6 +5,7 @@ import { ExecutionResult } from "graphql";
 import {
   Client as SubscriptionClient,
   ClientOptions as SubscriptionClientOptions,
+  CloseCode,
   createClient as createSubscriptionClient,
   Sink,
 } from "graphql-ws";
@@ -36,35 +37,6 @@ export interface GadgetConnectionOptions {
 }
 
 const isCloseEvent = (event: any): event is CloseEvent => event?.type == "close";
-
-/** Helper function to await the setup dance on a graphql-ws Client object before we start using it. */
-export const connectionOpened = traceFunction(
-  "api-client.await-websocket-open",
-  async (subscriptionClient: SubscriptionClient, timeoutMs: number) => {
-    const unsubscribes: Function[] = []; // eslint-disable-line @typescript-eslint/ban-types
-    const clearListeners = () => unsubscribes.forEach((fn) => fn());
-    await new Promise<void>((resolve, reject) => {
-      const wrappedReject = (err: unknown) => {
-        clearTimeout(timeout);
-        reject(err);
-      };
-
-      const wrappedResolve = () => {
-        clearTimeout(timeout);
-        resolve();
-      };
-
-      const timeout = setTimeout(() => {
-        void subscriptionClient.dispose();
-        wrappedReject(new Error("Timeout opening websocket connection to Gadget API"));
-      }, timeoutMs);
-
-      unsubscribes.push(subscriptionClient.on("connected", wrappedResolve));
-      unsubscribes.push(subscriptionClient.on("closed", wrappedReject));
-      unsubscribes.push(subscriptionClient.on("error", wrappedReject));
-    }).finally(clearListeners);
-  }
-);
 
 /**
  * Represents the current strategy for authenticating with the Gadget platform.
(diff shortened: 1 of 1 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -350,4 +321,50 @@ export class GadgetConnection {
...
+    const clearListeners = () => {
+      unsubscribes.forEach((fn) => fn());
+      unsubscribes = [];
+    };
+
+    return await new Promise<SubscriptionClient>((resolve, reject) => {
+      const wrappedReject = (err: any) => {
+        if (err && err.code == CloseCode.ConnectionAcknowledgementTimeout && attempts > 0) {
+          attempts -= 1;
+          subscriptionClient = this.newSubscriptionClient(overrides);
```
thegedge, line 342: We should probably ` subscriptionClient.dispose` the old client, eh?

angelini: ohh good call, I'll add that

thegedge:
I think this one is worth a 🎩 , so I'll give that a try sometime today.

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -350,4 +321,51 @@ export class GadgetConnection {
...
+
+    const clearListeners = () => {
+      unsubscribes.forEach((fn) => fn());
+      unsubscribes = [];
+    };
+
+    return await new Promise<SubscriptionClient>((resolve, reject) => {
+      const wrappedReject = (err: any) => {
+        if (err && err.code == CloseCode.ConnectionAcknowledgementTimeout && attempts > 0) {
+          attempts -= 1;
```
thegedge, line 341: Any reason you chose to do retries instead of repeating until the `setTimeout` fires? If the user is game waiting for N seconds overall, should we help them out by retrying until that timeout occurs? Or do you see it as "if we can't do it in N attempts, something bad is going on, just fail fast"?

thegedge (approved):
Just some questions, but game to 🚢 as is!

packages/api-client-core/src/GadgetConnection.ts
angelini: That's a good point. I added the `attempts` counter out of a general fear of any kind of infinite loop + hitting a service we control.

But the userland timeout should be good enough to stop this.

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -166,24 +138,23 @@ export class GadgetConnection {
...
-        lazyCloseTimeout: 100000,
-        retryAttempts: 0,
-      });
-
+      let subscriptionClient: SubscriptionClient | null = null;
       let transaction;
       try {
         // The server will error if it receives any operations before the auth dance has been completed, so we block on that happening before sending our first operation. It's important that this happens synchronously after instantiating the client so we don't miss any messages
-        await connectionOpened(subscriptionClient, options.connectionTimeoutMs ?? 10000);
+        subscriptionClient = await this.waitForOpenedConnection(3, options.connectionTimeoutMs ?? 10000, {
```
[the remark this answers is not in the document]
angelini: The reason I split them out here is that the `options` for this function has a very specific type and it is forwarded as is to `newSubscriptionClient`.

These two first args are settings for `waitForOpenedConnection` whereas the options obj is for `newSubscriptionClient`.

I wonder if there's a way to make that more clear.

airhorns:
Apologies for the radio silence from me! I think this code is getting complicated enough that we really need some integration tests of it in this repo. I think the move might be to use a real websocket server running `graphql-ws` in the same process as the test is running in -- want to take a crack at setting that up sometime soon @angelini ?

angelini:
> I think this code is getting complicated enough that we really need some integration tests of it in this repo. I think the move might be to use a real websocket server running graphql-ws in the same process as the test is running in

Agreed, I was testing this by lowering the timeouts to the minimum and reproducing the error locally - but it's a pain to test. And there's no good harness to test it in this repo.

If this solution turns out to be helpful - I'd like to refactor this connection code and in doing so I'll add a new test suite for it.

PR #21, opened 2022-03-17
Added script to make sure api-client-core dep versions are correct

scott-rc:
I added a script so that our binding packages don't accidentally depend on an incompatible version of `@gadgetinc/api-client-core`.

This should help prevent our generated clients from using a `@gadgetinc/api-client-core` version that our bindings don't support - which causes TypeScript to complain.

### DIFF
--- a/scripts/tsconfig.json
+++ b/scripts/tsconfig.json
@@ -1,0 +1,15 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "lib": ["es2020"],
+    "moduleResolution": "node",
+    "types": ["node"]
+  },
+  "include": ["."],
+  "ts-node": {
+    "esm": true,
+    "moduleTypes": {
+      "*": "esm"
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


scripts/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "moduleResolution": "Node"
+  },
```
thegedge, line 5: nit question: Should we make these options closer to our other tsconfigs? Like https://github.com/gadget-inc/js-clients/blob/98f6b02e00cb6a1624a5ed81cb949fea5a0b6519/packages/api-client-core/tsconfig.json#L3-L11

super nit: not sure if it's case-sensitive, but we typically use the all lowercase value `"node"`

airhorns (approved):
Yeah makes a lot of sense to me!

scripts/tsconfig.json
scott-rc: I added the `lib` and `type` options but *I don't think* the other options referenced in `api-client-core` are relevant to the script folder.

I also lowercased `"node"`. I don't have a preference, it was my editor that autocompleted `"Node"`.

PR #23, opened 2022-03-18
v0.3.0

scott-rc:
Now that #21 has been merged, specifically [this line](https://github.com/gadget-inc/js-clients/pull/21/files#diff-1f344ac391eeecc21ec0f01fb07430a47f4b80d20485c125447d54c33c4bbfc4R21), I think we should release a new version of the react package.

Since we bumped the `api-client-core`'s left most non-zero digit, I assume this means we should bump reacts as well? Let me know what you think 🤔

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.2.3",
+  "version": "0.3.0",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",


thegedge (commented):
Yeah, I think it makes sense to release a new version, even if it's just a dependency bump 👍 

I think whether or not to bump the "major" or "minor" number (put those in quotes, because I interpret them as second and third when pre-v1) is likely subjective. A major bump in a dependency doesn't necessarily mean a breaking change in this library, but it definitely simplifies things when one can't easily or fully assess the impact of the dep bump.

What do we think about keeping the versions in "major lockstep"? In other words, every client we build has the same "major" version as `api-client-core` (but can differ on the minor/revision numbers).

airhorns (approved):

airhorns:
Yep I think we should do this! 

> What do we think about keeping the versions in "major lockstep"? In other words, every client we build has the same "major" version as api-client-core (but can differ on the minor/revision numbers).

I think thats a good call, keeps shit a bit simpler.

thegedge (approved):
Had a good discussion about why to not keep them in lockstep: if we ever wanted to do a big rewrite of the react client, for example, it would be odd to force a bump on api-client-core. They're not super coupled together, and such a bump would waste peoples' time if they think they're receiving a breaking change (when really it's a no-op major version bump).

PR #25, opened 2022-03-21
Add release workflow to automatically publish npm packages

scott-rc:
Added a `release.yml` workflow to automatically publish our packages. It uses [this action](https://github.com/JS-DevTools/npm-publish) under the hood which only publishes a package if it's `package.json`'s `version` changes.

I made the `release.yml` workflow re-use the `test.yml` workflow so we don't have to duplicate our test steps. The `release.yml` is only set to run on the main branch.

I know I initially wanted to have this be a manual action but I changed my mind... go fast and break things right!?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,29 @@
+name: Release
+on:
+  push:
+#    branches:
+#      - main
+jobs:
+  test:
+    uses: ./.github/workflows/test.yml
+  release:
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: ./.github/actions/setup-test-env
+      - name: Build all packages
+        shell: nix develop -c bash -eo pipefail -l {0}
+        run: yarn build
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: ./.github/actions/setup-test-env
+      - name: Build all packages
+        shell: nix develop -c bash -eo pipefail -l {0}
+        run: yarn build
+      - name: Publish @gadgetinc/api-client-core
+        uses: JS-DevTools/npm-publish@v1
```
thegedge, line 19: I'm going to assume this one is smart enough to know whether or not it needs to release? :)

thegedge (approved):
> go fast and break things right

`:gottagofast:`

In all seriousness though, this is great. It's likely super rare that we merge a version bump and then suddenly think "oops, not ready to publish".

airhorns (approved):
This is majestic! Nice!

PR #26, opened 2022-03-24
Added inlineSources: true to tsconfig.json

scott-rc:
**What**

This fixes our source map issues when someone uses our react client with create-react-app v5.

```shell
Failed to parse source map from 'node_modules/@gadgetinc/api-client-core/src/AnyClient.ts' file:
Error: ENOENT: no such file or directory, open 'node_modules/@gadgetinc/api-client-core/src/AnyClient.ts'

Failed to parse source map from 'node_modules/@gadgetinc/react/src/index.ts' file:
Error: ENOENT: no such file or directory, open 'node_modules/@gadgetinc/react/src/index.ts'
```

**Why**

Because we want a seamless out-of-the-box experience with creact-react-app.

**How**

Webpack can't find the original source code that our source maps are pointing to, so we have 2 options:

1. Include the `src` directory in our bundle
2. Set `inlineSources: true` in our tsconfig.json

I went with option 2 because it's what [@opentelemtry-js](https://github.com/open-telemetry/opentelemetry-js/blob/main/tsconfig.base.json) does...

**Note**

Our `gql-query-builder` dependency has the same source map issue so our users are still going to get warnings. I'm going to start working on a PR for them after this.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,7 @@
     "allowJs": true,
     "noImplicitAny": true,
     "sourceMap": true,
+    "inlineSources": true,
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
     "forceConsistentCasingInFileNames": true,


thegedge:
Nice find! What are the pros/cons of inline source maps vs the other options?

scott-rc:
> Nice find! What are the pros/cons of inline source maps vs the other options?

I can't find any pros/cons comparing `inlineSources` vs including the original ts files.

The only comparisons I can find are between `sourceMap` (emit separate `.js.map` files) and `inlineSourceMap` (embed the source map inside the `.js` file). We chose `sourceMap` and I think we should stick with it.

thegedge (approved):
My only concern would be that this makes the generated JS larger. If someone doesn't need to use the source maps, they still incur the hit for them. What do you think? Should we instead include the TS files, or is the increase in size small enough that we can worry about it later?

I'll ✅ and let you make that decision :)

scott-rc:
> My only concern would be that this makes the generated JS larger.

I don't think this will be a problem because

1. `inlineSources` will only increase the size of our source maps
2. our source maps will still be generated as separate `.js.map` files

Increasing the size of our source map files *might* be annoying, but supposedly they only get loaded [when you open the developer tools](https://stackoverflow.com/questions/44315460/when-do-browsers-download-sourcemaps#:~:text=Source%20Maps%20are%20downloaded%20and,not%20aware%20of%20the%20sourceMap.&text=Once%20Devtools%20is%20attached%20the%20modal%20triggers%20and%20enables%20the%20sourceMapManager.).

thegedge:
Ah, my bad. For some reason, in my brain, I was thinking the maps were inlined in the source, not the source inlined in the maps. I think I was reading the docs for `inlineSourceMaps`, not `inlineSources`. Forgive my RTFM stupidity 🤦 

🚀

airhorns (approved):
Oh heck, this makes a lot of sense, nice find! @thegedge I initially made the same mistake as well -- this is the content within the sourcemaps, not where the sourcemaps themselves are located, and I think we're all on the same page that its important the JS file doesn't actually contain an inline sourcemap. @scott-rc good call on investigating an existing lib to see how they do it 👍

PR #27, opened 2022-03-24
Update gql-query-builder to v3.6.0

scott-rc:
The source map fix for gql-query-builder [was merged](https://github.com/atulmy/gql-query-builder/pull/69) 🎉

Let's update to the newest version.


airhorns (approved):
🤘

PR #31, opened 2022-03-31
Conditionally include sort, filter and search if options provided

garymardell:
Some models do not support sorting, filtering and searching however previously the query was hard-coded to include these options regardless. This meant that the generated query referenced fields and graphql types (`WidgetSort`, `WidgetFilter`) that do not exist. 

This PR updates the query builder to conditionally include the variables only if an option is provided and therefore they will not appear as part of the query. The generated SDK that uses `api-client-core` will not pass in these options if not supported by the model. The tests have been updated to reflect the newly generated queries. 

@airhorns


airhorns (approved):
Works for me!

garymardell:
> There are a couple more changes happening in this repo in the next couple days such that we can release after that?

Yup, this works for me

airhorns:
Gave you perms

airhorns:
Gonna merge this so it gets in before the version bump

PR #33, opened 2022-04-04
Changed maybeFindOne to return record instead of property

scott-rc:
I noticed `maybeFindOne` was returning an indexed property instead of the record itself. I don't think that's the intended behaviour.

TypeScript allowed it because `findOne` returned `GadgetRecord<any>`, which is equivalent to `GadgetRecordImplementation<any> & any`, which is equivalent to `any`. I've changed the `InternalModelManager`s methods to return `GadgetRecord<unknown>` which should help prevent something like this from accidentally happening again.


packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -194,19 +194,19 @@ export class InternalModelManager {
...
-    return hydrateRecord(response, result);
+    return await hydrateRecord(response, result);
   }
 
-  async maybeFindOne(id: string): Promise<GadgetRecord<any> | null> {
-    const list = await this.findOne(id, false);
-    return list[0] ?? null;
+  async maybeFindOne(id: string): Promise<GadgetRecord<unknown> | null> {
+    const record = await this.findOne(id, false);
+    return record ?? null;
```
scott-rc, line 206: This is the change in question

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -194,19 +194,19 @@ export class InternalModelManager {
     this.capitalizedApiIdentifier = camelize(apiIdentifier);
   }
 
-  async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<any>> {
+  async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<unknown>> {
     const response = await this.connection.currentClient.query(internalFindOneQuery(this.apiIdentifier), { id }).toPromise();
     const assertSuccess = throwOnEmptyData ? assertOperationSuccess : assertNullableOperationSuccess;
     const result = assertSuccess(response, ["internal", this.apiIdentifier]);
-    return hydrateRecord(response, result);
+    return await hydrateRecord(response, result);
```
LucasUnplugged, line 201: Do we need `await` here? It doesn't look like `hydrateRecord` is an async function. If it's because you know (or suspect) that `hydrator` might be async, then should we also change `hydrateRecord` to be async + await the hydrator call?

scott-rc: Interesting... I added the `await` because ESLint told me to

```
/Users/scott/Code/gadget/js-clients/packages/api-client-core/src/InternalModelManager.ts
  201:12  error  Returning an awaited promise is required in this context  @typescript-eslint/return-await
```

but you're right, it's not an async function so I'm not exactly sure why ESLint is complaining... I'm gonna look into this more.

LucasUnplugged: Waaaa??? That's so weird! It looks like it's because of the change from `any` to `unknown`. I'm fine with leaving it in, just a little confused why eslint thinks it's needed. 🤷🏻‍♂️

LucasUnplugged (changes requested):
@scott-rc Looks like using `unknown` is breaking the resulting output. `any` is not a good answer at all, but I'm going to investigate typing more closely next, so maybe revert that part of your PR for now, and I can see about improving that next.

LucasUnplugged:
This is what I mean. Using `any` is a total copout for now, but I think it would be better than causing TS errors for consumers of the client, for now.

Screen Shot 2022-04-04 at 3 46 24 PM

LucasUnplugged (approved):
Thanks @scott-rc !

scott-rc:
Yeah, I fell down a TypeScript rabbit hole trying to figure out the ESLint error... I reverted the `any` -> `unknown` change.

PR #34, opened 2022-04-04
Fix empty `findFirst` cases

LucasUnplugged:
Supports [sc-1556]

Adds logic to properly address empty `findFirst` empty results. The base `findMany` runner and internal method now account for 3 possible expected behaviours:
1. For `findMany`, where results of empty array are possible.
2. For `findFirst`, where results of empty array throw an error.
3. For `maybeFindFirst`, where results of an empty array are returned as `null`, with no error thrown.

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -206,12 +206,21 @@ export class InternalModelManager {
     return record ?? null;
   }
 
-  async findMany(options?: Record<string, any>, throwOnEmptyData = true, isFirstQuery = false): Promise<GadgetRecordList<any>> {
+  async findMany(options?: Record<string, any>, throwOnEmptyData?: boolean, isFirstQuery = false): Promise<GadgetRecordList<any>> {
     const response = await this.connection.currentClient
       .query(internalFindManyQuery(this.apiIdentifier, isFirstQuery), options)
       .toPromise();
-    const assertSuccess = throwOnEmptyData ? assertOperationSuccess : assertNullableOperationSuccess;
-    const connection = assertSuccess(response, ["internal", `list${this.capitalizedApiIdentifier}`]);
+
+    let connection;
+    if (throwOnEmptyData === false) {
+      // If this is a nullable operation, don't throw errors on empty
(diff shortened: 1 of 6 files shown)


packages/api-client-core/src/support.ts
```diff
@@ -127,11 +129,13 @@ export const assertOperationSuccess = (response: OperationResult<any>, dataPath:
   }
 
   const result = get(response.data, dataPath);
+  const edges = get(result, ["edges"]);
+  const dataArray = edges ?? result;
```
LucasUnplugged, line 133: In some cases, I was seeing the actual content array nested inside of here, which was not caught by the default `dataPath` that those instances were passing. I felt it would be more robust to test for that centrally (here), than to have to figure out when to add that to the path and when not to.

thegedge: I'm game for that. It's not really something that people use outside of this library, so we can always come back later and adjust.

thegedge (approved):

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -206,12 +206,21 @@ export class InternalModelManager {
...
-    const connection = assertSuccess(response, ["internal", `list${this.capitalizedApiIdentifier}`]);
+
+    let connection;
+    if (throwOnEmptyData === false) {
+      // If this is a nullable operation, don't throw errors on empty
+      connection = assertNullableOperationSuccess(response, ["internal", `list${this.capitalizedApiIdentifier}`]);
+    } else {
+      // Otherwise, passthrough the `throwOnEmptyData` flag, to account for
+      // `findMany` (allows empty arrays) vs `findFirst` (no empty result) usage.
+      connection = assertOperationSuccess(response, ["internal", `list${this.capitalizedApiIdentifier}`], throwOnEmptyData);
```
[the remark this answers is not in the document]
LucasUnplugged: Would it? From what I saw in `assertOperationSuccess`, it would only throw id the data was `undefined` or `null`, which I'm assuming it would never be for a `findMany` operation.

In any case, I'm not picky about this at all, but though a search for a list would be more reasonably expected to return an empty array (rather than an error).

Either way, if you're okay with it like this, I'll go ahead and merge.

PR #35, opened 2022-04-05
Fix type check in `maybeFindOne`

LucasUnplugged:
The `maybeFindOne` func + hook were not properly checking for non-null types — the hydration process was creating an object with a few basic properties.

Added proper checking for this, in that way that doesn't make TS scream.

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -1,5 +1,5 @@
 import { GadgetConnection } from "./GadgetConnection";
-import { GadgetRecord } from "./GadgetRecord";
+import { GadgetRecord, RecordShape } from "./GadgetRecord";
 import { GadgetRecordList } from "./GadgetRecordList";
 import { GadgetTransaction } from "./GadgetTransaction";
 import {
@@ -194,14 +194,16 @@ export class InternalModelManager {
     this.capitalizedApiIdentifier = camelize(apiIdentifier);
   }
 
-  async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<any>> {
+  async findOne(id: string, throwOnEmptyData: true): Promise<GadgetRecord<RecordShape>>;
+  async findOne(id: string, throwOnEmptyData: false): Promise<GadgetRecord<RecordShape> | null>;
+  async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<RecordShape> | null> {
     const response = await this.connection.currentClient.query(internalFindOneQuery(this.apiIdentifier), { id }).toPromise();
     const assertSuccess = throwOnEmptyData ? assertOperationSuccess : assertNullableOperationSuccess;
     const result = assertSuccess(response, ["internal", this.apiIdentifier]);
-    return hydrateRecord(response, result);
+    return result ? hydrateRecord<RecordShape>(response, result) : null;
   }
 
-  async maybeFindOne(id: string): Promise<GadgetRecord<any> | null> {
+  async maybeFindOne(id: string): Promise<GadgetRecord<RecordShape> | null> {
     const record = await this.findOne(id, false);
     return record ?? null;
   }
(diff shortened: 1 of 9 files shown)


packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -203,7 +203,7 @@ export class InternalModelManager {
 
   async maybeFindOne(id: string): Promise<GadgetRecord<any> | null> {
     const record = await this.findOne(id, false);
-    return record ?? null;
+    return record && "id" in record ? record : null;
```
scott-rc, line 206: Interesting,`this.findOne(id, false)` can return a non-null object without an id?

Should that even be allowed to happen?

LucasUnplugged: So I don't think this could really happen previously (not sure about the hook, but the func couldn't), because it would throw if the data was empty.

BUT... once I added the ability to not throw on empty data (for the `maybeFindOne` func), it meant that the code could execute past that step, onto the hydration step, which still resulted in an object being created with a few basic keys. So I'm checking for the one key that it will always have if the data is not empty.

LucasUnplugged: So I missed your point, which @thegedge reiterated as well — given the output type of `this.findOne`, a `null` output should not be possible.

Based on his suggestion, I've tightened up the typing around `GadgetRecord` — similar to what you had done with `unknown`, but a bit more specific, with an `extends` based on `Record<string, any>`.

LucasUnplugged:
@thegedge I tried doing it exactly how we discussed, using `Record<string, unknown>` and not tacking the nullish types with it, but neither of those attempts worked, because of the way one of the nested types works (`Select`, I think it was).

But this at least tightens things up a bit, so it's still a win, IMO.

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -194,16 +194,16 @@ export class InternalModelManager {
...
     const result = assertSuccess(response, ["internal", this.apiIdentifier]);
-    return hydrateRecord(response, result);
+    return hydrateRecord<RecordShape>(response, result);
   }
 
-  async maybeFindOne(id: string): Promise<GadgetRecord<any> | null> {
+  async maybeFindOne(id: string): Promise<GadgetRecord<RecordShape> | null> {
     const record = await this.findOne(id, false);
-    return record ?? null;
+    return "id" in record ? record : null;
```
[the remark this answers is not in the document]
LucasUnplugged: So not as is, BUT... now that I think about it, `findOne` is going to throw on empty anyway, right (rhetorical question — yes, it will)? So if I change the logic there to skip hydration on null (and thus change the return to include null as an option), then yes, I can do what you're saying.

I didn't want to touch the hydration part at first, but it shouldn't affect the behaviour at all, and it will be more robust to changes, as you say. I'll do that then.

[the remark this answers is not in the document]
thegedge: If you're up for it, you can also tighten up the return types by overloading the function signature for when throwOnEmptyData is true/false, but I'm game to come back to that later 🚀

thegedge (approved):
👏 🚢

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.4.1",
+  "version": "0.4.2",
```
thegedge, line 3: Oh, one thing I forgot: I'm not sure how we're publishing now, but I think it would be good for this change to be the last commit.

scott-rc: Assuming the tests pass, our [release action](https://github.com/gadget-inc/js-clients/blob/main/.github/workflows/release.yml) will pick up the change and immediately publish the new version.

scott-rc (approved):

packages/react/package.json
thegedge: Then probably good to have the bump at the end. I think that better reflects reality. It would feel weird to have the bump in one commit, and changes after it also included in that release, eh?

LucasUnplugged: Oh, good question. I know @scott-rc did some work on that recently: Scott, do you know if that commit order helps/matter at this point?

scott-rc: From the release action's perspective, it won't matter. However, from our git history perspective, the version was changed on the first commit but the published package will have 3 additional commits inside of it.

PR #36, opened 2022-04-06
Restore original signature of `findOne` function

LucasUnplugged:
Unfortunately, the signature change to the `findOne` function resulted in typing issues, even with the overload designed to prevent it.

However, I've added a robust way to check if the record is empty, on the GadgetRecord class itself, so I'm no longer relying on the `id` prop.

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -194,19 +194,19 @@ export class InternalModelManager {
     this.capitalizedApiIdentifier = camelize(apiIdentifier);
   }
 
-  async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<Record<string, any> & ({ id: string } | { id: never })>> {
+  async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<RecordData>> {
     const response = await this.connection.currentClient.query(internalFindOneQuery(this.apiIdentifier), { id }).toPromise();
     const assertSuccess = throwOnEmptyData ? assertOperationSuccess : assertNullableOperationSuccess;
     const result = assertSuccess(response, ["internal", this.apiIdentifier]);
     return await hydrateRecord(response, result);
   }
 
-  async maybeFindOne(id: string): Promise<GadgetRecord<RecordShape> | null> {
+  async maybeFindOne(id: string): Promise<GadgetRecord<RecordData> | null> {
     const record = await this.findOne(id, false);
-    return record.id ? record : null;
+    return record.isEmpty() ? null : record;
   }
 
-  async findMany(options?: Record<string, any>, throwOnEmptyData?: boolean, isFirstQuery = false): Promise<GadgetRecordList<any>> {
+  async findMany(options?: RecordData, throwOnEmptyData?: boolean, isFirstQuery = false): Promise<GadgetRecordList<any>> {
     const response = await this.connection.currentClient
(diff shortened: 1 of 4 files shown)


LucasUnplugged:
Sorry for the back-and-forth gents. It's been a bit frustrating trying to get the various integration, fixture, and code-gen tests working with all these changes.

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -194,21 +194,19 @@ export class InternalModelManager {
     this.capitalizedApiIdentifier = camelize(apiIdentifier);
   }
 
-  async findOne(id: string, throwOnEmptyData: true): Promise<GadgetRecord<RecordShape>>;
-  async findOne(id: string, throwOnEmptyData: false): Promise<GadgetRecord<RecordShape> | null>;
-  async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<RecordShape> | null> {
```
thegedge, line 199: I guess maybe we got the overloads wrong. Perhaps to deal with an optional arg one needs to actually do something like:

```
async findOne(id: string): Promise<GadgetRecord<RecordShape>>;
```

We'd probably also have to something similar on `findOneRunner` to make it work. As much as I'd like to do that, let's worry about that in a future PR.

packages/api-client-core/src/operationRunners.ts
```diff
@@ -29,7 +29,7 @@ export const findOneRunner = async <Shape extends RecordShape = any>(
   const response = await modelManager.connection.currentClient.query(plan.query, plan.variables).toPromise();
   const assertSuccess = throwOnEmptyData ? assertOperationSuccess : assertNullableOperationSuccess;
   const record = assertSuccess(response, [operation]);
-  return record ? hydrateRecord<Shape>(response, record) : null;
+  return hydrateRecord<Shape>(response, record);
```
thegedge, line 32: Just a refresher, but `hydrateRecord` is okay with the possibility that record could be `null`, eh?

thegedge (approved):
🚀 

This is hard stuff, working between repos. Perhaps we could find a better way of testing this without having to push? 🤔

What I've been doing for `mobx-quick-tree` is building it in a separate package and just doing a `cp -r` of the source over into a workspace project. It's not ideal, but it's been working well for making sure things work. I wonder if we a similar strategy could be applied to `js-clients`?

packages/api-client-core/src/operationRunners.ts
LucasUnplugged: Yeah, that's actually why I missed it at first: it does an `Object.assign(defaultObject, incomingData)`, type of thing, so you end up with just the default object.

In practical terms, it will either throw an error, or end up with a null return, so there shouldn't really be any instances where that "empty" object gets used.

LucasUnplugged:
Yeah, I might be able to use gitpkg to publish off a branch, but I didn't have the right permissions last night. I'll look at it once I have more time.

thegedge:
Yeah, gitpkg is perhaps the best. I wish there was an even easier way though. I think `yarn link` is that, but I would want node resolution to understand symlinks.

packages/api-client-core/src/InternalModelManager.ts
LucasUnplugged: Yeah, I can try that another time — maybe when I'm working on fixing the typing issues of the hooks.

PR #37, opened 2022-04-07
Use query builder for internal find many query

garymardell:
Implements conditional arguments for `search`, `sort` and `filter` for the internal find many query. The internal version of https://github.com/gadget-inc/js-clients/pull/31.

In order to make this easier I have converted this to also use the graphql query builder that is used by the `operationBuilder` to generate the query rather than using a hardcoded string. I have also broken out into 2 methods for the `findFirst` case and the default `findMany` as the queries were sufficiently different (no pagination options, no page info, no cursor and different operation name). This meant I could remove a bunch conditional logic within the query itself. 

I have added a bunch of tests similar to those in `operationBuilder.spec` for the generated queries.

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -1,3 +1,4 @@
+import { query } from "gql-query-builder";
 import { GadgetConnection } from "./GadgetConnection";
 import { GadgetRecord, RecordShape } from "./GadgetRecord";
 import { GadgetRecordList } from "./GadgetRecordList";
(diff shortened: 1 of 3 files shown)


packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -44,61 +45,78 @@ export const internalFindOneQuery = (apiIdentifier: string) => {
...
+              first: { value: options?.first, type: "Int" },
+              last: { value: options?.last, type: "Int" },
+            },
+          },
+        ],
+      },
+    ],
+    null,
+    { operationName: `InternalFindMany${capitalizedApiIdentifier}` }
+  );
```
airhorns, line 119: Woop woop for objects instead of strings!

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -207,9 +225,8 @@ export class InternalModelManager {
   }
 
   async findMany(options?: RecordData, throwOnEmptyData?: boolean, isFirstQuery = false): Promise<GadgetRecordList<any>> {
-    const response = await this.connection.currentClient
-      .query(internalFindManyQuery(this.apiIdentifier, isFirstQuery), options)
-      .toPromise();
+    const plan = isFirstQuery ? internalFindFirstQuery(this.apiIdentifier, options) : internalFindManyQuery(this.apiIdentifier, options);
```
airhorns, line 228: If this function is now calling different functions based on `isFirstQuery` or not, is there any point to re-using it from `findFirst`? Should we just make them each call their own query builder? Lucas and I went back and forth on how much to re-use these functions to implement each other, and I like the re-use to avoid bloating the bundle size too much, and/or so we don't need to duplicate too much, but it also seems a bit convoluted at this point.

garymardell: Yeah this could be separated out. Once I refactored the query itself into using objects rather than strings it was becoming harder to read with conditionals inside objects. Other than reading the same `listX` graphql field very little of the query was really shared.

garymardell: I've pushed another commit that breaks it out into it's own operation to get a feel for what it's like. I think it's definitely simpler, it removes the `throwOnEmptyData` option from `findMany` and `isFirstQuery`. This means the combination of options is significantly reduced.

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -44,61 +45,78 @@ export const internalFindOneQuery = (apiIdentifier: string) => {
...
+          {
+            operation: `list${capitalizedApiIdentifier}`,
+            fields: [
+              {
+                edges: ["node"],
+              },
+            ],
+            variables: {
+              ...defaultVariables,
+              first: { value: 1, type: "Int" },
```
LucasUnplugged, line 71: When I tried using this query builder (elsewhere), I was never able to hardcode values like here; it always seemed to use the value _only_ for type definition (so "number" here), but it was then list that field as accepting an external variable, which you'd have to set.

Can you confirm that this behaviour is _not_ what we'll get here?

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -44,61 +45,78 @@ export const internalFindOneQuery = (apiIdentifier: string) => {
     `;
 };
 
-export const internalFindManyQuery = (apiIdentifier: string, isFirstQuery = false) => {
+export const internalFindFirstQuery = (apiIdentifier: string, options?: RecordData) => {
```
LucasUnplugged, line 48: Overall, I like (and prefer) this separation, but was encouraged to avoid it when I did this. I think this is cleaner, though, with each function serving one purpose.

I think that once this changes, we'll have to update some stuff in the main monorepo, including docs — but I believe that only matters once we update the app to point to a new version of the client.

LucasUnplugged (commented):
Just have the one question; looks good otherwise!

packages/api-client-core/src/InternalModelManager.ts
LucasUnplugged: To elaborate: IMO, the desired behaviour for `findFirst` is that there is no variable for `first`; you get the hardcoded value of 1, and _that's it_. Otherwise, it's a misnomer, which I would find weird 😅

airhorns: Yeah, this is a strange thing about how `gql-query-builder` works, it returns both a query document AND query variables to then pass to that document. When you have values to pass to the document, why not just hardcode them in it? It totally could, but it also supports generating queries with variables you don't yet have the values for, and I guess they decided that all variables should be passed the same. This function returns a `plan` object that has both a `.query` and a `.variables` on it, and we pass those variables down to the execution bits later.

airhorns (approved):

packages/api-client-core/src/InternalModelManager.ts
LucasUnplugged: @airhorns oh, I see what you're saying: we're just feeding those variables back in, like on line 238; and since we're not using `value: options?.first`, there's no way to override it externally, so it ends up being the same thing as if it was hardcoded.

I did a quick scan of the codebase, and the only other call to `internalFindManyQuery` is related to docs, so that's fine.

LucasUnplugged (approved):
Just do a version bump, but LGTM! 🚢 🚢

packages/api-client-core/src/InternalModelManager.ts
garymardell: > oh, I see what you're saying: we're just feeding those variables back in, like on line 238; and since we're not using value: options?.first, there's no way to override it externally, so it ends up being the same thing as if it was hardcoded.

That is correct. 

> I did a quick scan of the codebase, and the only other call to internalFindManyQuery is related to docs, so that's fine.

Yeah, I will update the `exampleInternalFindFirstQuery` to use the new method once this has been merged.

PR #38, opened 2022-05-04
Add option to specify initial browser session token

kira-bruneau:
This will allow us to manage browser session tokens our self for the public app oauth flow.

fixes ggt-1457

### DIFF
--- a/packages/api-client-core/src/ClientOptions.ts
+++ b/packages/api-client-core/src/ClientOptions.ts
@@ -10,6 +10,15 @@ export interface ClientOptions {
 
 /** Options to configure a specific browser-based authentication mode */
 export interface BrowserSessionAuthenticationModeOptions {
+  /**
+   * The initial token to set for browser authentication.
+   * This is useful when your session is initialized by some external authentication system, like OAuth.
+   */
+  initialToken?: string;
+
+  /**
+   * Configures how the authentication token is persisted. See `BrowserSessionStorageType`.
+   */
   storageType: BrowserSessionStorageType;
 }
 
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -10,6 +10,7 @@ import {
   Sink,
 } from "graphql-ws";
 import WebSocket from "isomorphic-ws";
+import { isObject } from "lodash";
 import { getCurrentSpan } from ".";
 import type { AuthenticationModeOptions, BrowserSessionAuthenticationModeOptions } from "./ClientOptions";
 import { BrowserSessionStorageType } from "./ClientOptions";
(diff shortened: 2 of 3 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -114,13 +115,21 @@ export class GadgetConnection {
 
   enableSessionMode(options?: true | BrowserSessionAuthenticationModeOptions) {
     this.authenticationMode = AuthenticationMode.BrowserSession;
+
+    let sessionTokenStore;
```
jasong689, line 119: nit: Could we type this as `BrowserStorage` like in the class property? Doesn't need to be if it's inferred but I don't think it is

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -114,13 +115,21 @@ export class GadgetConnection {
...
+      sessionTokenStore = window.localStorage;
     } else if (options.storageType == BrowserSessionStorageType.Session) {
-      this.sessionTokenStore = window.sessionStorage;
+      sessionTokenStore = window.sessionStorage;
     } else {
-      this.sessionTokenStore = new InMemoryStorage();
+      sessionTokenStore = new InMemoryStorage();
+    }
+
+    if (options && isObject(options) && "initialToken" in options && options.initialToken) {
```
jasong689, line 128: nit: I don't think the check of `options` is necessary here. I believe `isObject` returns false if `options` is undefined

jasong689 (approved):
Code looks good, but are there any tests we can add/should add for this?

jasong689:
Might need to find out from @kristianpd or @airhorns if we need to bump the package version

packages/api-client-core/src/GadgetConnection.ts
kira-bruneau: Oh yep, thanks!

packages/api-client-core/src/GadgetConnection.ts
kira-bruneau: we should need to explicitly type this here, it should be inferred from use.

kira-bruneau:
> Might need to find out from @kristianpd or @airhorns if we need to bump the package version

Oh yep, I think we'll just need to bump the minor version.

packages/api-client-core/src/ClientOptions.ts
```diff
@@ -10,6 +10,7 @@ export interface ClientOptions {
 
 /** Options to configure a specific browser-based authentication mode */
 export interface BrowserSessionAuthenticationModeOptions {
+  initialToken?: string;
```
airhorns, line 13: Can you add a doc comment to explain what this does / if anyone else should use it?

airhorns (approved):
LGTM!

PR #39, opened 2022-05-06
Don't override `initialToken` on ws connect

kira-bruneau:
I was wondering why auth wasn't working with the `initialToken` option, and then I realized it was being clobbered by the existing setup that sets it when the websocket connection is opened :woman_facepalming:


kira-bruneau:
Hmm, I wonder if this could be an issue though if you have an existing expired token in your local storage :thinking:. I'll just push a change to check the options instead.

jasong689 (approved):
Ah that's fun 😄

PR #41, opened 2022-05-16
chore(react): expand react peer dependency to include ^18.0.0

danroberts:
I haven't found any incompatibilities with this library and react 18 --
but if you try to install it on a project with react 18 you will get a
peer dependency error, I think we can support both

image

https://linear.app/gadget-dev/issue/GGT-1522/gadgetincreact-is-not-compatible-with-react-18

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.4.2",
+  "version": "0.4.3",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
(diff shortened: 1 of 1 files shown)


airhorns (approved):
LGTM! Does urql / `@urql/react` support 18 ok?

LucasUnplugged (approved):
LGTM too!

danroberts:
> LGTM! Does urql / `@urql/react` support 18 ok?

It seems compatible. Some of the more 18-ish features like suspense seem to be in development still, but I don't see anything that would make it not work for basic use cases.

danroberts:
I've tested this out a bunch in a react 18 app, I don't see any problems, so I'm going to merge.

PR #44, opened 2022-05-27
Add shopify-react package for embedded/public app support

jasong689:
Adding a new package to help developers build embedded/public apps. This provider makes it easy to go through the oauth flow with both Shopify and Gadget. The end result is a set of hooks that can be used in combination with `@gadgetinc/react` and an app's js client to make authenticated calls to Gadget. This package is frontend only and would not work properly in an SSR context.

What I did to 🎩  this:
- run `yarn workspace @gadgetinc/shopify-react build`
- add `@gadgetinc/shopify-react` as a dependency with `file:/path/to/module` in faker
- run `yarn install` in faker
- update imports of the GadgetProvider and useGadgetContext to `@gadgetinc/shopify-react`
- test faker auth flow

closes GGT-1440
closes GGT-1603


packages/shopify-react/src/crypto.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const generateCodeVerifier = () => {
+  const bytes = Crypto.lib.WordArray.random(32);
+  return bytes.toString(Crypto.enc.Hex);
+};
+
+export const generateCodeChallenge = (verifier: string) => Crypto.SHA256(verifier).toString(Crypto.enc.Base64url);
+
+export const getGadgetAccessToken = async (appUrl: string, verifier: string, authCode: string) => {
+  const response = await fetch(`${appUrl}oauth/token`, {
```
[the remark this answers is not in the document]
kira-bruneau: Other than that though, this looks good to merge!

packages/shopify-react/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+Once you have your JS client installed, you can install the React hooks library with yarn or npm:
+
+```
+yarn add @gadgetinc/shopify-react @gadgetinc/react @shopify/app-bridge-react react
+# or
+npm install --save @gadgetinc/shopify-react @gadgetinc/react @shopify/app-bridge-react react
+```
+
+## Example usage
```
airhorns, line 47: Can we add some explicit notes about not following the instructions in the shopify docs and in the `@gadget/react` docs

packages/shopify-react/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./context";
+export * from "./crypto";
+export * from "./hooks";
+export * from "./Provider";
```
airhorns, line 4: I think we need some tests for some of this stuff, it's complicated!

packages/shopify-react/README.md
```diff
@@ -0,0 +1,122 @@
...
+export function MyComponent() {
+  return (
+    <GadgetProvider type={AppType.Embedded} gadgetAppUrl={appUrl} shopifyApiKey={apiKey} client={getClient}>
+        <ProductManager />
+    </GadgetProvider>
+  );
+}
+
+function ProductManager() {
+  const { client, loading, appBridge } = useGadgetContext<Client>();
```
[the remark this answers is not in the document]
kira-bruneau: Don't we want the `client` prop for the react hooks below, or do you mean this should be updated to `api`? Also I just remembered - `client` is optional, so we should probably add a check here too.

packages/shopify-react/src/crypto.ts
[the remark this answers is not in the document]
jasong689: Yep, this got me too so it's built using `URL` now

packages/react-shopify-app-bridge/spec/useGadget.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+const TestUseGadgetReturnsAppropriateTypes = () => {
+  const { loading, appType, isEmbedded, appBridge, canAuth, isAuthenticated } = useGadget();
+
+  assert<IsExact<typeof loading, boolean>>(true);
+  assert<IsExact<typeof isEmbedded, boolean>>(true);
+  assert<IsExact<typeof canAuth, boolean>>(true);
+  assert<IsExact<typeof isAuthenticated, boolean>>(true);
+  assert<IsExact<typeof appType, AppType | undefined>>(true);
+  assert<IsExact<typeof appBridge, ClientApplication<AppBridgeState> | undefined>>(true);
+};
```
kristianpd, line 15: Not sure this test really provides any value with maybe the exception of the `| undefined` assertions as we already have some lint checks.

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -0,0 +1,274 @@
...
+type GadgetProviderProps = {
+  children: JSX.Element | JSX.Element[];
+  appBridge?: ClientApplication<AppBridgeState>;
+  forceRedirect: boolean;
+  runningShopifyAuth: boolean;
+  runningGadgetAuth: boolean;
+  setAuth: (value: string) => void;
+  isEmbedded: boolean;
+  gadgetAppUrl: string;
+  token?: string;
```
kira-bruneau, line 34: Did we still want to :fire: this prop in addition to `client` that was removed?

packages/react-shopify-app-bridge/spec/useGadget.spec.ts
kira-bruneau: Yeah, TypeScript should take care of most of this, but it could be helpful to catch possible problems casting through any?

PR #45, opened 2022-06-21
Updating react-shopify-app-bridge oauth url to be app specific

jasong689:
We're updating Gadget's Shopify OAuth URLs to be app specific e.g. https://sample.gadget.app/api/connections/auth/shopify vs https://app.gadget.dev/connections/auth/shopify/callback. This PR updates `react-shopify-app-bridge` to use the new app specific route instead of the root domain.

This shouldn't be merged before https://github.com/gadget-inc/gadget/pull/3558

closes GGT-1744.


kristianpd:
Let's be careful to only merge this when we're ready with GGT-1753

PR #47, opened 2022-06-21
Create token key by host and shopify api key

jasong689:
We need to be able to support the same host on multiple apps so we'll need to keep tokens by both the shopify api key as well as host.

closes GGT-1734


kristianpd (approved):
:shipit:

PR #48, opened 2022-06-27
Only generate code challenge and verifier if we're in the destination…

jasong689:
We're running into some issues on Firefox where we end up in a redirect loop because we're generating and passing code challenges when in the incorrect context (e.g. during an app install which wasn't initiated via embedded app we're generating a code challenge on the initial redirect. Since firefox creates different localStorage contexts for when in the root of the host and in an iframe this breaks our redirect flow). A solution is to only generate the challenge and verifier when we've reached the embedded context. This however has the following effects:

- The initial install will only install the app to the Gadget app. There needs to be a second oauth dance for creating the session token for the embedded portion. This is something we can't get around for Firefox
- All other browsers have to follow the same flow as Firefox, even those that allow localStorage contexts to be shared when in the root vs iframe. This results in more redirects for Chrome when not necessarily needed

I don't think this is great in terms of user experience but allows us to support Firefox. There may be a better way to support this (or Shopify sessions), but I'm not sure if at this point is worthwhile to devote more time to.

### DIFF
--- a/packages/react-shopify-app-bridge/src/Provider.tsx
+++ b/packages/react-shopify-app-bridge/src/Provider.tsx
@@ -34,6 +34,7 @@ type GadgetProviderProps = {
   gadgetAppUrl: string;
   token?: string;
   originalQueryParams?: URLSearchParams;
+  inDestinationContext?: boolean;
 };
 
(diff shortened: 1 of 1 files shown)


jasong689:
we're using session tokens

PR #49, opened 2022-07-01
Add support for a custom auth method which can be used to implement platform native auths from userland

airhorns:
To build support for Shopify session token authentication, we need the client to be able to set arbitrary headers to send back to the Gadget backend. We don't wanna build super bespoke shopify code into a client used in a variety of non-shopify places, so this adds a hook were folks constructing the connection can create connections that pass a function that does whatever it wants to the outgoing request. Woop woop!

### DIFF
--- a/packages/api-client-core/src/ClientOptions.ts
+++ b/packages/api-client-core/src/ClientOptions.ts
@@ -47,7 +47,7 @@ export interface AuthenticationModeOptions {
   apiKey?: string;
 
   // Use a web browser's `localStorage` or `sessionStorage` to persist authentication information.
-  // This allows a web user to log in and keep their sessiontheir cookie (containing their credentials) will be sent.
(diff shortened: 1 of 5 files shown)


packages/api-client-core/src/ClientOptions.ts
```diff
@@ -47,7 +47,7 @@ export interface AuthenticationModeOptions {
   apiKey?: string;
 
   // Use a web browser's `localStorage` or `sessionStorage` to persist authentication information.
-  // This allows a web user to log in and keep their sessiontheir cookie (containing their credentials) will be sent.
+  // This allows a web user to log in and keep their session (containing their credentials) will be sent.
```
kristianpd, line 50: ~will be sent~

kristianpd (approved):

PR #50, opened 2022-07-01
Shopify session token auth provider

airhorns:
Fixes [GGT-1757](https://linear.app/gadget-dev/issue/GGT-1757) and fixes [GGT-1681](https://linear.app/gadget-dev/issue/GGT-1681)

We got a report from a user listing their app on the Shopify App Store that their app didn't pass review because it couldn't auth in incognito. After much investigation, we determined that the current localStorage/sessionStorage strategy won't ever work inside an iframe inside an incognito window with the default browser privacy settings. Browsers treat iframe'd localStorage/sessionStorage the same way as third party cookies, which means they just fully disallow access to these objects inside an iframe inside an incognito window. Just touching the value on `window` throws an error.

So, we're gonna do the thing we were intending on doing eventually a bit earlier, and make our provider use Shopify's suggested implementation for third party communication within the iframe context: session tokens. Session tokens are a JWT token we ask the AppBridge for, and then forward along to the Gadget backend, which knows how to decode them and bootstrap a Gadget session to match.  This means we can remove the (painstakingly crafted) code exchange stuff we were doing before that managed a token client side in storage, as we're now just a client of Shopify's JWTs. They're the authority on what a session is, we just need to proxy it through to Gadget.


packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -1,160 +1,150 @@
...
+    if (!appBridge) return;
+    // setup the api client to always query using the custom shopify auth implementation
+    api.connection.setAuthenticationMode({
+      custom: {
+        processFetch: async (_input, init) => {
+          const headers = new Headers(init.headers);
+          headers.append("Authorization", `ShopifySessionToken ${await getSessionToken(appBridge)}`);
+          init.headers ??= {};
+          headers.forEach(function (value, key) {
+            (init.headers as Record<string, string>)[key] = value;
```
LucasUnplugged, line 60: Is `init.headers` being used externally, after this mutation? I'm just curious because we don't seem to be using `init` again (after line 56).

LucasUnplugged (commented):
I can't comment on the auth flow itself, but this LGTM in general.

kristianpd:
FYI @rdraward, @jasong689 or @airhorns can help explain this big change for the docs!

kristianpd:
FYI @jasong689 when this ships, we'll need to solve the "do I need to re-oauth" problem. 

This should be a high priority for us next week, but I don't think it's going to block Showday approval.

https://linear.app/gadget-dev/issue/GGT-1803/nice-way-to-ask-for-new-scopes-from-react-land-if-granted-scopes-are

We may want to do it with GraphQL as @airhorns suggested, you can also look at how the CLI app does it as a reference in case it solves any unexpected issues:

```typescript

function checkHeadersForReauthorization(headers, app) {
  if (headers.get("X-Shopify-API-Request-Failure-Reauthorize") === "1") {
    const authUrlHeader =
      headers.get("X-Shopify-API-Request-Failure-Reauthorize-Url") ||
      `/api/auth`;

    const redirect = Redirect.create(app);
    redirect.dispatch(
      Redirect.Action.REMOTE,
      authUrlHeader.startsWith("/")
        ? `https://${window.location.host}${authUrlHeader}`
        : authUrlHeader
    );
  }
}

export default function returnTopLevelRedirection(req, res, redirectUrl) {
  const bearerPresent = req.headers.authorization?.match(/Bearer (.*)/);

  // If the request has a bearer token, the app is currently embedded, and must break out of the iframe to
  // re-authenticate
  if (bearerPresent) {
    res.status(403);
    res.header("X-Shopify-API-Request-Failure-Reauthorize", "1");
    res.header("X-Shopify-API-Request-Failure-Reauthorize-Url", redirectUrl);
    res.end();
  } else {
    res.redirect(redirectUrl);
  }
}
```

Depending on the level of control we want to give our users, this could either be wrapped up behind a `Provider` prop, or a separate flow. /cc @rdraward when we make our decision!

packages/react-shopify-app-bridge/src/Provider.tsx
airhorns: Yeah, it is, this is a kinda weird function that mutates the stuff we're about to pass to a `fetch` function. See https://github.com/gadget-inc/js-clients/blob/main/packages/api-client-core/src/GadgetConnection.ts#L250

jasong689 (approved):
Just a small fix for handling initial installs. Went ahead and 🎩 on the following:
- Firefox standard mode
- Firefox strict mode
- Firefox private
- Safari
- Safari private
- Chrome
- Chrome incognito
Both flows of an initial install and existing install were tested. We should merge https://github.com/gadget-inc/js-clients/pull/51 so that we have the updated api client core version

PR #53, opened 2022-07-12
Reauthenticate if scopes have changed

jasong689:
We recently added a `shopifyConnection` GraphQL query that allows the `Provider` to know whether or not we should re-auth with Shopify for updated scopes. This PR requests the `requiresReauthentication` field and forces a re-authentication if that returns true.

closes GGT-1850


kristianpd (approved):

airhorns (approved):
Makes sense to me!

PR #55, opened 2022-07-29
Change api-client-core to target es2019 instead of es2020

jasong689:
We ran into some issues with NextJs minification where it cannot correctly minify code that uses null coalescing. Let's compile down to `es2919` instead where null coalesce does not exist. Will bump deps for other packages once this one is merged

before:
Screen Shot 2022-07-29 at 12 42 30 PM

after:
Screen Shot 2022-07-29 at 12 41 57 PM


airhorns (approved):

airhorns:
😢 , thanks for figuring this out!

PR #56, opened 2022-07-29
Bump other packages to include updated api-client-core version

jasong689:
We need to bump package versions to update dep for `api-client-core`


airhorns:
Is this still needed?

jasong689:
nope, all good

PR #57, opened 2022-08-04
Dispose of the client with catch

angelini:
The client's dispose function can throw, let's handle it with a `.catch` to ensure it doesn't cause and "Uncaught promise rejection".


scott-rc (approved):
Looks good 👍

PR #58, opened 2022-08-04
Delegate connection timeout to graphql-ws

scott-rc:
This changes `waitForOpenedConnection` to use `graphql-ws`'s [`connectionAckWaitTimeout`](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/client.ClientOptions.md#connectionackwaittimeout), [`lazy: false`](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/client.ClientOptions.md#lazy), and [`onNonLazyError`](https://github.com/enisdenjo/graphql-ws/blob/master/docs/interfaces/client.ClientOptions.md#onnonlazyerror) to handle the connection timeout.

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -172,16 +172,15 @@ export class GadgetConnection {
       try {
         // The server will error if it receives any operations before the auth dance has been completed, so we block on that happening before sending our first operation. It's important that this happens synchronously after instantiating the client so we don't miss any messages
         subscriptionClient = await this.waitForOpenedConnection({
-          isFatalConnectionProblem(errorOrCloseEvent) {
-            // any interruption of the transaction is fatal to the transaction
-            console.warn("Transport error encountered during transaction processing", errorOrCloseEvent);
-            return true;
-          },
           connectionAckWaitTimeout: DEFAULT_CONN_ACK_TIMEOUT,
           ...options,
           // super ultra critical option that ensures graphql-ws doesn't automatically close the websocket connection when there are no outstanding operations. this is key so we can start a transaction then make mutations within it
           lazyCloseTimeout: 100000,
-          retryAttempts: 0,
+          // any interruption of the transaction is fatal to the transaction
+          shouldRetry(errorOrCloseEvent) {
+            console.warn("Transport error encountered during transaction processing", errorOrCloseEvent);
+            return false;
+          },
         });
 
         const client = new Client({
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -5,7 +5,6 @@ import { ExecutionResult } from "graphql";
 import {
   Client as SubscriptionClient,
   ClientOptions as SubscriptionClientOptions,
-  CloseCode,
   createClient as createSubscriptionClient,
   Sink,
 } from "graphql-ws";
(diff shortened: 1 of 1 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -368,60 +361,18 @@ export class GadgetConnection {
...
-      resetListeners();
-    }).finally(clearListeners);
+      subscriptionClient = this.newSubscriptionClient({
+        ...options,
+        lazy: false,
+        onNonLazyError: reject,
+      });
+
+      unsubscribe = subscriptionClient.on("connected", () => resolve(subscriptionClient));
+    }).finally(() => unsubscribe?.());
```
angelini, line 375: Is this `.` supposed to be here? Should it not be `unsubscribe?()`?

angelini:
So we were already setting `connectionAckWaitTimeout`, why would it be any different now? It didn't seem to stop after 10s with that set in production.

packages/api-client-core/src/GadgetConnection.ts
jasong689: I think the dot is required if you use `?`

scott-rc: Function calls still need the `.` -- it does look weird at first glance though 😅 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls

scott-rc:
> So we were already setting `connectionAckWaitTimeout`, why would it be any different now? It didn't seem to stop after 10s with that set in production.

Before, we didn't pass the `onNonLazyError` option to `newSubscriptionClient` so it used the default noop one we set [here](https://github.com/gadget-inc/js-clients/blob/main/packages/api-client-core/src/GadgetConnection.ts#L329).

I'm pretty sure that's why it wasn't stopping -- `graphql-ws` would timeout and throw during a non-lazy connection, which would be caught by the default noop handler.

Now, `newSubscriptionClient` will throw and we should end up [here](https://github.com/gadget-inc/js-clients/pull/58/files#diff-81f9f58a09da15eb852189a4964efa304d79e834a08f3b29430002c6524a9c9cR220).

scott-rc:
Gonna add @airhorns because this change is subtle and affects a lot of code paths.

jasong689 (commented):
It looks like we used to create a new client on `ConnectionAcknowledgementTimeout`, do you think that's still necessary? Aside from that, I think the changes with your comment makes sense. I'm not sure if we have tests for these situations in `gadget`. We could build and pack then test using the tgz

scott-rc:
> It looks like we used to create a new client on `ConnectionAcknowledgementTimeout`, do you think that's still necessary? Aside from that, I think the changes with your comment makes sense. I'm not sure if we have tests for these situations in `gadget`. We could build and pack then test using the tgz

I don't think so. The point of this change is to utilize `ConnectionAcknowledgementTimeout` as our timeout instead of having 2 different ones. I changed `DEFAULT_CONN_ACK_TIMEOUT` to `10_000` to match the previous timeout so the length _should_ be the same.

I'm down to test using the tgz if we think it's necessary.

airhorns:
Is this just a refactor or are we suspicious of this code? As you said Scott it is super sensitive and is in place because we've had lots of trouble with this before. The default way that graphql-ws and urql seem to want to work for subscriptions is to allow operations/messages to buffer up on the client while the connection is being established. That makes sense if you are really just subscribing to backend notifications, but in our case, we are layering on a bunch of backend stuff to the idea of a graphql-ws connection, like a backend transaction. That means that we don't want to start executing the function within `api.transaction` until we really know we have a transaction open, and we want to throw an error as soon as we know we can't connect. Know what I mean? 

One thing we should definitely do is write some tests. graphql-ws is well tested on its own but this is a really advanced usage of it and I am not sure we're really exercising this stuff we've layered in, so I think before shipping this change we should write some tests to prove how the old behaviour works. https://github.com/gadget-inc/js-clients/blob/main/packages/api-client-core/spec/GadgetConnection.spec.ts has some tests for straight HTTP calls but not much for a ws session.

scott-rc:
This is just a refactor.

You're right, let me add some tests. I've got some other stuff I want to get done before this so I'm gonna mark this as a draft and circle back later.

PR #59, opened 2022-08-06
Add a flag to let developers know that they're viewing a page request

jasong689:
I'm not entirely sure what to call this flag but it is there to indicate that the page has rendered because a user has visited the page directly via something like a link instead of the shopify admin (if the app is set as embedded). This is so that developers can add their own behavior i.e. redirect to app listing, ask for shop url to install on, or show some splash page.

### DIFF
--- a/packages/react-shopify-app-bridge/README.md
+++ b/packages/react-shopify-app-bridge/README.md
@@ -96,7 +96,7 @@ export default function App() {
 
 // An example component that uses the Gadget React hooks to work with data in the Shopify backend
 function ProductManager() {
-  const { loading, appBridge } = useGadget();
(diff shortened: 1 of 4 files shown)


airhorns:
So the specific situation is an app designed for embedding is being rendered in a non-embedded context? Maybe `isRootFrameRequest`?

airhorns (approved):
Code LGTM but how would a developer discover this flag exists? We need to document this somewhere, I think the readme would be fine for now

jasong689:
>Code LGTM but how would a developer discover this flag exists? We need to document this somewhere, I think the readme would be fine for now
Ah yep, will update README and look at adding some docs as well

PR #60, opened 2022-08-11
Fix support for standalone mode

jasong689:
When we switched over to using Shopify Session Tokens we broke support for standalone apps that still need to go through OAuth and render some sort of UI. This PR handles the following:
- skip using app bridge altogether if running in standalone mode
- skip requesting a session's shop since that relies on Shopify Session Tokens
- redirect to OAuth endpoint if we're in standalone mode and there is an hmac and shop param

I 🎩  the following flows:
- install a standalone app
- load a standalone app
- install an embedded app
- load an already installed embedded app
- load an embedded app in a non embedded context


jasong689:
@kristianpd bump

kristianpd (approved):

PR #61, opened 2022-08-16
[hooks] Return a Gadget-owned error wrapper object with easy access to validation errors from GraphQL responses

airhorns:
When a Gadget backend rejects an action because the supplied params are invalid, it returns an error in the format like:
```
{
  success: false,
  errors: {
    message: "GGT_INVALID_RECORD: Record is invalid. 2 error needs to be corrected",
    validationErrors: [
       { field: "title", message: "is too short" },
       { field: "body", message: "is required" }
    ]
}
```

The point of all that is to always return a top level error message regardless of what kind of error happened (be it like a backend error where all we have is a message) and then provide structured error details if we do have them (the by-field error messages).

This data is really useful for driving form validations where you want to show a user which fields are broken. They need to correct those fields before they can move on, so we need to make this data from the GraphQL response available in React land really easily. Before this change, the way you would get at it is like so:

```typescript
const [{data, fetching, error}, mutate] = useAction(api.widget.create);
const validationErrors = error?.graphQLErrors.?[0]?.originalError.validationErrors // any
```

which sucks IMO. The type of the `error` object returned before this change is an `urql` `CombinedError` object which wraps all the errors that can occur client side, like network errors and GraphQL execution errors. I think we should return our own type that works like this:

```typescript
const [{data, fetching, error}, mutate] = useAction(api.widget.create);
const validationErrors = error.validationErrors // FieldErrorMessage[] | undefined
```

This does just that! We introduce a new class called `ErrorWrapper` which is the `error` in all our React hook results, and give it nicer types for having these validation errors on them. `ErrorWrapper` was also a nice place to DRY up some of the repeated error handling logic across the hooks.

### DIFF
--- a/packages/react/src/utils.ts
+++ b/packages/react/src/utils.ts
@@ -1,0 +1,179 @@
+import { GadgetError, gadgetErrorFor, getNonNullableError, InvalidFieldError, InvalidRecordError } from "@gadgetinc/api-client-core";
+import type { CombinedError, OperationResult } from "@urql/core";
+import { GraphQLError } from "graphql";
+import { AnyVariables, Operation, OperationContext, UseQueryState } from "urql";
+
+/**
+ * The inner result object returned from a query result
+ **/
+export interface ReadHookState<Data = any, Variables = Record<string, any>> {
+  fetching: boolean;
+  stale: boolean;
+  data?: Data;
+  error?: ErrorWrapper;
+  extensions?: Record<string, any>;
+  operation?: Operation<Data, Variables>;
+}
+
+/**
+ * The return value of a `useGet`, `useFindMany`, `useFindOne` etc hook.
+ * Includes the data result object and a refetch function.
+ **/
+export declare type ReadHookResult<Data = any, Variables extends AnyVariables = AnyVariables> = [
+  ReadHookState<Data, Variables>,
+  (opts?: Partial<OperationContext>) => void
+];
+
+/**
+ * The inner result object returned from a mutation result
+ */
+export interface ActionHookState<Data = any, Variables = Record<string, any>> {
+  fetching: boolean;
+  stale: boolean;
+  data?: Data;
+  error?: ErrorWrapper;
+  extensions?: Record<string, any>;
+  operation?: Operation<Data, Variables>;
+}
+
+/**
+ * The return value of a `useAction`, `useGlobalAction`, `useBulkAction` etc hook.
+ * Includes the data result object and a function for running the mutation.
+ **/
+export declare type ActionHookResult<Data = any, Variables extends AnyVariables = AnyVariables> = [
+  ActionHookState<Data, Variables>,
+  (variables: Variables, context?: Partial<OperationContext>) => Promise<OperationResult<Data, Variables>>
+];
+
+const generateErrorMessage = (networkErr?: Error, graphQlErrs?: GraphQLError[]) => {
+  let error = "";
+  if (networkErr !== undefined) {
+    error = `[Network] ${networkErr.message}`;
+  } else if (graphQlErrs !== undefined) {
+    graphQlErrs.forEach((err) => {
+      error += `[GraphQL] ${err.message}\n`;
+    });
+  } else {
+    error = "Unknown error";
+  }
+
+  return error.trim();
+};
+
+const rehydrateGraphQlError = (error: any): GraphQLError => {
+  if (typeof error === "string") {
+    return new GraphQLError(error);
+  } else if (error?.message) {
+    return new GraphQLError(error.message, error.nodes, error.source, error.positions, error.path, error, error.extensions || {});
+  } else {
+    return error;
+  }
+};
+
+/**
+ * An error returned by any of the Gadget react hooks.
+ * Always has a message, but can be inspected to retrieve more detailed errors from either the network, the raw GraphQL layer, or Gadget specific errors like validation errors.
+ * Not intended for creating outside of Gadget-owned code.
+ **/
+export class ErrorWrapper extends Error {
+  /** @private */
+  static forClientSideError(error: Error, response?: any) {
+    return new ErrorWrapper({
+      executionErrors: [error],
+      response,
+    });
+  }
+  /** @private */
+  static forErrorsResponse(errors: Record<string, any>[], response?: any) {
+    return new ErrorWrapper({
+      executionErrors: errors.map(gadgetErrorFor),
+      response,
+    });
+  }
+  /** @private */
+  static forMaybeCombinedError(error?: CombinedError | null) {
+    if (!error) return undefined;
+    return new ErrorWrapper({
+      networkError: error.networkError,
+      executionErrors: error.graphQLErrors,
+      response: error.response,
+    });
+  }
+  /** @private */
+  static errorIfDataAbsent(result: UseQueryState<any>, dataPath: string[]) {
+    const nonNullableError = getNonNullableError(result, dataPath);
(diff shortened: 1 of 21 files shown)


packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+ **/
+export declare type ActionHookResult<Data = any, Variables = Record<string, any>> = [
+  ActionHookState<Data, Variables>,
+  (variables?: Variables, context?: Partial<OperationContext>) => Promise<OperationResult<Data, Variables>>
+];
+
+const generateErrorMessage = (networkErr?: Error, graphQlErrs?: GraphQLError[]) => {
+  let error = "";
+  if (networkErr !== undefined) {
+    return (error = `[Network] ${networkErr.message}`);
```
thegedge, line 51: nit: Did you intend to `return` here? If so, you can drop the `error =`, and perhaps move the `let` below to show it's only needed below.

Also, what if `networkErr` was `undefined`, and `graphQlErrs` was either `undefined` or an empty array? Would the intent be to still return an empty error string, or some default message?

packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+    });
+  }
+
+  return error.trim();
+};
+
+const rehydrateGraphQlError = (error: any): GraphQLError => {
+  if (typeof error === "string") {
+    return new GraphQLError(error);
+  } else if (typeof error === "object" && error.message) {
```
thegedge, line 66: `typeof null == "object"`, so I'd guard with `error?.message` (or some other thing to prevent accessing a prop on `null`).

packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+   * A list of errors the backend reported for specific fields being invalid for the records touched by an action. Is a shortcut for accessing the validation errors of a `GadgetInvalidRecordError` if that's what is in the `executionErrors`.
+   **/
+  public validationErrors?: InvalidFieldError[];
+  /**
+   * A list of errors encountered by the backend when processing a request. Populated if the client successfully communicated with the backend, but the backend was unable to process the request and rejected it with an error.
+   * Includes GraphQL syntax errors, missing or invalid argument errors, data validation errors, or unexpected errors encountered when running backend logic.
+   *
+   * This property allows this object to match the interface of urql's `CombinedError` object.
+   *
+   * @deprecated
```
thegedge, line 136: Not sure if there's a small message we could include with `@deprecated` to say what to use now, but I always like it when a lib specifies it :)

thegedge (approved):
So much nicer! 👏 

Just a few small things, but nothing blocking :)

packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+ * The inner result object returned from a query result
+ **/
+export interface ReadHookState<Data = any, Variables = Record<string, any>> {
+  fetching: boolean;
+  stale: boolean;
+  data?: Data;
+  error?: ErrorWrapper;
+  extensions?: Record<string, any>;
+  operation?: Operation<Data, Variables>;
+}
```
LucasUnplugged, line 16: It looks like your hands are tied for `ReadHookResult`, but for this interface, what do you think of just extending the `urql` type instead, to avoid hardcoding type defs for props you're not actually decorating here? Like so:
```suggestion
export interface ReadHookState<Data = any, Variables = Record<string, any>> extends UseQueryState<Data, Variables> {
  error?: ErrorWrapper
}
```

packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+ **/
+export declare type ReadHookResult<Data = any, Variables = Record<string, any>> = [
+  ReadHookState<Data, Variables>,
+  (opts?: Partial<OperationContext>) => void
+];
+
+/**
+ * The inner result object returned from a mutation result
+ */
+export interface ActionHookState<Data = any, Variables = Record<string, any>> {
```
LucasUnplugged, line 30: Same question here.

packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+   **/
+  public executionErrors: (GraphQLError | GadgetError)[];
+  /**
+   * An error encountered when trying to communicate with the backend from the client. Includes things like connection timeouts, connection interrupts, or no internet connection errors
+   **/
+  public networkError?: Error;
+  /**
+   * A list of errors the backend reported for specific fields being invalid for the records touched by an action. Is a shortcut for accessing the validation errors of a `GadgetInvalidRecordError` if that's what is in the `executionErrors`.
+   **/
+  public validationErrors?: InvalidFieldError[];
```
LucasUnplugged, line 129: There doesn't seem to be any code in this PR that actually sets this property, because it doesn't seem to be passed into any of the `new ErrorWrapper(...)` calls. Is this an accident, or was this more about thinking ahead to something we'll likely need?

packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+    const nonNullableError = getNonNullableError(result, dataPath);
+    let error = ErrorWrapper.forMaybeCombinedError(result.error);
+    if (!error && nonNullableError) {
+      error = ErrorWrapper.forClientSideError(nonNullableError);
+    }
+    return error;
+  }
+
+  /** Class name of this error -- always `ErrorWrapper` */
+  public name: "ErrorWrapper";
```
LucasUnplugged, line 114: Two Qs:
1. Couldn't you just do `name = "ErrorWrapper"` here, instead of only assigning a value in the constructor?
2. What do you think of replacing this with a `get name() { return "ErrorWrapper" }` method instead? Just that should make it publicly consumable _and_ immutable, with very little more code.

packages/react/src/utils.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.name = "ErrorWrapper";
+    this.message = message;
+    this.executionErrors = normalizedExecutionErrors;
+    this.graphQLErrors = normalizedExecutionErrors;
+    this.networkError = networkError;
+    this.validationErrors = validationErrors;
+    this.response = response;
+  }
+
+  toString() {
```
LucasUnplugged, line 170: Great call including this! Great error class in general — very easy to consume.

LucasUnplugged (approved):
```TS
const validationErrors = error?.graphQLErrors.?[0]?.originalError.validationErrors // any
```
Yeah, that's gross ☹️ 

Just some questions/suggestions, also nothing blocking.

packages/react/src/utils.ts
airhorns: I think that is cleaner from a reuse perspective but I find that these type definitions are in a way user facing code, in that when folks hit go to type definition on a thing, they get taken to this text exactly. For that reason I kinda felt like hardcoding it was nice to show it all in one place as opposed to asking the user to understand how we subtype and override urqls types, you know?

packages/react/src/utils.ts
airhorns: Ooops, you are right, I forgot to implement this actual piece! Thanks!

packages/react/src/utils.ts
airhorns: Can't do it the nice way cause TypeScript doesn't allow you to put statements before the `super` call in the constructor if you use this style. Is annoying.

packages/react/src/utils.ts
LucasUnplugged: Yeah, fair point. I don't imagine this would be at a high risk of changing, at least on a minor version.

PR #81, opened 2022-08-24
React docs

airhorns:
Adds api docs and some short examples for using the react hooks. Next step: embedding this same content into the Gadget docs at docs.gadget.dev . I find reviewing markdown kind of unpleasant, so here's a link to the rendered version: https://github.com/gadget-inc/js-clients/blob/docs/packages/react/README.md

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -23,7 +23,7 @@
 2. A **typed** and **hydrated** result based on any given `select` option
 
 This library wraps [urql](https://github.com/FormidableLabs/urql) to communicate with a Gadget-generated API, so it benefits from all of
-the same [features](https://github.com/FormidableLabs/urql/blob/main/README.md#-features) as urql.
+the same [features](https://github.com/FormidableLabs/urql/blob/main/README.md#-features) as `urql`.
 
 ## Installation
 
@@ -33,6 +33,9 @@ To install the JS client for your app, you must set up the Gadget NPM registry, 
 
 ```
 npm config set @gadget-client:registry https://registry.gadget.dev/npm
+
+# then
+
 yarn add @gadget-client/my-app-slug
 # or
 npm install @gadget-client/product-tagger-8000
@@ -48,20 +51,37 @@ yarn add @gadgetinc/react react
 npm install --save @gadgetinc/react react
 ```
 
+And finally, you must instantiate an instance of your API client, and then wrap your application in the `Provider` component exported from this library. In your root-most React component, you can add a `Provider` around the other components of your application:
+
+```javascript
+// import the API client for your specific application from your client package, see your app's installing instructions
+import { Client } from "@gadget-client/my-gadget-app";
+// import the required Provider object and some example hooks from this package
+import { Provider } from "@gadgetinc/react";
+
+// instantiate the API client for our app
+const api = new Client({ authenticationMode: { browserSession: true } });
+
+export const MyApp = (props) => {
+  // wrapp the application in the <Provider> so the hooks can find the current client
+  return <Provider value={api.connection.currentClient}>{props.children}</Provider>;
+};
+```
+
 ## Example usage
 
 ```typescript
+// import the API client for your specific application from your client package, see your app's installing instructions
 import { Client } from "@gadget-client/my-gadget-app";
+// import the required Provider object and some example hooks from this package
 import { Provider, useAction, useFindMany } from "@gadgetinc/react";
 import React from "react";
 
-const api = new Client({
-  authenticationMode: {
-    browserSession: true,
-  },
-});
+// instantiate the API client for our app
+const api = new Client({ authenticationMode: { browserSession: true, }});
 
 export function MyComponent() {
+  // ensure any components which use the @gadgetinc/react hooks are wrapped with the provider, and passed the current api client
   return (
     <Provider value={api.connection.currentClient}>
       <WidgetDeleter />
@@ -70,9 +90,7 @@ export function MyComponent() {
 }
 
 function WidgetDeleter() {
-  const [_, deleteWidget] = useAction(api.widget.delete);
-
-  // if id or _all weren't fields on a widget, this would be a type error
+  // `useFindMany` executes a backend fetch to get a list of widgets from the backend. we select only the id and name fields of each widget to demonstrate type-safe sub selection
   const [result, refresh] = useFindMany(api.widget, {
     select: {
       id: true,
@@ -80,11 +98,14 @@ function WidgetDeleter() {
     },
   });
 
+  // `useAction` sets up a mutation we can run to delete a specific widget when a user clicks a button
+  const [_, deleteWidget] = useAction(api.widget.delete);
+
+  // handle all the different states the data fetch can be in
   if (result.error) return <>Error: {result.error.toString()}</>;
   if (result.fetching && !result.data) return <>Fetching...</>;
   if (!result.data) return <>No widgets found</>;
 
-  // If id/name weren't selected above, they wouldn't typecheck below
   return (
     <>
       {result.data.map((widget) => (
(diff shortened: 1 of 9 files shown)


packages/react/README.md
```diff
@@ -48,20 +51,34 @@ yarn add @gadgetinc/react react
 npm install --save @gadgetinc/react react
 ```
 
+And finally, you must wrap your application in the `Provider` component exported from this library. In your root-most React component, you can add a `Provider` around the other components of your application:
+
+```javascript
+import { Provider } from "@gadgetinc/react";
+
+export const MyApp = (props) => {
+  return <Provider>{props.children}</Provider>;
```
thegedge, line 60: We directly export the urql `<Provider>`, which needs a client. What do you think about showing it here, and excluding it from all other examples? Or can keep it in all other examples, but would need to provide `value={client}` here.

packages/react/README.md
```diff
@@ -101,3 +119,593 @@ function WidgetDeleter() {
...
+
+### `useFindMany(manager: ModelFinder, options: ManyFinderOptions = {}): [{data, fetching, error}, refetch]`
+
+`useFindMany` fetches a page of records from your Gadget database, optionally sorted, filtered, or searched. `useFindMany(api.widget, { ... })` is the React equivalent of `await api.widget.findMany({ ... })`. Options:
+
+- `manager`: The model manager you for the model you want to find a page of records for. The value always comes from an instance of your application's API client instantiated elsewhere. Required. Example: `api.widget`, or `api.shopifyProduct`
+- `options`: Options for making the call to the backend. Not required.
+  - `select`: A list of fields and subfields to select. See the [Select option](#the-select-option) docs. Optional.
+  - `filter`: A list of filters to limit the set of returned records down to. Optional. See the Model Filtering section in your application's API documentation to see the available filters for your models.
```
thegedge, line 196: nit: here and any other filter explanation, I'd drop the "down to", since it's redundant.

Another thought: would be really neat to be able to link this to the "Model Filtering" section of some model in the latest/som app you visited on the docs. Not sure if that'd be useful though 🤔

packages/react/README.md
```diff
@@ -101,3 +119,593 @@ function WidgetDeleter() {
...
+
+```javascript
+import React from "react";
+import { api } from "../api"; // some file that instantiates @gadget-client/your-app-slug
+
+export const ShowWidgetNames = () => {
+  const [{ data, fetching, error }, _refetch] = useFindMany(api.widget);
+
+  if (fetching) return "Loading...";
+  if (error) return `Error loading data: ${error}`;
```
thegedge, line 222: Not sure if, for all the examples, you want to deal with `!data`, which is typical when `pause: true`.

packages/react/README.md
```diff
@@ -101,3 +119,593 @@ function WidgetDeleter() {
...
+```
+
+The `refetch` function returned as the second element can be executed in order to trigger a refetch of the most up to date data from the backend. See [`urql`'s docs](https://formidable.com/open-source/urql/docs/basics/react-preact/#reexecuting-queries) on re-executing queries for more information.
+
+### `useFindBy(findFunction: ModelFindFunction, fieldValue: any, options: FindByOptions = {}): [{data, fetching, error}, refetch]`
+
+`useFindBy` fetches one record from your backend Gadget, looked up by a specific field and value. `useFindBy` uses the existing specific-field finder functions generated within your application's API client, like `api.user.findByEmail` or similar. `useFindBy` expects to be passed a field-specific finder function as the first argument, and then the value of the field to find by as the second argument, and then optionally accepts a third options object. If no record is found matching the conditions, then the returned object will have `null` for the `data`. `useFindBy(api.widget.findByEmail, "hi@gadget.dev")` is the React equivalent of `api.widget.findByEmail("hi@gadget.dev")`. Options:
```
thegedge, line 388: Do we want to call out here that these are only generated in your client for fields that have a uniqueness validation?

packages/react/README.md
```diff
@@ -101,3 +119,593 @@ function WidgetDeleter() {
...
+        Purge data
+      </button>
+    );
+  } else {
+    return "Purge completed";
+  }
+};
+```
+
+### `useGet(singletonModelManager: SingletonModelManager, options: GetOptions = {}): [{data, fetching, error}, refetch]`
```
thegedge, line 585: Just a mental note I'm gonna poke away: if we never really have any other singletons, we should perhaps consider more useful hooks like `useCurrentSession` and `useCurrentUser`.

packages/react/README.md
```diff
@@ -101,3 +119,593 @@ function WidgetDeleter() {
...
+- server side errors where the client sent information to the server but the server failed to process it due to a bug or transient issue.
+
+Each of these error cases is broken out on the `error` object returned by `useAction` (and any of the other hooks). The `error` object is an `ErrorWrapper` object, which has a number of properties for figuring out exactly what went wrong:
+
+- `error.message: string` - A top level error message which is always present
+- `error.networkError: Error | undefined` - An error thrown by the browser when trying to communicate with the server
+- `error.executionErrors: (GraphQLError | GadgetError)[] | undefined` - Any errors thrown by the GraphQL API, like missing parameters or invalid selections, and any errors thrown by the server concerning invalid data or backend processing errors.
+- `error.validationErrors: { apiIdentifier: string, message: string }[] | undefined` - Any validation errors returned by the server. A shortcut to accessing the `.validationErrors` property of the first `InvalidRecordError` in the `.executionErrors` of the outer `ErrorWrapper` object. Useful for building form validations.
+
+#### Default selections
```
thegedge, line 709: `###` instead of `####`? This doesn't look like a subsection of "Errors from the returned `error` object"

thegedge (approved):
🚀

packages/react/README.md
```diff
@@ -101,3 +122,595 @@ function WidgetDeleter() {
...
+
+`@gadgetinc/react` contains a variety of React hooks for working with your Gadget application's API from a React application. Specific code examples for your application's API can be found within your application's docs at https://docs.gadget.dev/api/<your-application-slug>
+
+### Setup
+
+Your React application must be wrapped in the `Provider` component from this library for the hooks to function properly. No other wrappers (like `urql`'s) are necessary.
+
+### `useFindOne(manager: ModelFinder, id: string, options: SingleFinderOptions = {}): [{data, fetching, error}, refetch]`
+
+`useFindOne` fetches one record from your Gadget database with a given `id`. `useFindOne(api.widget, 42)` is the React equivalent of `await api.widget.findOne(42)`. `useFindOne` expects a record with the given id to be found in the backend database, and will return an error in the `error` property of the result object if no record is found with the given `id`. `useFindOne` takes these options:
```
rdraward, line 138: When viewing the actual README, this paragraph is hard to get through. Starting sentences with `useFindOne` and ending with `id.` makes it really tough to read (the periods make it seem like we're accessing another property at first glance).

`id`. `useFindOne(api.widget, 42)`
`await api.widget.findOne(42)`. `useFindOne`
`id`. `useFindOne`

Just adding breaks might be the easiest way to get around this:

```suggestion
`useFindOne` fetches one record from your Gadget database with a given `id`.
`useFindOne(api.widget, 42)` is the React equivalent of `await api.widget.findOne(42)`.
`useFindOne` expects a record with the given id to be found in the backend database, and will return an error in the `error` property of the result object if no record is found with the given `id`.
`useFindOne` takes these options:
```

packages/react/README.md
```diff
@@ -101,3 +122,595 @@ function WidgetDeleter() {
...
+
+- `manager`: The model manager you for the model you want to find a record of. The value always comes from an instance of your application's API client instantiated elsewhere. Required. Example: `api.widget`, or `api.shopifyProduct`
+- `id`: The backend id of the record you want to find. Required.
+- `options`: Options for making the call to the backend. Not required.
+  - `select`: A list of fields and subfields to select. See the [Select option](#the-select-option) docs. Optional.
```
rdraward, line 143: Do we need Optional here? If so, should we label all sub-not required fields as optional?

packages/react/README.md
```diff
@@ -101,3 +122,595 @@ function WidgetDeleter() {
...
+
+  return (
+    <span className="widget-name">
+      {data?.id}: {data?.name}
+    </span>
+  );
+};
+```
+
+The `refetch` function returned as the second element can be executed in order to trigger a refetch of the most up to date data from the backend. See [`urql`'s docs](https://formidable.com/open-source/urql/docs/basics/react-preact/#reexecuting-queries) on re-executing queries for more information.
```
rdraward, line 186:
```suggestion
The `refetch` function returned as the second element can be executed in order to trigger a refetch of the most up-to-date data from the backend. See [`urql`'s docs](https://formidable.com/open-source/urql/docs/basics/react-preact/#reexecuting-queries) on re-executing queries for more information.
```

packages/react/README.md
```diff
@@ -101,3 +122,595 @@ function WidgetDeleter() {
...
+
+### `useMaybeFindOne(manager: ModelFinder, id: string, options: SingleFinderOptions = {}): [{data, fetching, error}, refetch]`
+
+`useMaybeFindOne` fetches one record from your Gadget database with a given `id`. `useMaybeFindOne(api.widget, 42)` is the React equivalent of `await api.widget.maybeFindOne(42)`. `useMaybeFindOne` will return `data: null` and `error: null` if no record with the given `id` is found in the backend database. `useMaybeFindOne` otherwise behaves identically to `useFindOne`, and accepts the same options.
```
rdraward, line 190: Same comment as above for the intro paragraph

packages/react/README.md
```diff
@@ -101,3 +122,595 @@ function WidgetDeleter() {
...
+  first: 10,
+  after: "some-cursor-value",
+});
+
+// the current page's start and end cursor are available for use to then make later requests for different pages
+data.endCursor; // used for forward pagination, pass to the `after:` variable
+data.startCursor; // used for backwards pagination, pass to the `before:` variable
+```
+
+An easy way to do pagination is using React state, or for a better user experience, using the URL with whatever router system works for your application. We use React state to demonstrate pagination in this example:
```
rdraward, line 304: 👏 great build-up to the example

packages/react/README.md
```diff
@@ -101,3 +122,595 @@ function WidgetDeleter() {
...
+```javascript
+// fetch only the widget id and name fields
+const [{ data, fetching, error }, _refetch] = useFindOne(api.widget, props.id, { select: { id: true, name: true } });
+```
+
+The `refetch` function returned as the second element can be executed in order to trigger a refetch of the most up to date data from the backend. See [`urql`'s docs](https://formidable.com/open-source/urql/docs/basics/react-preact/#reexecuting-queries) on re-executing queries for more information.
+
+### `useFindBy(findFunction: ModelFindFunction, fieldValue: any, options: FindByOptions = {}): [{data, fetching, error}, refetch]`
+
+`useFindBy` fetches one record from your backend looked up by a specific field and value. `useFindBy` uses the existing specific-field-finder functions generated within your application's API client, like `api.user.findByEmail` or similar. These functions are generated for Gadget fields that have Unique Validations added to them.
```
rdraward, line 391: It could be a good idea to link to Gadget docs for how to add these kinds of actions/functions

rdraward (approved):
Great content @airhorns!

My only concern is that we are using `code blocks` for everything and it gets a bit tough to read in some places:
CleanShot 2022-08-29 at 12 52 32@2x

That said, I don't think that should block us from merging - we can do a separate clean up/format at a later time

PR #91, opened 2022-09-09
Export useQuery and useMutation from react package

jasong689:
We currently export the urql Provider from our react package so let's do the same for `useQuery` and `useMutation`. This allows users to use the same version urql that our react package uses and not rely on a peer dependency should they need custom queries or mutations. We use the provider and `useQuery` in `@gadgetinc/react-shopify-app-bridge`

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -753,3 +753,77 @@ export const ShowWidgetNames = () => {
   );
 };
 ```
+
+### `urql` exports
+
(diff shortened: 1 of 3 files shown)


airhorns:
Into it but can you add docs / examples to the readme?

jasong689:
> Into it but can you add docs / examples to the readme?

Added a separate section in the docs related to the urql exports and a sample code snippet

thegedge (approved):
Game for this 🚀

airhorns (approved):

PR #96, opened 2022-09-14
Runtime react hook tests

airhorns:
I am working on improving the react hooks for building web forms and wanted some real tests that actually exercise the hooks! So, this wires that up with `@testing-library/react`'s `renderHook` function and a mock urql client copied in from the Gadget repo. They work pretty good!


packages/api-client-core/package.json
```diff
@@ -24,7 +24,7 @@
     "@urql/exchange-multipart-fetch": "^1.0.0",
     "cross-fetch": "^3.0.6",
     "gql-query-builder": "^3.6.0",
-    "graphql": "^15.5.0",
+    "graphql": "^16.5.0",
```
airhorns, line 27: We should be using the newer version, urql is compatible with it and it has some bug fixes. I changed it because it aligns with other stuff so we don't have type clashes.

packages/react/src/useFindBy.ts
```diff
@@ -79,12 +79,12 @@ export const useFindBy = <
...
-    const nonNullableError = getNonNullableError(result, dataPath);
-
     if (records.length > 1) {
       error = ErrorWrapper.forClientSideError(getNonUniqueDataError(finder.modelApiIdentifier, finder.findByVariableName, value));
-    } else if (nonNullableError) {
-      error = ErrorWrapper.forClientSideError(nonNullableError);
+    } else if (result.data && !records[0]) {
+      error = ErrorWrapper.forClientSideError(
+        new GadgetNotFoundError(`${finder.modelApiIdentifier} record with ${finder.findByVariableName}=${value} not found`)
+      );
```
airhorns, line 87: We documented `useFindBy` as throwing if the record isn't found for the field value, but it didn't actually do that. I made it do it... I think we want that?

packages/react/src/useGet.ts
```diff
@@ -62,9 +62,10 @@ export const useGet = <
 
   const [result, refresh] = useQuery({ query: plan.query, variables: plan.variables });
 
-  let data = result.data;
-  if (data) {
-    data = hydrateRecord(result, get(result.data, [manager.get.operationName]));
+  let data = null;
+  const rawRecord = result.data && get(result.data, [manager.get.operationName]);
+  if (rawRecord) {
+    data = hydrateRecord(result, rawRecord);
```
airhorns, line 68: Same bug with returning a `new GadgetRecord({})` in the error case

packages/react/src/useFindBy.ts
thegedge: Would we also want something like a `useMaybeFindBy`? Just wondering if throwing forces people into something that doesn't fit the way they're building their react apps (e.g., Suspense / `ErrorBoundary` / preferred `error` value)

PR #100, opened 2022-09-18
Wrap `useAction` promise results

airhorns:
We wrap urql's `useMutation` hook to return a fancy hydrated value with our own error class and our own hydrated data. This worked for the `const [result] = useAction` bit, but wasn't applied to the `const [_, run] = useAction(); const result = await run()` part.

This fixes that woop woop!

### DIFF
--- a/packages/react/src/useAction.ts
+++ b/packages/react/src/useAction.ts
@@ -9,11 +9,11 @@ import {
   LimitToKnownKeys,
   Select,
 } from "@gadgetinc/api-client-core";
-import { useMemo } from "react";
-import { useMutation } from "urql";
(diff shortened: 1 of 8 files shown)


packages/react/src/useAction.ts
```diff
@@ -77,6 +77,28 @@ export const useAction = <
     F["variablesType"]
   >(plan.query);
 
+  return [
+    processResult(result, action),
+    useCallback(
```
thegedge, line 82: Good call on `useCallback`. We don't know how people will use this, so giving this function a stable identity is the better default.

thegedge (approved):
Nice catch 👏

PR #101, opened 2022-09-19
Add a `currentChanges` to `GadgetRecord`

LucasUnplugged:
- Improved GadgetRecord performance, by greatly minimizing the number of object-to-array conversions, and avoiding generating all changes just to see if there is one change.
- Added `currentChanges` method to get the current value of changed fields, rather than both current AND previous values in a nested object.
- Added a new test for this feature.

### DIFF
--- a/packages/api-client-core/spec/GadgetRecord.spec.ts
+++ b/packages/api-client-core/spec/GadgetRecord.spec.ts
@@ -138,6 +138,23 @@ describe("GadgetRecord", () => {
     });
   });
 
+  it("should allow you to ask for CURRENT values of changed props on the entire object", () => {
+    const product = new GadgetRecord<SampleBaseRecord>(productBaseRecord);
+    expectNoChanges(product, "name", "body", "count");
+
+    product.name = "A newer name";
+    product.count = 123;
+
+    expect(product.changed("name")).toEqual(true);
+    expect(product.changed("count")).toEqual(true);
+    expect(product.changed("body")).toEqual(false);
+    expect(product.changed()).toEqual(true);
+    expect(product.toChangedJSON()).toEqual({
+      name: "A newer name",
+      count: 123,
+    });
+  });
+
   it("should allow dirty tracking on array fields", () => {
     const product = new GadgetRecord<SampleBaseRecord>({ ...productBaseRecord, anArray: [1, 2, 3] });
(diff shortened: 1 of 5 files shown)


packages/api-client-core/src/GadgetRecord.ts
```diff
@@ -98,7 +137,7 @@ export class GadgetRecordImplementation<Shape extends RecordShape> {
   changed(prop?: string | ChangeTracking, tracking = ChangeTracking.SinceLoaded) {
     return prop && typeof prop == "string"
       ? this.changes(prop, tracking).changed
-      : Object.keys(this.changes(prop === undefined ? tracking : (prop as ChangeTracking))).length > 0;
+      : this.hasChanges(prop === undefined ? tracking : (prop as ChangeTracking));
```
LucasUnplugged, line 140: @airhorns my implementation has a tradeoff of using more memory, but that cost seemed small, compared to the cost of:
1. Converting the fields object to an array.
2. Looping through all fields to generate a change object.
3. Converting that object to an array again.

I did a quick test of the `GadgetRecord.spec` test, and that was almost 10% faster. Considering the testing overhead would be the same for both, the performance diff should be even bigger.

Let me know if you don't think it's worth doing though, and I can revert this change.

packages/api-client-core/spec/GadgetRecord.spec.ts
```diff
@@ -138,6 +138,23 @@ describe("GadgetRecord", () => {
...
+    expectNoChanges(product, "name", "body", "count");
+
+    product.name = "A newer name";
+    product.count = 123;
+
+    expect(product.changed("name")).toEqual(true);
+    expect(product.changed("count")).toEqual(true);
+    expect(product.changed("body")).toEqual(false);
+    expect(product.changed()).toEqual(true);
+    expect(product.currentChanges()).toEqual({
```
airhorns, line 152: I feel like the two functions `.currentChanges()` and `.changes()` don't exactly explain what is different about each other just looking at the names. What do you think about `.changes()` and `.changedJSON()` (or maybe `.toChangedJSON()`)?

airhorns (approved):
LGTM! The extra tracker complexity is a bit noisy but it sounds far more performant so 🚢

PR #110, opened 2022-10-04
Include more information in unexpected close errors

thegedge:
We're seeing a few close events here and there, but it's hard to know what's happening because there's no reason. I figure we include the code in the error/message too, which I'm hoping will shed a lot more light on the specific cause of these close events.

Not easy code to test, but I'll gitpkg a release and try some manual QA.


thegedge:
🎩 ✅ Running with no issues on one of my test apps.

angelini (approved):

PR #118, opened 2022-10-24
bump @gadgetinc/react version

jasong689:
We didn't update `@gadgetinc/react` package version after merging #112

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.8.1",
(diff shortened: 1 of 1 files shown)


thegedge (approved):
Oopsies!

jasong689:
Want to ensure it's bumped since I'm updating the dependency version in `@gadgetinc/react-shopify-app-bridge` and noticed it hadn't been updated

thegedge:
Looks like we deleted one of the two apps we were using for CI, hence the failures. We'll see how we can get this reinstated.

PR #130, opened 2022-11-24
Default urql's requestPolicy to cache-and-network so that users see changed server side data by default

airhorns:
We enable urql's handy dandy client side caching features because they work great and allow for snappy rerenders of components where data hasn't changed. We left it to `urql` to chose the default `requestPolicy`, and their default is `cache-only`. This request policy means that a `refresh()` call made in the browser won't actually refresh by default, you have to pass `refresh({requestPolicy: "network-only"})` to force a network call to be made. I think this default is pretty confusing for users who didn't turn on caching themselves, but I also think the caching sets up a good user experience, so I think we should turn off the full-client-side caching mode.

I also verified that this request policy doesn't really have any effect on imperatively calls made where you do `api.something.findOne()`, only really on react hooks where the provider is set up to provide a cache.

### DIFF
--- a/packages/api-client-core/spec/GadgetConnection.spec.ts
+++ b/packages/api-client-core/spec/GadgetConnection.spec.ts
@@ -27,6 +27,14 @@ describe("GadgetConnection", () => {
     expect(connection.authenticationMode).toEqual(AuthenticationMode.Anonymous);
   });
 
+  it("should default to the cache-and-network request policy for urql clients", () => {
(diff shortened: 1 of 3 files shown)


rdraward (approved):
A couple of nitpicky comments in the documentation, but looks great!
Thanks for tackling this so quickly @airhorns

LucasUnplugged (approved):
Bizarre default on their part, when `cache-and-network` is available. This LGTM!

PR #133, opened 2022-11-29
Add return explicit return type to `toJSON`

scott-rc:
This adds an explicit return type to our `toJSON` methods instead of the inferred `any` type.

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -127,7 +127,7 @@ export const internalCreateMutation = (apiIdentifier: string) => {
   const capitalizedApiIdentifier = capitalize(apiIdentifier);
   return `
     ${internalErrorsDetails}
-    
+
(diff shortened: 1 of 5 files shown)


packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -220,7 +220,7 @@ export class InternalModelManager {
     const response = await this.connection.currentClient.query(internalFindOneQuery(this.apiIdentifier), { id }).toPromise();
     const assertSuccess = throwOnEmptyData ? assertOperationSuccess : assertNullableOperationSuccess;
     const result = assertSuccess(response, ["internal", this.apiIdentifier]);
-    return await hydrateRecord(response, result);
+    return hydrateRecord(response, result);
```
scott-rc, line 223: ESLint started [complaining](https://github.com/gadget-inc/js-clients/actions/runs/3577183140/jobs/6015887356) about these awaits after I added those return types 😕

PR #134, opened 2022-11-30
Update storage token to be keyed by endpoint

jasong689:
We noticed that session tokens are stored at the static `token` key and this ends up being re-used across different Gadget apps if you're building multiple embedded apps

closes GGT-2915

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -100,6 +100,10 @@ export class GadgetConnection {
     this.baseClient = this.newBaseClient();
   }
 
+  private get sessionStorageKey() {
+    return `${sessionStorageKey}-${this.endpoint}`;
+  }
+
   get currentClient() {
     return this.currentTransaction?.client || this.baseClient;
   }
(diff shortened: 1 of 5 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -100,6 +100,10 @@ export class GadgetConnection {
     this.baseClient = this.newBaseClient();
   }
 
+  private get sessionStorageKey() {
+    return `${sessionStorageKey}-${this.endpoint}`;
```
airhorns, line 104: This'll break anyone logged in with a current session when this ships for people. I think it wouldn't be too hard to check both the old deprecated key and also the new key for a little bit during the rollout, but only set it to the new key?

jasong689: Hm yea, I was thinking it's more convenient for anyone with one already stored at `token` to refetch a new session than having to understand that if there's one at `token` they'll get 401s until they clear localStorage when building embedded apps. Although that might be a smaller pool of users

airhorns: Talked IRL -- ok with logging people out in order to overcome having to clear localstorage for folks working on multiple apps locally.

airhorns (approved):

kira-bruneau (approved):

PR #136, opened 2022-12-01
Don't use `localStorage` for auth if it isn't available

airhorns:
All our docs tell folks to construct their API client with no explicit `authenticationMode` for simplicity, and so that we can make an educated decision about what authentication mode to use internally based on the environment. In Node, we use anonymous authentication, and in the browser, we try to use a cookie-like authentication that works cross-domain with localStorage. 

`localStorage` however errors in some browsers if you try to touch in when the browser window is incognito, or it just doesn't work in private browsing in safari. Sad. This breaks apps trying to get through Shopify's review process, as Shopify requires that you can log in/out of the app in incognito. We support this just fine, you just have to (before this change) ensure you do `new Client({authenticationMode: { anonymous: true }})` or similar, which our docs don't tell you to do rn.

So, I think our default authentication mode for browser contexts that don't have localStorage shouldn't just be "shit the bed", it should fall back to the temporary storage. For Shopify apps, this temporary auth method gets replaced by the provider later, so what you pass doesn't even matter, it just gets replaced. For this reason, I think falling back to something is a fine default.

### DIFF
--- a/packages/api-client-core/spec/GadgetConnection-suite.ts
+++ b/packages/api-client-core/spec/GadgetConnection-suite.ts
@@ -1,13 +1,12 @@
 import gql from "gql-tag";
 import nock from "nock";
-import { BrowserSessionStorageType } from "../src";
-import { AuthenticationMode, GadgetConnection } from "../src/GadgetConnection";
+import { AuthenticationMode, BrowserSessionStorageType, GadgetConnection } from "../src";
+import { base64 } from "./helpers";
 
-const base64 = (str: string) => Buffer.from(str).toString("base64");
 nock.disableNetConnect();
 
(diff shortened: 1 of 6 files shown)


airhorns:
cc @rdraward does this make good sense to you?

jasong689 (approved):
These changes make sense, but I almost wonder if it makes sense to have Durable store be opt in and just default to anonymous. I don't see any issues with the code changes

kristianpd (approved):
bring back nock `https` and :shipit:!

rdraward:
> cc @rdraward does this make good sense to you?

Makes sense to me @airhorns! Thanks for the quick fix!

PR #141, opened 2022-12-07
Allow case insensitive environment to be specified and default to NODE_ENV

jasong689:
Update the gadget connection to use `NODE_ENV` value if one is not explicitly passed through the options. We'll also allow developers to use any case they'd like to specify he env as long as it matches "development" or "production" when lower cased.

## PR Checklist


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -69,7 +72,7 @@ export class GadgetConnection {
   private websocketsEndpoint: string;
   private websocketImplementation: any;
   private _fetchImplementation: typeof fetch;
-  private environment: "Development" | "Production";
```
jasong689, line 72: this is used for `getCurrentSpan()?.setAttributes({ applicationId: this.options.applicationId, environmentName: this.environment });`. not sure if sure if there are concerns that the case here changes

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -90,7 +93,14 @@ export class GadgetConnection {
     this.websocketImplementation = options.websocketImplementation ?? WebSocket;
     this.websocketsEndpoint = options.websocketsEndpoint ?? options.endpoint + "/batch";
     this.websocketsEndpoint = this.websocketsEndpoint.replace(/^http/, "ws");
-    this.environment = options.environment ?? "Development";
+    const environment = options.environment ?? process.env["NODE_ENV"] ?? "development";
+    const parsedEnvironment = this.parseEnvironment(environment);
+    if (parsedEnvironment) {
+      this.environment = parsedEnvironment;
+    } else {
+      console.warn("Invalid environment", environment, "defaulting to development");
```
jasong689, line 101: Is this sufficient to warn users that they've provided an invalid environment?

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -32,14 +32,17 @@ export const $transaction = Symbol.for("gadget/transaction");
...
+
 export interface GadgetConnectionOptions {
   endpoint: string;
   authenticationMode?: AuthenticationModeOptions;
   websocketsEndpoint?: string;
   subscriptionClientOptions?: GadgetSubscriptionClientOptions;
   websocketImplementation?: any;
   fetchImplementation?: typeof fetch;
-  environment?: "Development" | "Production";
+  environment?: string;
```
jasong689, line 45: String is coming

PR #157, opened 2023-03-08
Adds a specific error for timeouts opening a websocket connection to …

infiton:
Adds a dedicated error for timeouts opening a websocket connection to the gadget API, and marks these errors as not caused by the client.

## PR Checklist

- [x] Any immediate changes are slated for release in Gadget via a generated package dependency bump

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.12.1",
+  "version": "0.13.0",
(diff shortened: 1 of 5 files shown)


airhorns (approved):
Is this just for better visibility on these errors? If so LGTM

infiton:
> Is this just for better visibility on these errors

and allows us to mark these errors as not caused by the client

thegedge (approved):
Makes sense. Aligns with the `GadgetUnexpectedCloseError` too, in terms of being `causedByClient = false`.

PR #160, opened 2023-03-13
Minor versions of react and react-shopify-app-bridge packages

jasong689:
We released `0.13.0` of `api-client-core` and updated it in Gagdet's userland deps. This means new clients are generated with `^0.13.0` dependency and no longer resolves to the same version listed under `react` and `react-shopify-app-bridge` (`^0.12.0`). Since we're `<1.0.0`, semver `^0.12.0` only matches the patch version and not the minor version.

Tested locally by installing the current version of client and `react-shopify-app-bridge` which results in the error. Then building a local version with `0.13.0` as the dependency and the error goes away. I updated minor versions for these since `api-client-core` was updated a minor version as well.

closes GGT-3639

## PR Checklist

### DIFF
--- a/packages/react-shopify-app-bridge/package.json
+++ b/packages/react-shopify-app-bridge/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react-shopify-app-bridge",
-  "version": "0.8.0",
+  "version": "0.9.0",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
@@ -37,7 +37,7 @@
     "react-dom": "^18.2.0"
   },
   "peerDependencies": {
-    "@gadgetinc/react": "^0.10.0",
+    "@gadgetinc/react": "^0.11.0",
     "@shopify/app-bridge-react": "^2.0.0 || ^3.0.0",
     "react": "^17.0.2 || ^18.0.0",
     "react-dom": "^17.0.2 || ^18.0.0"
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.10.0",
+  "version": "0.11.0",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",


infiton (approved):
damn; sorry about that

I can take this as an action to document and make a test to make sure this doesn't happen again

jasong689:
> I can take this as an action to document and make a test to make sure this doesn't happen again
That makes sense. Did you already have a place in mind for documentation? We can perhaps add a test Gadget side to ensure `api-client-core` in the generated client matches the dependency so `react`

infiton:
> Did you already have a place in mind for documentation?

thought a slab doc; but open to ideas

airhorns (approved):
Yup makes sense to me! I thought we actually already had a test for this in place, can you poke around this repo and look Kyle? It must be not testing what we think it is testing. I think @scott-rc maybe set it up

scott-rc:
> Yup makes sense to me! I thought we actually already had a test for this in place, can you poke around this repo and look Kyle? It must be not testing what we think it is testing. I think @scott-rc maybe set it up

Wow... so @infiton did update the `api-client-core` dependency [here](https://github.com/gadget-inc/js-clients/commit/24176175550fd89d7d73c27b8d0754633e83ffc3#diff-02b759e0dd8e7c74e7cf4ae0ebc3115281ebe6766855ead3e477dc7a76e6dca6L22-R22) and [here](https://github.com/gadget-inc/js-clients/commit/24176175550fd89d7d73c27b8d0754633e83ffc3#diff-1f344ac391eeecc21ec0f01fb07430a47f4b80d20485c125447d54c33c4bbfc4L22-R22) for the react packages (which is why our [version checking script](https://github.com/gadget-inc/js-clients/blob/main/scripts/check-dependency-versions.ts) passed), but he didn't update the react packages' own versions, so they weren't published to npm with their updated `api-client-core` dependency 😅 

I'm not sure how to catch this from happening in the future 🤔

jasong689:
@scott-rc I think what we can do is have a test Gadget side which ensures that whatever `api-client-core` version clients are generated with are resolved to the same versions as those in `react` and `react-shopify-app-bridge`

jasong689:
Right that'd only catch when we update `api-client-core` with a breaking change (minor version upgrades count right now) and forgot to publish `react` or `react-shopify-app-bridge`. We do already catch when we haven't updated the dependencies. I don't realistically think there's a reason to update dependencies but not publish the package, so maybe we should catch it here somehow.

PR #164, opened 2023-03-27
Retry connection init wait close error

angelini:
The retry version of: https://github.com/gadget-inc/gadget/pull/6140


scott-rc:
Should we set `connectionInitWaitTimeout` like we do the `connectionAckWaitTimeout`?

https://github.com/gadget-inc/js-clients/pull/164/files#diff-81f9f58a09da15eb852189a4964efa304d79e834a08f3b29430002c6524a9c9cR199

angelini:
`connectionInitWaitTimeout` is a server side option, so we can only set it within Gadget

PR #166, opened 2023-03-30
Change default provider's loading value to true

alanko0511:
Since OAuth and any other requests that set this `loading` value are executed almost instantly, setting the default value to false will make components that rely on this value flashes in the first render.

## PR Checklist


danroberts (approved):

jasong689 (approved):
Might be good to update the default value in `context` as well

PR #170, opened 2023-04-17
Ensure the react hooks return objects with the same identity if nothing changes on rerender

airhorns:
urql takes great care to ensure that if your component rerenders and calls `useQuery` again, but nothing about the query has changed, you get back the same object. This is so that if you use that object in a dependencies array, it's identity is stable and react won't expire memos or re-run useEffects.

Prior to this change, we didn't keep this same property. If you wrote code like this:

```javascript
const [{ data, fetching, error}] = useFindFirst(api.someModel);
const [something, setSomething] = useState(null)

useEffect(() => {
    setSomething(data)
}, [data]);
```

You'd accidentally get an infinite loop because the useEffect would trigger a rerender because it called setstate, which would rerun our react hook, which would return new data, which would be passed to the array of dependencies for use effect, which would trigger it to rerun, etc.

Instead, let's keep the same property that urql has, which is that you only get a new result object when something about the result changes.

### DIFF
--- a/packages/react/src/useGlobalAction.ts
+++ b/packages/react/src/useGlobalAction.ts
@@ -34,8 +34,10 @@ export const useGlobalAction = <F extends GlobalActionFunction<any>>(
 
   const [result, runMutation] = useGadgetMutation<any, F["variablesType"]>(plan.query);
 
+  const transformedResult = useMemo(() => processResult(result, action), [result, action]);
+
   return [
-    processResult(result, action),
+    transformedResult,
     useCallback(
       async (variables, context) => {
         const result = await runMutation(variables, context);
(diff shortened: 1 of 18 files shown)


packages/react/src/useGlobalAction.ts
```diff
@@ -53,6 +55,7 @@ const processResult = (result: UseMutationState<any, any>, action: GlobalActionF
     data = get(result.data, [action.operationName]);
     if (data) {
       const errors = data.errors;
+      data = data.result;
```
jasong689, line 58: Is this change related to keeping the same identity?

jasong689 (commented):
Change makes sense, just one question.

packages/react/src/useGlobalAction.ts
airhorns: You've got laser eyes! Nope, this was extra, I noticed that we didn't actually return the right value for global actions at all. We returned the whole data object , like `{success: true, errors: [...], result: <something>}` instead of just the `<something>`. I was gonna land this as semver major for this reason, tho I doubt that many people are using `useGlobalAction` with custom results from the backend.

thegedge (approved):
Oh, nice! Makes a lot of sense to `useMemo` the results here (I'm assuming urql does the same for us).

PR #172, opened 2023-04-19
Add a `useFetch` hook for making requests to backend app routes

airhorns:
We've noticed that lots of folks want to make requests to their backend HTTP routes from Gadget from their frontends. Not everything flows through the GraphQL API. Requests from the same browser to either the GraphQL API or a backend route should share the same identity though -- they are coming from the same person in the same browser! But, lots of developers accidentally fall into the trap of using our api client and hooks for making requests to GraphQL, then just doing normal fetches with `window.fetch` for their routes, and losing their session / identity.

We document and encourage folks to use the `api.fetch` call instead of the `window.fetch` call to solve this -- our fetch is what our GraphQL calls use under the hood, and so passes along the right auth information.

The next step is encouraging the same people to do the same in React. `api.fetch` is imperative async code similar to `api.widget.findMany`, we need the equivalent of `useFindMany` for React fetching. Enter `useFetch`!

This adds a `useFetch` hook that closely resembles our other hooks. It returns a very similar set of stuff to our other hooks, namely a `[{data, error, fetching}, reexecute]` tuple. 

Also, FWIW, I tried out a few different open source useFetch hooks, but couldn't find one that really suited our use case. `swc` and `use-http` are the two production grade ones I'm aware of, but they include a lot of bells and whistles that I think make sense if you're fetching all your data with fetch, like caching and all sorts of other goodies. Accordingly their bundle sizes are like 8kb. For us, we're only fetching the occasional bit of data, and we re-use a lot of what the connection has already to add those goodies, so I don't really want to duplicate all of it between our stuff and those libs stuff. I also couldn't find a good quality hook that was way simpler, so, I took a bit to make a good one for ourselves and add some good tests.


airhorns:
cc @rdraward @DevAOC if y'all have any input on the design of this hook, see the readme changes for the shape

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -320,12 +321,14 @@ export class GadgetConnection {
...
     );
 
-    return new Client({
+    const client = new Client({
       url: this.endpoint,
       fetch: this.fetch,
       exchanges,
       requestPolicy: this.requestPolicy,
     });
+    (client as any).gadgetConnection = this;
```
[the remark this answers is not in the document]
thegedge: How would you feel about using a symbol as the property key?

packages/react/src/useFetch.ts
```diff
@@ -0,0 +1,146 @@
...
+      return data;
+    },
+    [memoizedOptions, path]
+  );
+
+  useEffect(() => {
+    mounted.current = true;
+    if (startRequestByDefault(memoizedOptions)) {
+      void send().catch(() => {
+        // report error via return state
```
thegedge, line 136: As per the comment, did we want to `dispatch({ type: "error", payload: … });` here?

packages/react/src/useFetch.ts
```diff
@@ -0,0 +1,146 @@
...
+      });
+    }
+
+    return () => {
+      mounted.current = false;
```
thegedge, line 141: Does the fetcher support an `AbortSignal` param? Would it be worth adding one?

packages/react/src/useFetch.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export type FetchHookResult<T> = [FetchHookState<T>, (opts?: Partial<FetchHookOptions>) => Promise<T>];
+
+type FetchAction<T> = { type: "fetching" } | { type: "fetched"; payload: T } | { type: "error"; payload: ErrorWrapper };
+
+const reducer = <T>(state: FetchHookState<T>, action: FetchAction<T>): FetchHookState<T> => {
+  switch (action.type) {
+    case "fetching":
+      return { ...state, fetching: true, error: undefined };
```
thegedge, line 21: If `state.fetching`, should we return state instead of an object with a new identity? Similarly for others (but may need to check all of the params).

packages/react/src/useFetch.ts
```diff
@@ -0,0 +1,146 @@
...
+  const send = useCallback(
+    async (sendOptions?: Partial<FetchHookOptions>): Promise<T> => {
+      dispatch({ type: "fetching" });
+
+      let data: any;
+      let response: Response | undefined = undefined;
+
+      try {
+        // make the fetch call using GadgetConnection to pass along auth and other headers
+        response = await connection.fetch(path, { ...memoizedOptions, ...sendOptions });
```
thegedge, line 104: If `memoizedOptions?.json` is `true`, should we send `Accept: application/json` to give better guarantees about having a JSON response?

packages/react/src/useFetch.ts
```diff
@@ -0,0 +1,146 @@
...
+ *   });
+ *
+ *   if (result.error) return <>Error: {result.error.toString()}</>;
+ *   if (result.fetching && !result.data) return <>Fetching...</>;
+ *   if (!result.data) return <>No user found with id={props.id}</>;
+ *
+ *   return <div>{result.data.name}</div>;
+ * }
+ */
+export function useFetch(path: string, options: { stream: true } & FetchHookOptions): FetchHookResult<ReadableStream<Uint8Array>>;
```
thegedge, line 78: Curious: instead of building our own thing, do you think it would be better to build on top of something like https://use-http.com/?

packages/react/src/useFetch.ts
airhorns: Nope not needed, the `send` function sets its own error in the `catch` clause and will cause react to rerender

packages/react/src/useFetch.ts
airhorns: Good call, added!

packages/react/src/useFetch.ts
airhorns: I would love to not have to maintain this ourselves, but I looked at a couple of the options and I found they were bulky in terms of bundle size, and had fanciful APIs that we don't really need. Didn't go into enough detail in the PR description but `use-http` returns this big object you can then use to call something with any verb and kind of use for both reads and writes to your backend via REST API -- I think that's just kind of overkill for us. We want to encourage use of the GraphQL API for everything you can use it for, and this is more for tiny little things that you can kind of forget about.

packages/react/src/useFetch.ts
airhorns: 🤷 I think its ok to return an object with a new identity if a dispatch has happened -- they only really happen when something momentous really does happen. It's not perfectly optimal but it makes this code a lot simpler?

packages/react/src/useFetch.ts
airhorns: It does for sure! End users can pass in their own abort signal if they want to. I figured it was a lot of complexity for us to support both them passing a signal and us managing a signal internally, but I can do it if it seems super necessary. Was hoping to get something out but I think in the new AI-y world aborting streaming responses makes a lot of sense.

packages/react/src/useFetch.ts
thegedge: Up to you! The dispatch works similar to some of the other state stuff in that same identity won't cause a re-render. From https://react.dev/reference/react/useReducer#dispatch.

> If the new value you provide is identical to the current state, as determined by an [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will skip re-rendering the component and its children.

packages/react/src/useFetch.ts
thegedge: I'm game for holding off and seeing how it goes. I'm gonna guess the typical use case would be to abort on unmount, so if we see that happening or hear our users mention that, we can bake it in 🚀

packages/react/src/useFetch.ts
thegedge: I'm a-o-k going with what we have if you've dug in some. I think the main thing is if we're happy with the API we can always return and build on another library.

I think another option is to make it easy for others to use whatever fetch library they prefer, if they want to. To do that we could expose some of the stuff our own `connection.fetch` does, eh? No rush on that though.

thegedge (approved):
🚀

danroberts (approved):
My only complaint here which shouldn't block the PR is that I think it's likely people will get confused by the genericness of `useFetch` and could easily attempt to query non-gadget endpoints like useFetch("https://my-other-thing") and get pretty confused. 

Some other names I could imagine could be `useRoute` or `useGadgetFetch`

PR #174, opened 2023-04-19
fix(frontends): Prevent flash of unloaded content

danroberts:
We initialize the GadgetProvider 'loading' parameter to 'false', and then immediately start a load. This results in a microsecond where the 'isAuthorized' logic will report false (because it is not loading and not true), which we use to show a different UI when an embedded app is accessed outside of a shopify embed.

This results is a flash of the message 'This application is only viewable in the shopify admin' when first loading a page inside the shopify admin.


danroberts:
duplicate of https://github.com/gadget-inc/js-clients/pull/166

PR #175, opened 2023-04-19
Release v12.1 with fix to GadgetProvider loading default value

danroberts:
Releasing a new version of the react GadgetProvider which is initially set loading = true


jasong689 (changes requested):
@danroberts i think this is wrong package. should be `react-shopify-app-bridge`

jasong689 (approved):

alanko0511 (approved):

PR #176, opened 2023-04-20
Exclude the `__typename` field from GadgetRecord.toJSON()

airhorns:
Riley noticed a discord user trying to pass `record.toJSON()` back as an input to a different action call. This didn't work out of the box because the `.toJSON()` includes all the fields that came back from loading the record, including `__typename`. When sending this back at the GraphQL API, the `__typename` attribute causes GraphQL to get mad as you are passing fields which don't exist as input fields.


jcao49 (approved):
🚀💯

kristianpd (approved):

PR #178, opened 2023-04-21
Make passing in the params by model api identifier optional

jasong689:
This PR makes it optional to wrap your model params for create/update in the model api identifier. The only exception is if there is a field with the same api identifier as the model, then the params must be wrapped in the model api identifier; this is because there's no way to tell if the consumer is passing in model params or field value.

It would be nice to be able to test that the correct query/variables are called but the mutations rely on transactions via ws. This is a bit difficult to mock I think without adding some interfaces so that we can create a mock transaction, not sure if that was worth adding here.

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -5,6 +5,7 @@ import { GadgetRecordList } from "./GadgetRecordList";
 import type { GadgetTransaction } from "./GadgetTransaction";
 import {
   GadgetClientError,
+  GadgetOperationError,
   assert,
   assertMutationSuccess,
   assertNullableOperationSuccess,
@@ -239,9 +240,36 @@ export class InternalModelManager {
   constructor(
     private readonly apiIdentifier: string,
     readonly connection: GadgetConnection,
-    readonly options?: { pluralApiIdentifier: string }
+    readonly options?: { pluralApiIdentifier: string; hasAmbiguousIdentifiers?: boolean }
   ) {
     this.capitalizedApiIdentifier = camelize(apiIdentifier);
+  }
+
+  private validateRecord(record: RecordData) {
+    if (!this.options || !this.options.hasAmbiguousIdentifiers) {
+      return true;
+    }
+    const keys = Object.keys(record);
+
+    // any keys that aren't the apiIdentifier are invalid
+    return keys.every((key) => key === this.apiIdentifier);
+  }
+
+  private getRecordFromData(record: RecordData, functionName: string) {
+    let recordData = record;
+
+    if (!this.validateRecord(record)) {
+      throw new GadgetOperationError(
+        `Invalid arguments found in variables. Did you mean to use ${functionName}({ ${this.apiIdentifier}: { ... } })?`,
+        "GGT_INVALID_RECORD_DATA"
+      );
+    }
+
+    if (this.apiIdentifier in record) {
+      recordData = recordData[this.apiIdentifier];
+    }
+
+    return recordData;
   }
 
   async findOne(id: string, throwOnEmptyData = true): Promise<GadgetRecord<RecordData>> {
@@ -292,7 +320,7 @@ export class InternalModelManager {
     return await this.transaction(async (transaction) => {
       const response = await transaction.client
         .mutation(internalCreateMutation(this.apiIdentifier), {
-          record: record[this.apiIdentifier],
+          record: this.getRecordFromData(record, "create"),
         })
         .toPromise();
       const result = assertMutationSuccess(response, ["internal", `create${this.capitalizedApiIdentifier}`]);
@@ -323,7 +351,7 @@ export class InternalModelManager {
       const response = await transaction.client
         .mutation(internalUpdateMutation(this.apiIdentifier), {
           id,
-          record: record[this.apiIdentifier],
+          record: this.getRecordFromData(record, "update"),
         })
         .toPromise();
       const result = assertMutationSuccess(response, ["internal", `update${this.capitalizedApiIdentifier}`]);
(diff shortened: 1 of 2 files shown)


thegedge:
Nice. This deals with the internal API. Do we have plans to do something similar for the public API?

jasong689:
We certainly do! https://github.com/gadget-inc/gadget/pull/6455

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -216,6 +216,15 @@ export class InternalModelManager {
     this.capitalizedApiIdentifier = camelize(apiIdentifier);
   }
 
+  private getRecordFromData(record: RecordData) {
+    let recordData = record;
+    if (this.apiIdentifier in record) {
```
thegedge, line 221: > The only exception is if there is a field with the same api identifier as the model, then the params must be wrapped in the model api identifier; this is because there's no way to tell if the consumer is passing in model params or field value.

If we wanted one heuristic, we could check to see if there's only a single key that's the API identifier and is an object. It's also not perfect, in the case of JSON fields, but I think could be an added safety measure.

If we wanted to prevent any issues we could make it a problem to have a field with the same API identifier as its model. Worth doing that? If so, this PR or another?

jasong689: I think that makes sense considering there aren't any exported params for the internal api

airhorns: I think that a model with a field that has the same name will actually be a somewhat regularly occurring thing -- I feel like people do that all the time where they make an `Image` model with an `image` File field or what have you. I think its a bit overbearing to say that is a problem and that they should change their design because we have an ambiguity elsewhere. Does the heuristic still work if we don't make that a problem?

thegedge: > Does the heuristic still work if we don't make that a problem?

It doesn't in the case that I mentioned where the field is a JSON type, then we do have the potential for ambiguity (which I should have clarified is what I envisioned the problem could be scoped to).

I suspect this case is probably rare enough that a problem wouldn't be overbearing. In most cases it would probably end up in a failed API call because the nested schema likely doesn't match the model's schema, but here's the rare case I'm thinking of where both styles work:

```js
// `Event` model with fields:
//   - event: JSON (optional)
//   - name: String (optional)
//   - type: String (default = "unknown")

await api.internal.create({ event: { type: "test" });
await api.internal.create({ event: { event: { type: "test" } });
```

jasong689:
poke @airhorns @thegedge. when you get a chance 👀

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -239,11 +239,35 @@ export class InternalModelManager {
...
+
+    // any keys that aren't the apiIdentifier are invalid
+    return !keys.some((key) => key !== this.apiIdentifier);
+  }
+
+  private getRecordFromData(record: RecordData, functionName: string) {
+    let recordData = record;
+
+    if (!this.validateRecord(record)) {
+      throw Error(`Invalid arguments found in variables. Did you mean to use ${functionName}({ ${this.apiIdentifier}: { ... } })?`);
```
thegedge, line 261: Probably best to throw a Gadget-y error, eh?

https://github.com/gadget-inc/js-clients/blob/main/packages/api-client-core/src/support.ts

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -239,11 +239,35 @@ export class InternalModelManager {
...
+    }
+
+    if (this.apiIdentifier in record) {
+      recordData = recordData[this.apiIdentifier];
```
thegedge, line 265: 👍 looks like we don't have any codepath that could have extra params in the internal API right now, but we'll have to remember that if that changes in the future this would potentially break (since it would clobber those params). Very likely we'd catch that kind of change in CI.

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -239,11 +239,35 @@ export class InternalModelManager {
   constructor(
     private readonly apiIdentifier: string,
     readonly connection: GadgetConnection,
-    readonly options?: { pluralApiIdentifier: string }
+    readonly options?: { pluralApiIdentifier: string; hasAmbiguousIdentifiers: boolean }
```
airhorns, line 242: This is why this change won't break when new clients are generated that use old api-client-core versions right?

airhorns (approved):
LGTM!

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -239,11 +239,35 @@ export class InternalModelManager {
...
   }
 
+  private validateRecord(record: RecordData) {
+    if (!this.options || !this.options.hasAmbiguousIdentifiers) {
+      return true;
+    }
+    const keys = Object.keys(record);
+
+    // any keys that aren't the apiIdentifier are invalid
+    return !keys.some((key) => key !== this.apiIdentifier);
```
thegedge, line 254: nit: I find it takes me a bit to reason about multiple negations ("valid if not some key is not the api identifier"). What about:

```js
return keys.every((key) => key === this.apiIdentifier);
```

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -239,11 +239,35 @@ export class InternalModelManager {
...
     private readonly apiIdentifier: string,
     readonly connection: GadgetConnection,
-    readonly options?: { pluralApiIdentifier: string }
+    readonly options?: { pluralApiIdentifier: string; hasAmbiguousIdentifiers: boolean }
   ) {
     this.capitalizedApiIdentifier = camelize(apiIdentifier);
   }
 
+  private validateRecord(record: RecordData) {
+    if (!this.options || !this.options.hasAmbiguousIdentifiers) {
```
thegedge, line 248: 🙌

packages/api-client-core/src/InternalModelManager.ts
jasong689: yep!

thegedge (approved):

PR #180, opened 2023-04-26
remove filter never from select and add deep filter never to fix hand…

jasong689:
Added a test to verify that you can still select the nested circular props. Also to note is that with `DeepFilter` you can no longer pass in a schema that has an optional property. This is fine with Gadget schemas as these are typed as `prop: type | null` instead of `prop?: type`

We will probably want to bump a minor version for this so that this change doesn't get picked up by older client packages

cc @angelini 

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/types.ts
+++ b/packages/api-client-core/src/types.ts
@@ -36,6 +36,27 @@ export type NonNeverKeys<Selection> = {
  */
 export type FilterNever<T extends Record<string, unknown>> = NonNeverKeys<T> extends never ? never : { [Key in NonNeverKeys<T>]: T[Key] };
 
+type InnerSelect<Schema, Selection extends FieldSelection | null | undefined> = Selection extends null | undefined
+  ? never
+  : Schema extends (infer T)[]
+  ? Select<T, Selection>[]
+  : Schema extends null
+  ? Select<Exclude<Schema, null>, Selection> | null
+  : {
+      [Key in keyof Selection & keyof Schema]: Selection[Key] extends true
+        ? Schema[Key]
+        : Selection[Key] extends FieldSelection
+        ? Select<Schema[Key], Selection[Key]>
+        : never;
+    };
+
+/**
+ * Filter out any keys in `T` that are mapped to `never` recursively.
+ */
+export type DeepFilterNever<T> = T extends Record<string, unknown> ? FilterNever<{
+  [Key in keyof T]: T[Key] extends Record<string, unknown> ? DeepFilterNever<T[Key]> : T[Key];
+}> : T;
+
 /**
  * Extract a subset of a schema given a selection
  *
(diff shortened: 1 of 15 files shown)


angelini:
Just a few notes re deployment:
- will need to increase the version of the package
- the ShipIt tool to update the version in everyone's environment currently requires too much disk space, we need to update the job to cleanup after itself.

jasong689:
@angelini we won't need to rebuild everyone's client for this. just new ones being generated going forward

packages/api-client-core/src/types.ts
```diff
@@ -52,10 +59,10 @@ export type Select<Schema, Selection extends FieldSelection | null | undefined>
   ? Select<T, Selection>[]
   : Schema extends null
   ? Select<Exclude<Schema, null>, Selection> | null
-  : FilterNever<{
```
thegedge, line 55: If we remove this, there is a window of time where the generated API client for an app (potentially) uses this version of `api-client-core` but doesn't `DeepFilterNever` the return.

I think maybe we take a different approach: what if we give the current `Select` a new name, say `InnerSelect`, and instead do this:

```ts
export type Select<Schema, Selection extends FieldSelection | null | undefined> = DeepFilterNever<InnerSelect<Schema, Selection>>
```

which doesn't require a client rebuild and would only need a minor version bump (since it's backwards compatible).

Also, since the `Select` above has no constraints on the `Schema` type, we either need to consider adding the `extends Record<string, unknown>` constraint or making `DeepFilterNever` not need it, e.g.

```ts
export type DeepFilterNever<T> = T extends Record<string, unknown>
  ? FilterNever<{
      [Key in keyof T]:
         T[Key] extends Record<string, unknown>
         ? DeepFilterNever<T[Key]>
         : T[Key];
    }>
  : T
}>;
```

I think putting the constraint on `Schema` should be fine since a schema is never a literal / array / non-object thingy.

What do you think?

thegedge: And note, by "potentially uses", I think it would require the users to manually specify an `api-client-core` that would result in a `yarn install` grabbing the new version. Unlikely, but it _could_ happen. If there's any concern there, we can double check existing `package.json`s to see if this would happen.

jasong689: @thegedge that makes sense. One other thing that @airhorns brought up was that we don't `yarn install` on every rebuild so even though we've updated their `api-client-core` version the old version is the one that's imported until someone runs `yarn install`. So I think in addition to your suggested change, I'm going to re-implement `Select` in the same way in our codegen and remove it once all packages have been updated.

packages/api-client-core/src/types.ts
```diff
@@ -36,26 +36,35 @@ export type NonNeverKeys<Selection> = {
...
       [Key in keyof Selection & keyof Schema]: Selection[Key] extends true
         ? Schema[Key]
         : Selection[Key] extends FieldSelection
         ? Select<Schema[Key], Selection[Key]>
         : never;
-    }>;
+    };
+
+/**
+ * Filter out any keys in `T` that are mapped to `never` recursively.
```
thegedge, line 54: Might be worth expanding a bit by calling out that objects, after having their `never`-valued keys filtered out, also get filtered out of their parent if they are empty. i.e.,

```ts
type A = DeepFilterNever<{ a: { b: never }, c: "test" }>;
type B = { c: "test" };  // same as A
type C = { a: {}, c: "test" };  // not this one though
```

Suggesting that because an empty object is distinct from `never`. Worth a double check that it's doing this.

thegedge (approved):
Super happy this worked out! 🚀

packages/api-client-core/src/types.ts
jasong689: Added an example in the comment and a test to make sure it's the case.

jasong689:
Failing tests look be a real issue. Will address

airhorns (approved):

packages/react/package.json
```diff
@@ -32,7 +32,8 @@
     "graphql": "16.5.0",
     "lodash": "^4.17.21",
     "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react-dom": "^18.2.0",
+    "typescript": "5.0.4"
```
jasong689, line 36: Added `typescript` as a dependency to this package only because we need to keep the root typescript pinned to `4.7.4` to keep prettier happy. Prettier does work if you recreate `yarn.lock` but that breaks the build so this is the easiest way to keep it working

PR #182, opened 2023-05-01
v0.13.6

jasong689:
Since the change we made to `Select` and addition of `DeepFilterNever` is backwards compatible, we can make this a patch update instead of a breaking minor version bump.

## PR Checklist


airhorns (approved):
Woop woop!

PR #189, opened 2023-05-04
Bump api-client-core to 0.13.8

jcao49:
Reverts gadget-inc/js-clients#188
Forgot to add the bang !


packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -212,7 +235,11 @@ export type RecordData = Record<string, any>;
 export class InternalModelManager {
   private readonly capitalizedApiIdentifier: string;
 
-  constructor(private readonly apiIdentifier: string, readonly connection: GadgetConnection) {
+  constructor(
+    private readonly apiIdentifier: string,
+    readonly connection: GadgetConnection,
+    readonly options: { pluralApiIdentifier: string }
```
jasong689, line 241: Could we make this optional and check in `bulkCreate` if it's set? Then you can throw an error (maybe `GadgetClientError`) when calling `bulkCreate` without a `pluralApiIdentifier`.

PR #193, opened 2023-05-09
Release 0.13.8

jcao49:
We need to publish `0.13.8` and the previous PR that tried to do that failed on lint.

Lint was fixed by Gao [here](https://github.com/gadget-inc/js-clients/commit/923cdae0bf2ee5b8963866be51de75f364ae168f) , but now we just have a version on main of api-client-core that is `0.13.8` but unpublished to npmjs.com (since the workflow didn't detect any changes in the `package.json`)

With this merge the .github workflow will see changes in `package.json`, causing the release workflow to run and actually publish `0.13.8`.

npmjs.com reports latest as 0.13.7 https://www.npmjs.com/package/@gadgetinc/api-client-core
## PR Checklist


jasong689 (approved):
This will also release the touch changes from @airhorns but I think that's ok

PR #194, opened 2023-05-09
fix(js-clients) - GGT-3592 - Remove deprecated and unsupported packages

ngc:
Removes the exchange-multipart-fetch package we used that would throw distracting warnings when creating a new Gadget app 

(now supported in urql natively)

## PR Checklist


packages/react-shopify-app-bridge/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react-shopify-app-bridge",
-  "version": "0.10.2",
+  "version": "1.10.2",
```
jasong689, line 3: I would probably just bump this to `0.11.2`. Minor versions > 1.0.0 are pretty much the same as major version changes. Were you able to test this locally? I think I had tried to do this at some point as was running into some problems with Shopify embedded apps.

ngc:
Still getting `shopify-app-bridge` type errors in CI. Going to break this into two PRs.

PR #195, opened 2023-05-09
fix(api-client-core) - GGT-2123 - Model api identifiers with underscores fail to run CRUD API

ngc:
(ex. hello_world => HelloWorld)

Came back to this ancient bug where model api identifiers with underscores fail to run CRUD API. 
This was because the behavior of the capitalize function was inconsistent with how we name our GraphQL types. 

internally this solves GGT-2123

## PR Checklist

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.13.9",
+  "version": "0.14.0",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.14.0",
+  "version": "0.13.10",
   "source": "src/index.ts",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
--- a/packages/api-client-core/src/support.ts
+++ b/packages/api-client-core/src/support.ts
@@ -211,10 +211,9 @@ export const capitalizeIdentifier = (str: string | undefined | null): string => 
 const capitalizeFirstCharacter = (str: string) => {
   const result = str === null || str === undefined ? "" : String(str);
   return result.charAt(0).toUpperCase() + result.slice(1);
-}
+};
 
 export const camelize = (term: string, uppercaseFirstLetter = true) => {
-
   let result = "" + term;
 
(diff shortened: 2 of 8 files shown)


airhorns:
Can you give me a little more info? This is technically a breaking change, and so I want to understand better if this will change existing queries that might work today to start being broken as well as fixing broken queries to start working.

ngc:
@airhorns 
Do you know of a good strategy I can use to test these potentially breaking changes?

A user was trying to name their apiIdentifiers using snake case (i.e. `my_model`)
The schema generator would successfully generate our GraphQL types for this model ex. `createMyModel` but `api-client-core` would mistakenly try to utilize `createMy_model`. 

This PR ensures the capitalization used in `api-client-core` matches that of the capitalization of apiIdentifiers in the SchemaGenerator so we don't fail to run mutations when using the api client.

airhorns (approved):
LGTM!

danroberts:
Given that we use `camelize` from `inflected` to generate the mutation names in GenerateGraphqlSchema, is there any reason we don't use it here to -- seems likely to generate additional bugs down the road. @airhorns

airhorns:
The main reason we don't use it here is to avoid increasing the bundle size by importing all of inflected for these tiny little utils. I think the move is probs to copy the utilities that we do need into this package, and then using these ones gadget side instead of inflected's?

packages/api-client-core/src/support.ts
```diff
@@ -203,12 +203,18 @@ export const get = (object: Record<string, any> | null | undefined, path: string
...
-  const result = str === null || str === undefined ? "" : String(str);
-  return result.charAt(0).toUpperCase() + result.slice(1);
+export const capitalize = (str: string | undefined | null): string => {
+  if (typeof str !== "string") return "";
+  return camelize(str, true);
 };
 
 export const camelize = (term: string, uppercaseFirstLetter = true) => {
+
+  const capitalize = (str: string) => {
```
airhorns, line 213: Its gonna be slightly faster to not redefine this function every time the outer function executes, can you move it out to the module level?

packages/api-client-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.13.9",
```
airhorns, line 3: Can you do only a patch version bump instead of a minor version? I think this should be viewed as a bugfix which should just increase the patch version

danroberts:
I think we should be able to do import { camelize } from 'inflected' and shake out the rest of the code we don't need. 

+1 to consuming this definition in the gadget package tho.

danroberts (approved):
👍 after validating that this produces the same results against a copy of all the api identifiers in the database.

PR #197, opened 2023-05-12
Ensure that `api.fetch` doesn't mangle query params when used with relative urls

airhorns:
Fixes a bug where `useFetch("/some/path?foo=bar")` would send the included query params URL encoded.


danroberts (approved):
👍

PR #198, opened 2023-05-12
useFetch `sendImmediately` option

airhorns:
While working on a form that submitted data to a GET endpoint I found it annoying that `useFetch` insisted on issuing the request as soon as the component renders. I think that is the right default behaviour (kinda like how `useFindMany` starts finding right away), but should be opt-out-able. This adds a new `sendImmediately: false` option to the `useFetch` hook allowing disabling this automatic-send behavior for GET requests.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -80,7 +80,7 @@ import { Provider, useAction, useFindMany } from "@gadgetinc/react";
 import React from "react";
 
 // instantiate the API client for our app
-const api = new Client({ authenticationMode: { browserSession: true, }});
+const api = new Client({ authenticationMode: { browserSession: true } });
 
 export function MyComponent() {
   // ensure any components which use the @gadgetinc/react hooks are wrapped with the provider, and passed the current api client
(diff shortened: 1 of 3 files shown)


packages/react/README.md
```diff
@@ -763,6 +764,48 @@ export function DummyProducts(props) {
...
+  const [{ data, fetching, error }, resend] = useFetch("/products");
+  // fetching will be `true`
+}
+```
+
+`useFetch` will _not_ immediately issue HTTP requests for HTTP verbs other than `GET`, like `POST`, `PUT`, etc. The HTTP request will only be sent when you call the returned `send` function.
+
+```javascript
+export function PostRequest(props) {
+  // will not automatically send the request when the component renders -- call the `send` function to run the request
```
thegedge, line 783: super nit: "call the `send` function to run the request" is a bit redundant with the comment below about fetching being false. Either way, maybe we have the comments for this example and the delayed get below be exactly the same?

packages/react/spec/useFetch.spec.ts
```diff
@@ -169,7 +169,28 @@ describe("useFetch", () => {
     expect(reexecuteResult).toEqual("fixed");
   });
 
-  test("it doesn't automatically start sending POST requests", async () => {
+  test("it automatically starts sending requests with no method specified", async () => {
```
thegedge, line 172: Maybe I missed it, but it look like there's no test to ensure it's a GET request when no method is specified. Wonder if we should add one / expectations?

thegedge (approved):
Definitely a great option to expose! 🚀

packages/react/spec/useFetch.spec.ts
airhorns: It is a good point -- right now we're actually inheriting that behavior from `fetch` itself which I think is the right move. I will add a test to ensure we pass no method and let it pick the default.

PR #201, opened 2023-05-17
support optional model api identifier in react hooks

jasong689:
We need to update the react hooks to also allow optional model api identifiers. This follows the same logic that the internal and public model managers.

closes GGT-4093

## PR Checklist

### DIFF
--- a/packages/react/src/useAction.ts
+++ b/packages/react/src/useAction.ts
@@ -79,9 +79,43 @@ export const useAction = <
     transformedResult,
     useCallback(
       async (variables, context) => {
+        if (action.hasAmbiguousIdentifier) {
+          if (Object.keys(variables).some((key) => !action.paramOnlyVariables?.includes(key) && key !== action.modelApiIdentifier)) {
+            throw Error(`Invalid arguments found in variables. Did you mean to use ({ ${action.modelApiIdentifier}: { ... } })?`);
+          }
+        }
+
+        let newVariables: Exclude<F["variablesType"], null | undefined>;
+        if (action.hasCreateOrUpdateEffect) {
+          if (
+            action.modelApiIdentifier in variables &&
+            typeof variables[action.modelApiIdentifier] === "object" &&
+            variables[action.modelApiIdentifier] !== null
+          ) {
+            newVariables = variables;
+          } else {
+            newVariables = {
+              [action.modelApiIdentifier]: {},
+            } as Exclude<F["variablesType"], null | undefined>;
+            for (const [key, value] of Object.entries(variables)) {
+              if (action.paramOnlyVariables?.includes(key)) {
+                newVariables[key] = value;
+              } else {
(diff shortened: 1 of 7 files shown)


packages/react/src/useAction.ts
```diff
@@ -79,6 +79,37 @@ export const useAction = <
     transformedResult,
     useCallback(
       async (variables, context) => {
+        if (action.hasAmbiguousIdentifier) {
```
jasong689, line 82: This is the same logic that we generate in the API client

packages/react/src/useAction.ts
```diff
@@ -79,9 +79,43 @@ export const useAction = <
...
+            newVariables = variables;
+          } else {
+            newVariables = {
+              [action.modelApiIdentifier]: {},
+            } as Exclude<F["variablesType"], null | undefined>;
+            for (const [key, value] of Object.entries(variables)) {
+              if (action.paramOnlyVariables?.includes(key)) {
+                newVariables[key] = value;
+              } else {
+                if (key === "id") {
```
jasong689, line 104: `id` is kind of special in that if it's an update effect then we have to include it as a separate variable

airhorns (approved):
Code LGTM, can you update the readme and docstrings of the hooks to make sure the docs reflect the new reality too?

packages/react/src/useAction.ts
```diff
@@ -79,9 +77,43 @@ export const useAction = <
...
+          if (
+            action.modelApiIdentifier in variables &&
+            typeof variables[action.modelApiIdentifier] === "object" &&
+            variables[action.modelApiIdentifier] !== null
+          ) {
+            newVariables = variables;
+          } else {
+            newVariables = {
+              [action.modelApiIdentifier]: {},
+            } as Exclude<F["variablesType"], null | undefined>;
```
thegedge, line 97: This has to be casted? :(

Is it TypeScript not figuring it out, or do we have something typed incorrectly?

packages/react/src/useAction.ts
thegedge: 👍 

Should we be checking if it's an update? Or can we look at `action.variables` to see if it should be nested under the model API identifier or not? I'm just wondering if we could future-proof against creates where the end user can supply an id, but it won't be "lifted" to a variable like we do for updates.

thegedge (approved):
🚀

packages/react/src/useAction.ts
jasong689: I get `Type '{ [x: string]: {}; }' is not assignable to type 'Exclude<F["variablesType"], null | undefined>'.`

packages/react/src/useAction.ts
jasong689: Changed this up to check `action.variables`

airhorns (approved):
This works for me!

PR #203, opened 2023-06-02
Update the react provider to accept a full instance of an api client instead of just the urql client

airhorns:
This changes the signature of the React `<Provider/>` component in a backwards compatible way to accept the `api` client instance instead of an urql client instance. We're likely to need access to more than just the urql client in the react hooks over time. We already have a nasty hack to get back from the urql client to the actual GadgetConnection object in order to power `useFetch`, and I expect more of this kind of thing to happen in the future. Let's update the API from being

```typescript
<Provider value={api.connection.currentClient} />
```

to be

```typescript
<Provider api={api} />
```

This matches the Shopify provider component, which takes an `api` prop as well.

We ensure this change is backwards compatible by still supporting the `value` prop, and still expecting it to be passed an urql client. Added a few tests to this effect as well.

Gadget side PR here: https://github.com/gadget-inc/gadget/pull/6923

### DIFF
--- a/packages/react/src/GadgetProvider.tsx
+++ b/packages/react/src/GadgetProvider.tsx
@@ -1,22 +1,98 @@
-import type { GadgetConnection } from "@gadgetinc/api-client-core";
-import { $gadgetConnection } from "@gadgetinc/api-client-core";
+import type { AnyClient, GadgetConnection } from "@gadgetinc/api-client-core";
+import { $gadgetConnection, isGadgetClient } from "@gadgetinc/api-client-core";
+import type { ReactNode } from "react";
 import React from "react";
-import type { Client } from "urql";
-import { Provider } from "urql";
+import type { Client as UrqlClient } from "urql";
+import { Provider as UrqlProvider } from "urql";
 
-export const GadgetContext = React.createContext<Client | undefined>(undefined);
-export const GadgetProvider: React.FC<React.PropsWithChildren<{ value: Client }>> = ({ children, value }) => {
+/**
+ * React context that stores the current urql client
+ *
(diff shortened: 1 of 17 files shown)


packages/react/src/GadgetProvider.tsx
```diff
@@ -1,22 +1,98 @@
...
+  value: UrqlClient;
+  children: ReactNode;
+}
+
+/**
+ * Provider wrapper component that passes an api client instance to the other hooks.
+ *
+ * This component is __required__ as a wrapper around any React components using the other hooks in this library, like `useFindMany` or `useAction`.
+ *
+ * Pass an instance of your app's api client to the `value` prop.
```
jasong689, line 52:
```suggestion
 * Pass an instance of your app's api client to the `api` prop.
```

packages/react/src/GadgetProvider.tsx
```diff
@@ -1,22 +1,98 @@
...
+ * React context that stores the current urql client
+ *
+ * urql doesn't have its own useClient hook, so we store it on our own context to get at the client object later
+ **/
+export const GadgetUrqlClientContext = React.createContext<UrqlClient | undefined>(undefined);
+
+/**
+ * React context that stores an instance of the JS Client for an app (AKA the `api` object)
+ */
+export const GadgetClientContext = React.createContext<AnyClient | undefined>(undefined);
```
jasong689, line 18: Doesn't seem like this context is consumed anywhere? This is just to match the above urql context I'm assuming.

jasong689 (approved):

jcao49 (approved):
🌟 LGTM!

PR #208, opened 2023-06-07
Switch to using pnpm

airhorns:
We're gonna use pnpm gadget side -- let's use it here too!


package.json
```diff
@@ -12,16 +12,16 @@
...
   ],
   "scripts": {
-    "lint": "yarn run lint:prettier && yarn run lint:eslint",
-    "lint:prettier": "node --max-old-space-size=4096 node_modules/.bin/prettier --check \"(packages|scripts)/**/*.{js,ts,tsx}\"",
-    "lint:eslint": "node --max-old-space-size=4096 node_modules/.bin/eslint --quiet --ext ts,tsx packages scripts",
-    "lint:fix": "node --max-old-space-size=4096 node_modules/.bin/prettier --write --check \"(packages|scripts)/**/*.{js,ts,tsx}\" && eslint --ext ts,tsx --fix packages scripts",
-    "typecheck": "yarn workspaces run tsc --noEmit",
-    "build": "yarn workspaces run build",
+    "lint": "pnpm lint:prettier && pnpm lint:eslint",
+    "lint:prettier": "NODE_OPTIONS=\"--max-old-space-size=4096\" prettier --check \"(packages|scripts)/**/*.{js,ts,tsx}\"",
```
airhorns, line 16: @thegedge I figured out a way to not have to pull out the right file from the node package binary by just using the `NODE_OPTIONS` env var to set the options, instead of running the node binary explicitly. Vibe?

thegedge: I do! I thought I tried that and it complained about not being able to put `--max-old-space-size` in NODE_OPTIONS. I'll try it again though.

packages/react/package.json
```diff
@@ -24,15 +24,19 @@
     "urql": "^3.0.3"
   },
   "devDependencies": {
+    "@gadgetinc/api-client-core": "workspace:*",
```
thegedge, line 27: Curious about this one vs the one in dependencies.

thegedge (approved):
🙌 

LGTM, but not every test passed for me locally. I tried it with yarn and the same issue though, so it's not a pnpm-specific issue we have to deal with in this PR 🚀 (but probably worth fixing)

PR #213, opened 2023-06-22
Add the select option to the internal model manager

airhorns:
We recently added a feature to the internal API for only selecting some fields of the returned records. This gives improved performance as the default can exclude computed fields which are often expensive and unused. This feature already exists in the GraphQL API, but the JS client didn't know about it. This teaches it!

I also took the opportunity to type the options and improve the JSDoc comments for the
InternalModelManager class so folks know what the heck this thing does.

This PR Gadget side actually tests this end to end: https://github.com/gadget-inc/gadget/pull/7140

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -43,23 +43,108 @@ const internalHydrationPlan = (modelApiIdentifer: string) => `
 export const internalFindOneQuery = (apiIdentifier: string) => {
   const capitalizedApiIdentifier = capitalizeIdentifier(apiIdentifier);
   return `
-    query InternalFind${capitalizedApiIdentifier}($id: GadgetID!) {
+    query InternalFind${capitalizedApiIdentifier}($id: GadgetID!, $select: [String!]) {
       ${internalHydrationPlan(apiIdentifier)}
       internal {
-        ${apiIdentifier}(id: $id)
+        ${apiIdentifier}(id: $id, select: $select)
       }
     }
     `;
 };
 
-export const internalFindFirstQuery = (apiIdentifier: string, options?: RecordData) => {
-  const capitalizedApiIdentifier = capitalizeIdentifier(apiIdentifier);
+/**
+ * A list of fields to select from the internal API
+ * Matches the format of the Public API `select` option, but only allows going one level deep -- no relationships can be selected using the internal API.
+ *
+ * Supports passing a list of strings as a shorthand.
+ *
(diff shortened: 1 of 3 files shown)


airhorns:
ping @jcao49 @angelini

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -368,14 +554,38 @@ export class InternalModelManager {
...
     });
   }
 
+  /** @private */
   async transaction<T>(callback: (transaction: GadgetTransaction) => Promise<T>): Promise<T> {
     return await this.connection.transaction(callback);
   }
 }
+
+function formatInternalSelectVariable(select: InternalFieldSelection | undefined): undefined | string[] {
```
angelini, line 581: If we pass `select: { id: false }` that's the equivalent of passing `[]`.

I'm not sure why we have support for `Record<string, boolean>` instead of just a string array. I know it's to match our existing API but what does this object syntax get you that the array doesn't?

angelini (approved):
Looks good, just a couple API questions

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -43,23 +43,108 @@ const internalHydrationPlan = (modelApiIdentifer: string) => `
...
+  /**
+   * How to sort the returned records
+   * Matches the format and behavior of the Public API `sort` option
+   *
+   * @example
+   * {
+   *   sort: { publishedAt: "Descending" }
+   * }
+   **/
+  sort?: Record<string, "Ascending" | "Descending"> | Record<string, "Ascending" | "Descending">[];
```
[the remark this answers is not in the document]
airhorns: In your example, it'd be a,b because since ES6 key iteration order on objects is actually stable and defined, its in the order you see them in the code. This is subtle though, and the GraphQL API actually supports both with a weird coercion thing, so the type supports both here too.

packages/api-client-core/src/InternalModelManager.ts
airhorns: Good question! It gets consistency with all the other api functions in the JS client. The Public API supports the nested selection style, and I think that'll be what acquaints most folks with the idea of JS selections at all. I think its kind of regrettable that the internal api works different at a graphql level, but since just about everyone is actually using the JS client and doesn't really care what it does under the hood, I think we should make it quack as similarly as possible!

airhorns:
Gonna merge this so we can get it out with the next release!

PR #215, opened 2023-06-24
Suspense support for React hooks

airhorns:
React's new Suspense support is the perfect thing for components declaring that they need something to happen before they can fully render. With all the upcoming auth work that we are doing, I think we're going to have lots of situations on the frontend where we want to know current login state, but where we don't have it yet because its a GraphQL call away.

We could solve this by wrapping the whole app in a thing that doesn't render until the session has been fetched. Thats what the `<ClerkProvider/>` does. I think we could do the same thing in our provider, but that'd require us assuming that every provider wants a session, which I don't think is a great assumption.

Instead, I'd rather have the components that need to know login state block, but the ones that don't ... not! We can do that with suspense. If we add a `<RequireLogin/>` component, it can access session state with a suspense, and then when it arrives, actually decide what to do. It would use the normal, community-agreed-upon mechanism for showing intermediate state spinners or what have you as well. I think our app template should have a base-level suspense that shows a big spinner, but that's nice cause users can customize it or add suspense catchers farther down the tree.

So, this adds suspense support to our React hooks! You can see it as an alternative to the `fetching` boolean. Instead of managing the fetching state, you punt it to your parent. See the tests for invocation examples.

One annoyance with the way urql's suspense support works is that you have to turn suspense support on at the whole client level, and then individual hooks can choose to enable it or not. By default in urql, when you turn on suspense support in the client, all hooks start using it, and I don't think we want that, we just want folks to be able to use it if they want to. So, there's a bit of args munging to have suspense on at the client level so urql will do it at all, and then we opt-in on a per-hook basis.

### DIFF
--- a/packages/blog-example/src/App.tsx
+++ b/packages/blog-example/src/App.tsx
@@ -1,5 +1,5 @@
 import { useFetch, useFindMany } from "@gadgetinc/react";
-import { useEffect, useState } from "react";
+import { Suspense, useEffect, useState } from "react";
 import { api } from "./api";
 import "./styles/App.css";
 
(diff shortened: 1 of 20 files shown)


airhorns:
ping @kristianpd  @jasong689

packages/react/src/utils.ts
```diff
@@ -182,3 +240,38 @@ export class ErrorWrapper extends Error {
...
+  requestPolicy?: RequestPolicy;
+  suspense?: boolean;
+}
+
+/** Generate `urql` query argument object, for `useQuery` hook */
+export const useMemoizedQueryArgs = <Plan extends QueryPlan, Options extends QueryOptions>(plan: Plan, options?: Options): UseQueryArgs => {
+  // use a memo as urql rerenders on context identity changes
+  const context = useMemo(() => {
+    return {
+      suspense: !!options?.suspense,
```
jasong689, line 261: i'm guessing this keeps suspense off by default for all hooks

packages/blog-example/src/App.tsx
```diff
@@ -42,12 +42,49 @@ function ExampleFindMany() {
...
+  );
+}
+
 function App() {
   return (
     <div className="App">
       <h1>Vite + Gadget</h1>
-      <ExampleFetch />
-      <ExampleFindMany />
+      {/* <ExampleFetch />
```
jasong689, line 85: Were these commented out for a reason?

jasong689 (approved):

packages/blog-example/src/App.tsx
airhorns: Whoops nope thanks

packages/react/src/utils.ts
airhorns: Correct!

kristianpd (approved):

PR #216, opened 2023-06-26
Add the graphql operation to the client URL

airhorns:
We have this Gadget side, I think it'd be really helpful API client side to see in the network tab which operation is doing what:

CleanShot 2023-06-25 at 21 32 47@2x

### DIFF
--- a/packages/api-client-core/src/urlParamExchange.ts
+++ b/packages/api-client-core/src/urlParamExchange.ts
@@ -1,0 +1,34 @@
+import type { Exchange } from "@urql/core";
+import type { DocumentNode, OperationDefinitionNode } from "graphql";
+import { onPush, pipe } from "wonka";
+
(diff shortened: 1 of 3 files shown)


packages/api-client-core/src/urlParamExchange.ts
```diff
@@ -0,0 +1,34 @@
...
+    const firstSelection = lastDefinition.selectionSet.selections.find((node) => node.kind == "Field") as any;
+    return firstSelection.name.value;
+  }
+};
+
+export const urlParamExchange: Exchange = ({ forward }) => {
+  return (ops$) =>
+    pipe(
+      ops$,
+      onPush((op) => {
```
jasong689, line 22: was wondering what the difference between `tap` and `onPush` was and turns out they're the same.

jasong689 (approved):

PR #217, opened 2023-06-26
Add a useApi hook for accessing the current api client without having a reference to it

airhorns:
We need this for the auth helpers that ideally don't need to be passed an instance of the api every time, like:

```typescript
export const useSession = () => {
  const api = useApi();
  const [{ data, fetching, error }] = useGet(api.currentSession, {
    select: {
      id: true,
      state: true,
      createdAt: true,
      userId: true
    }
  });
  return data;
}
```

With this hook, we can export that kind of framework code, and rely on the user `<Provider/>`-ing us the api client we need


jasong689 (approved):
❤️

PR #220, opened 2023-06-30
Auth hooks and components

kristianpd:
This PR adds a few new basic hooks and react components for working with Gadget's upcoming auth system 

- hooks: `useSession`, `useUser`, `useIsSignedIn` 
- components: `<SignedIn>`, `<SignedOut>`, `<SignedInOrRedirect>` conditionally render react `children` or not depending on the user's session state. `SignedInOrRedirect` will also `window.location.assign` redirect the browser if the user is not signed in. They will be sent to the `Provider`'s configured `signInPath`. 

## PR Checklist


packages/react/spec/auth/useUser.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  test("it returns null when the user is logged out", async () => {
+    const { result, rerender } = renderHook(() => useUser(), { wrapper: TestWrapper(superAuthApi) });
+
+    expectMockSignedOutUser();
+
+    rerender();
+    expect(result.current).toBe(null);
+  });
```
airhorns, line 26: Can you test what happens when the call to the backend errors?

packages/react/spec/components/SignedInOrRedirect.spec.tsx
```diff
@@ -0,0 +1,59 @@
+import "@testing-library/jest-dom";
+import { render } from "@testing-library/react";
+import React from "react";
+import { superAuthApi } from "../../spec/apis";
+import { TestWrapper } from "../../spec/testWrapper";
+import { expectMockSignedInUser, expectMockSignedOutUser } from "../../spec/utils";
+import { SignedInOrRedirect } from "../../src/components/SignedInOrRedirect";
+
+describe("SignedInOrRedirectOrRedirect", () => {
```
airhorns, line 9:
```suggestion
describe("SignedInOrRedirect", () => {
```

packages/react/spec/components/SignedInOrRedirect.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { location } = window;
+  const mockAssign = jest.fn();
+
+  beforeAll(() => {
+    // @ts-expect-error mock
+    delete window.location;
+    // @ts-expect-error mock
+    window.location = { assign: mockAssign };
```
airhorns, line 17: `jest.spyOn(window.location, "assign")` no worky? I don't super care just curious

packages/react/spec/components/SignedInOrRedirect.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  afterAll(() => {
+    window.location = location;
+  });
+
+  test("redirects when signed out", () => {
+    const component = (
+      <h1>
+        <SignedInOrRedirect>Hello, Jane!</SignedInOrRedirect>
```
airhorns, line 31: I think we're going to need to make the way this thing redirects pluggable, so that when we have our own router, it can push the route client side, or if someone wants to use this with their own router in a next.js project, they can. Thoughts on adding a new provider prop now?

packages/react/src/auth/useSession.ts
```diff
@@ -0,0 +1,37 @@
...
+import { useGet } from "../../src/useGet";
+
+export interface GadgetSession {
+  id: string;
+  userId: string | null;
+  user: GadgetUser | null;
+  [key: string]: any;
+}
+
+export interface GadgetUser {
```
airhorns, line 11: Dang, I didn't realize this before, but putting the hooks in this package means they can't reference the types of the actual upstream Gadget models, so this type is not super useful. I think that if the hook below was actually passed an api client where the type of that client could be inferred, it could compute the return type properly, but it doesn't get the api client's type because it comes from `useApi`. Dang.

packages/react/src/auth/useIsSignedIn.ts
```diff
@@ -0,0 +1,6 @@
+import { useUser } from "./useUser";
+
+export const useIsSignedIn = () => {
```
airhorns, line 3: Can you add JSDocs for each of these hooks? Doesn't have to be huge

packages/react/src/GadgetProvider.tsx
```diff
@@ -48,15 +58,20 @@ export interface DeprecatedProviderProps {
...
  * </Provider>
  *
  * @example the Provider accepts the deprecated form of passing an urql client object right in -- this is deprecated and will be removed in a future version. Instead, just pass the whole api instance.
- * <Provider value={api.connection.currentClient}>
+ * <Provider api={api.connection.currentClient}>
+ *   <MyApp />
+ * </Provider>
+ *
+ * @example the Provider accepts option sign in and sign out paths.
+ * <Provider api={api.connection.currentClient} signInPath="/auth/signin" signOutPath="/auth/signout">
```
airhorns, line 74: Do you think we're gonna add more paths over time, such that we should make this API `paths={{signIn: ..., signOut: ...}}`?

packages/react/yarn.lock
```diff
@@ -0,0 +1,1073 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
airhorns, line 1: I think this is an accidental commit 🔥

airhorns (commented):
Can you add these to the readme for the react package?

packages/react/spec/components/SignedInOrRedirect.spec.tsx
kristianpd: I encountered this problem: https://remarkablemark.org/blog/2018/11/17/mock-window-location/#spy-on-method

This is the workaround for jsdom 14+

packages/react/src/auth/useSession.ts
kristianpd: I can avoid exporting `GadgetUser`, but the implementation currently has dependencies on the field `userId`. We could look at moving it into core if it lets us build better types, but do you think this should ship as is?

packages/react/src/GadgetProvider.tsx
kristianpd: I guess at that point it's borderline a `configuration` block. To keep it most flexible long term we could consider making the `auth` config it's own property: 

```tsx
<Provider api={api} auth={{ signInPath, signOutPath}} />
```

packages/react/src/auth/useSession.ts
```diff
@@ -0,0 +1,37 @@
+import { useApi } from "../../src/GadgetProvider";
+import { useGet } from "../../src/useGet";
+
+export interface GadgetSession {
```
jasong689, line 4: I had to type this in `@gadgetinc/auth` too. Maybe this needs to be moved to `api-client-core`

packages/react/src/auth/utils.ts
```diff
@@ -0,0 +1,9 @@
+import type { GadgetSession } from "./useSession";
+
+export const isSessionSignedOut = (session: GadgetSession) => {
+  return session?.userId == undefined;
```
jasong689, line 4: Could `userId` be any other falsy values? I'm thinking maybe it's better to be `!!session?.userId`

packages/react/src/utils.ts
```diff
@@ -266,7 +267,7 @@ export const useMemoizedQueryArgs = <Plan extends QueryPlan, Options extends Que
   return {
     query: plan.query,
     variables: plan.variables,
-    ...options,
+    ...omit(options, ["context", "suspense"]),
```
kristianpd, line 270: just being a little more explicit about what goes where

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.12.7",
+  "version": "0.13.0",
```
airhorns, line 3: Confusingly, npm will treat this as effectively a new major version, which means `npm update @gadgetinc/react` won't ever move the version for existing apps. Since this isn't a breaking change, can we make it a patch version instead? I don't feel super strongly. This is also why that `versions` test is failing

kristianpd: There is a breaking change to `GadgetClientContext` where before it simply returned the `api` client, and now it is called `GadgetConfigurationContext` and returns `{ api, auth }`

I think that warrants an major/minor version bump?

airhorns: I'm cool with that, but I think that context is best considered private to the implementation such that the provider should really be the thing that accepts all input you know

airhorns (approved):

packages/react/src/components/auth/SignedInOrRedirect.tsx
```diff
@@ -0,0 +1,30 @@
+import type { ReactNode } from "react";
+import React, { useContext, useEffect, useState } from "react";
+import { GadgetConfigurationContext } from "../../../src/GadgetProvider";
+import { useSession } from "../../../src/auth/useSession";
+import { isSessionSignedIn } from "../../../src/auth/utils";
+
+/**
```
jasong689, line 7: I think this is extra

jasong689 (approved):
🎩 on the auth template branch 👍

PR #223, opened 2023-07-06
useSession on apps without a User model

kristianpd:
Updates #220 to allow apps to use the `useSession` hook without the presence of a `User` model

## PR Checklist


airhorns (approved):

packages/react/spec/auth/useAuth.spec.ts
```diff
@@ -49,4 +49,14 @@ describe("useAuth", () => {
...
   });
+
+  test("it throws when the api client does not have a User model", async () => {
+    expect(() => {
+      const { rerender } = renderHook(() => useAuth(), { wrapper: TestWrapper(noUserModelApi) });
+
+      mockNetworkError();
+
+      rerender();
+    }).toThrowErrorMatchingInlineSnapshot(`"[Network] Network error"`);
```
jasong689, line 60: Why does this test assert a network error instead of a missing user model error?

PR #224, opened 2023-07-11
add a table of contents to the react package readme file

rdraward:
The README for `@gadgetinc/react` has become pretty big. A table of contents will be useful for finding the docs I want to look at

## PR Checklist

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -27,6 +27,39 @@
 This library wraps [urql](https://github.com/FormidableLabs/urql) to communicate with a Gadget-generated API, so it benefits from all of
 the same [features](https://github.com/FormidableLabs/urql/blob/main/README.md#-features) as `urql` as well!
 
+### Table of Contents
+
+ - [Installation](#installation)
+ - [Example usage](#example-usage)
+ - [Request caching](#request-caching)
+ - [API Documentation](#fastify-v1x-and-v2x)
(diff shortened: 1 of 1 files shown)


airhorns:
Is this materially better than the autogenerated table of contents right next to it on the right? I kinda feel like this just duplicates that

rdraward:
> Is this materially better than the autogenerated table of contents right next to it on the right? I kinda feel like this just duplicates that

I only get the auto-gen table of contents when I'm actually looking at the README file. Do you see something different @airhorns? I agree, I'd would rather not duplicate but unless the README is at the actual repo root level I don't seem to get a working toc

https://github.com/gadget-inc/js-clients/assets/20960033/d6697f7b-64a9-4990-a81f-004d41b96a0c

airhorns:
Oh dang sorry I was not paying close attention -- this is in the js-clients repo! I think this TOC totally makes sense over here, but, what do you think about deduplicating this content and adding a link to docs.gadget.dev from within the readme so we don't need to maintain both?

rdraward:
I hadn't seen the docs in our new ref section! Totally agree, a link is a much better way to reduce the maintenance burden. We don't need this info in two places

rdraward:
> @rdraward want to ship a link?

Sorry, I'll close this PR! We can totally replace the existing content with a link to our Ref docs

PR #225, opened 2023-07-14
Verify Session has associated User for auth components

jcao49:
This PR updates the logic to check for user on the session that is signed in/out. Handles a possible scenario in which a user record is deleted without having signed out, and so the session record still exists as `signed-in` with a user id that doesn't actually point anywhere anymore. Our current auth `SignedIn/Out/orRedirect` components would then misbehave and users would get caught in a weird state.

[GGT-4440](https://linear.app/gadget-dev/issue/GGT-4440/signedin-component-renders-children-when-a-session-has-no-user)

Current behaviour 
- with old template: https://hello-test--development.gadget.app/ 
- with new/upcoming template: https://new-auth-template--development.gadget.app/

1. Log in to app
2. Delete user record
3. Refresh app page - weird state, you can go back to `/` / and says you're signed out but you can go back to `/signed-in` and it seems you're signed in (though no data is rendered)

With updates to the auth components (stuck them in App.jsx): https://auth-react-components-test--development.gadget.app/ 
1. Log in to app
2. Delete user record
3. Refresh page- should not give you access to `/signed-in` and redirect you to the `/` with an option to sign back in

## PR Checklist

### DIFF
--- a/packages/react/spec/utils.ts
+++ b/packages/react/spec/utils.ts
@@ -35,6 +35,21 @@ export const expectMockSignedOutUser = () => {
   });
 };
 
+export const expectMockDeletedUser = () => {
+  expect(mockUrqlClient.executeQuery).toBeCalledTimes(1);
+  mockUrqlClient.executeQuery.pushResponse("currentSession", {
+    data: {
+      currentSession: {
+        id: "123",
+        userId: 1,
+        user: null,
+      },
+    },
+    stale: false,
+    hasNext: false,
+  });
+};
+
 export const mockInternalServerError = () => {
   expect(mockUrqlClient.executeQuery).toBeCalledTimes(1);
   mockUrqlClient.executeQuery.pushResponse("currentSession", {
--- a/packages/react/src/GadgetProvider.tsx
+++ b/packages/react/src/GadgetProvider.tsx
@@ -59,7 +59,7 @@ export interface DeprecatedProviderProps {
   children: ReactNode;
 }
 
-// default Gadgte auth signIn and signOut paths
+// default Gadget auth signIn and signOut paths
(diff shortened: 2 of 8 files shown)


packages/react/spec/utils.ts
```diff
@@ -35,6 +35,21 @@ export const expectMockSignedOutUser = () => {
...
+        id: "123",
+        userId: 1,
+        user: null,
+      },
+    },
+    stale: false,
+    hasNext: false,
+  });
+};
+
```
kristianpd, line 52: :+1:

kristianpd (approved):

packages/react/src/GadgetProvider.tsx
```diff
@@ -60,8 +60,8 @@ export interface DeprecatedProviderProps {
 }
 
 // default Gadgte auth signIn and signOut paths
```
jasong689, line 62: nit: What is Gadgte?

packages/react/src/components/auth/SignedInOrRedirect.tsx
```diff
@@ -21,7 +20,7 @@ export const SignedInOrRedirect = (props: { children: ReactNode }) => {
     }
```
jasong689, line 20: Should the above also check for `user`? 
`if (auth && !redirected && (!isSignedIn || !user))`

jasong689 (commented):
Looks good except for the one comment about `SignedInOrRedirect`

packages/react/src/GadgetProvider.tsx
jcao49: 😆 yeah @kristianpd what is it?

PR #227, opened 2023-07-20
Add some debug logging to the react-shopify-app-bridge for help debugging auth issues

airhorns:
I want to be able to understand better why the shopify app bridge is doing what it's doing. If we use the `console.debug` log level on the client, the logs won't show by default in developers consoles, until they adjust the logging filter! I think this'd be useful for us and for us collecting log samples from customers when things are broken.


airhorns:
ping @kristianpd

PR #230, opened 2023-07-21
Don't start new, single operation transactions in the internal model manager

airhorns:
Not exactly sure why, but before this change, the `InternalModelManager` class wrapped each of the write operations it could do in a transaction block. I don't think this really has an effect other than making them slower.
 - when already inside a transaction, the `this.connection.transaction` block doesn't do anything special, it just carries on the existing outer transaction
 - when not already inside a transaction, the `this.connection.transaction` block starts a graphql-ws client and sets it up and returns it. The InternalModelManager then takes it, does the one write with it, and returns, which means that user can't have ever added work into that transaction. The transaction could *only* have done the one thing inside it.

I think this is legacy thing from the very early days of this class where for some reason it *needed* every operation to be wrapped in a transaction. I dug up the place where this class started: https://github.com/gadget-inc/gadget/blame/7121c5bcb49332c5ad172e9aa485db837aab37b4/packages/api-client/src/InternalModelManager.ts and it looks like it really just started with this behaviour. I think there might have been a notion that writes always had to go through a transaction at the time, but that's definitely not the case anymore.

`this.connection.currentClient` will be the open transaction's client if there is one, or just the plain old, performant HTTP client if not. That HTTP client is gonna be a lot faster than roundtripping to open a websocket, wait for it to initialize, submit the write, then get the response, so I think this will be a nice win for simplicity and performance for anyone using the internal api in a script.

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -2,7 +2,6 @@ import { query } from "gql-query-builder";
 import type { GadgetConnection } from "./GadgetConnection";
 import type { GadgetRecord, RecordShape } from "./GadgetRecord";
 import { GadgetRecordList } from "./GadgetRecordList";
-import type { GadgetTransaction } from "./GadgetTransaction";
 import {
   GadgetClientError,
   GadgetOperationError,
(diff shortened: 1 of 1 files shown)


thegedge (approved):
👍 I definitely remember chatting about this super long ago, but can't remember specifics. Game though!

jasong689 (approved):
Makes sense to me!

PR #231, opened 2023-07-24
tiny operation compiler

airhorns:
I needed to add a feature to the bit that compiles GraphQL query strings from our `select:` option and dove into the codebase for `gql-query-builder` once more. I didn't like it in there -- there's a bunch of adapter layers, awkward input syntax, and bundle size bloat introduced by it, and it genuinely seemed like more work to add the feature there than to just rewrite. I tried it, and did it in a few hours, and the new thing is smaller and adds the new feature I need (directive support) right out of the box! Introducing `tiny-graphql-operation-compiler`. 

Build in gadget using it: [@gadgetinc/react-v0.13.1-gitpkg-da07cee](https://github.com/gadget-inc/gadget/pull/7653)

The bundle size is reduced a little as well:

On `main`:

```
> test-bundles@0.1.0 build /Users/airhorns/Code/js-clients/packages/test-bundles
> ts-node --transpileOnly --esm build.ts

vite v4.4.7 building for production...
✓ 249 modules transformed.
dist/api-read-stats.html      236.91 kB │ gzip: 49.90 kB
dist/test-bundle-api-read.js  267.31 kB │ gzip: 48.98 kB
dist/api-read-stats.html       237.47 kB │ gzip: 49.97 kB
dist/test-bundle-api-read.cjs  179.09 kB │ gzip: 40.81 kB
✓ built in 2.48s
vite v4.4.7 building for production...
✓ 523 modules transformed.
dist/react-read-stats.html      330.62 kB │ gzip: 60.12 kB
dist/test-bundle-react-read.js  372.50 kB │ gzip: 78.15 kB
dist/react-read-stats.html       332.36 kB │ gzip: 60.17 kB
dist/test-bundle-react-read.cjs  253.87 kB │ gzip: 66.02 kB
✓ built in 2.72s
```

On this branch:

```
> test-bundles@0.1.0 build /Users/airhorns/Code/js-clients/packages/test-bundles
> ts-node --transpileOnly --esm build.ts

vite v4.4.7 building for production...
✓ 221 modules transformed.
dist/api-read-stats.html      226.46 kB │ gzip: 48.78 kB
dist/test-bundle-api-read.js  251.58 kB │ gzip: 46.78 kB
dist/api-read-stats.html       226.99 kB │ gzip: 48.86 kB
dist/test-bundle-api-read.cjs  166.72 kB │ gzip: 38.91 kB
✓ built in 2.51s
vite v4.4.7 building for production...
✓ 495 modules transformed.
dist/react-read-stats.html      318.94 kB │ gzip: 58.88 kB
dist/test-bundle-react-read.js  356.77 kB │ gzip: 75.93 kB
dist/react-read-stats.html       321.61 kB │ gzip: 58.89 kB
dist/test-bundle-react-read.cjs  241.50 kB │ gzip: 64.12 kB
✓ built in 2.79s
```


packages/api-client-core/spec/InternalModelManager.spec.ts
```diff
@@ -134,13 +134,23 @@ describe("InternalModelManager", () => {
       const plan = internalFindManyQuery("widget", undefined);
       expect(plan).toMatchInlineSnapshot(`
         {
-          "query": "query InternalFindManyWidget ($after: String, $before: String, $first: Int, $last: Int) { internal  { listWidget (after: $after, before: $before, first: $first, last: $last) { pageInfo { hasNextPage, hasPreviousPage, startCursor, endCursor }, edges { cursor, node } } } }",
```
garymardell, line 137: Lots of speculation on this one... but I'm not sure whether you are doing query parsing caching at all but previously it would always generate the same query (with unused variables) and therefore (if cached) would hit the same cache key however now there are more distinct queries being generated depending on variables provided.

garymardell (approved):
Just one question but otherwise looks good to me ❤️ for adding directive support to the builder

packages/tiny-graphql-query-compiler/Readme.md
```diff
@@ -0,0 +1,127 @@
...
+To pass arguments using GraphQL variables, you can use the `Call` helper with the `Var` helper:
+
+```javascript
+import { compile, Call, Var } from "tiny-graphql-query-compiler";
+
+const query = compile({
+  type: "query",
+  fields: {
+    user: Call(
+      { id: Var({ type: "ID!" }) },
```
garymardell, line 69: I'm wondering whether the `required` option should be removed from `Var`. As far as I can tell it only serves to add the `!` onto the type name but as in the example above can just be done in the type name itself. 

Another example is the filter sort options where it is a list of required items, so we are already having to write `!` into the type name already.

packages/tiny-graphql-query-compiler/spec/query.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+        id: true,
+        name: true,
+        age: true,
+      },
+      directives: ["@live"],
+    });
+
+    expectValidGraphQLQuery(result);
+    expect(result).toMatchInlineSnapshot(`
+      "query GetUsers @live {
```
thegedge, line 174: 🙌

packages/tiny-graphql-query-compiler/spec/calls.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+        users(first: 10) { 
+          edges {
+            node {
+              id
+            }
+          } 
+        }
+      }"
+    `);
+  });
```
thegedge, line 237: A test I didn't see but I think would be good to have to clarify:

```js
compile({
  type: "query",
  fields: {
    something: Call({ length: Var({ type: "Int" }) }),
    another Call({ length: Var({ type: "Int!" }) }),
  },
});
```

So those are two different variables, but the argument name is the same. Do we throw? Does one clobber the other? Worth a test, eh?

packages/tiny-graphql-query-compiler/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+      } else {
+        return `${field} {
+${compileFieldSelection(value as FieldSelection, onlyPresentVariableValues)}
+}`;
+      }
+    })
+    .filter((value) => !!value)
+    .join("\n");
+
+  return indent(result, 2);
```
thegedge, line 55: nit: what do you think about avoiding the join + split + map + join by writing the above like:

```js
.filter((value) => !!value)
.map((value) => " ".repeat(2) + value)
.join("\n");
```

packages/tiny-graphql-query-compiler/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+  const result = Object.entries(fields)
+    .map(([field, value]) => {
+      if (typeof value === "boolean") {
+        return value ? field : false;
+      } else if (value instanceof FieldCall) {
+        let args = "";
+        let entries = Object.entries(value.args).filter(([_, value]) => value !== null && value !== undefined);
+
+        if (onlyPresentVariableValues) {
+          entries = entries.filter(([_, value]) => (value instanceof Variable ? value.present() : true));
```
thegedge, line 29: naming nit: I was actually reading `value.present()` as "present this variable" vs "is present" (which I figured out after checking out the class). A little room for confusion, so up to you if you'd like to maybe clarify that it's the latter!

thegedge (approved):
It's super simple, so I'm game for owning this. Nice work! Gives us a ton of room to play around with directives and argument calls in a syntax that works for us :)

PR #234, opened 2023-07-24
Convert api-client-core and react packages to emit both CJS and ESM

airhorns:
This adds build config for `@gadgetinc/api-client-core` and `@gadgetinc/react` to emit both CJS and ESM output into the same package. This is a bit tricky to configure but is totally possible! Why:
 - ESM is the future, gotta do this eventually
 - ESM treeshakes a lot better when it can be imported by the end-bundler
 - ESM compresses better when bundled altogether with other ESM
 - we need to build some experience with ESM
 - vite only supports bundling ESM in this monorepo style as best I can tell and i want to use it for assessing bundle size

Here's a build testing these out (prerelease) Gadget side: https://github.com/gadget-inc/gadget/pull/7665
and here's a gadget app with these running via yarn resolutions: https://auth-demo-whatever.gadget.app/edit/files/package.json


packages/react/spec/imports/named-esm-import.mjs
```diff
@@ -0,0 +1 @@
+import { Provider } from "../../dist/esm/index.js";
```
thegedge, line 1: Should we also do a `throw` if `Provider` is undefined, or does ESM give us some guarantees around that?

thegedge (commented):
Would you say the three key changes were:

1. including both `exports` and `main` in `package.json`,
2. using the `.js` suffix on all imports, and
3. setting up two `tsconfig` files for building each.

Did I miss anything? If so, that's not too bad at all! 🙌

packages/react/spec/imports/named-esm-import.mjs
airhorns: It will do it for us, I experienced that when putting this together

thegedge: Just to double check: do you think there's a scenario where we accidentally have `export const Provider = ...something evaluating to undefined...`? If not, 🚀

thegedge (approved):
I'm all for gaining experience on the CJS/ESM front, becoming experts in the JS ecosystem. I'm also all for going the route of supporting both, versus the more aggressive "only ESM" approach some packages have taken 🚀

airhorns:
ping @kristianpd

PR #239, opened 2023-07-25
`@live` Queries

airhorns:
Using the `@live` query directive, we can ask Gadget to give us back a stream of data as it changes from the backend! We can use the graphql-ws transport we already have setup and productionized for this, and we can use urql's ability to rerender at will to trigger a nice rerender. To start, live queries are only available in the React hooks for rendering. 

Depends on https://github.com/gadget-inc/gadget/pull/8413 for the json diff patch format stuff.

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -28,6 +28,9 @@
     "prerelease": "gitpkg publish"
   },
   "dependencies": {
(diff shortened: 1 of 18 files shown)


airhorns:
ping @garymardell

garymardell (approved):

PR #240, opened 2023-07-25
Add useSignOut hook

kristianpd:
This PR introduces some small breaking changes on an API we haven't officially released yet, so I opted to keep  0.13.5 to keep it easier to ugprade on apps with `^0.13.x` 

- replace `signOutPath` in favor of `signOutActionApiIdentifier`, this is the action on the `User` model we'll call
- add new `useSignOut` hook that calls the `signOutActionApiIdentifier` action
- change `<SignedOutOrRedirect />` to take a `path` prop to indicate where to redirect to instead of pulling it from the provider context

## PR Checklist


packages/react/src/auth/useSignOut.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export const useSignOut = (opts?: { redirectOnSuccess?: boolean; redirectToPath?: string }) => {
+  const redirectOnSuccess = opts?.redirectOnSuccess ?? true;
+  const redirectToPath = opts?.redirectToPath;
+  const api = useApi();
+  const user = useUser();
+  const context = useContext(GadgetConfigurationContext);
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const { signOutActionApiIdentifier, signInPath } = context!.auth;
+  if (signOutActionApiIdentifier && (api as any).user[signOutActionApiIdentifier]) {
```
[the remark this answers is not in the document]
airhorns: I think we should do that but I think we can defer until someone asks us for that configurability! Could be months!

packages/react/spec/auth/useSignOut.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+
+      await act(async () => {
+        await mutationPromise;
+      });
+    } catch (e) {
+      caughtError = e;
+    }
+    expect(caughtError).toMatchInlineSnapshot(`[ErrorWrapper: [GraphQL] GGT_INTERNAL_ERROR: User could not be signed out.]`);
+  });
+});
```
airhorns, line 170: 👏

airhorns (approved):
LGTM pending the oen comment

packages/react/src/auth/useSignOut.ts
```diff
@@ -0,0 +1,39 @@
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const [{ fetching, error, data }, signOutAction] = useAction((api as any).user[signOutActionApiIdentifier]);
```
jasong689, line 21: will the user need to hook into when the sign out action is running? maybe we should return fetching or enable suspense here

packages/react/src/auth/useSignOut.ts
```diff
@@ -0,0 +1,39 @@
+import { useCallback, useContext, useEffect } from "react";
+import { GadgetConfigurationContext, useApi } from "../../src/GadgetProvider";
+import { useAction } from "../../src/useAction";
+import { useUser } from "./useUser";
+
+/**
+ * Returns a callback that will call the configured `signOutActionApiIdentifier` on the `User` model and optionally redirect (by default).
+ * @returns The `Promise` to call to execute the `signOut` action. Will automatically redirect to the configured `signInPath` unless the `redirect: false` option is provided.
```
jasong689, line 8: maybe a bit misleading as this describes the return value of the function that's returned

jasong689 (approved):
code looks good. just a few considerations/comments

PR #241, opened 2023-08-01
`baseRouteURL` connection option for proper absolute route building

airhorns:
Outside of Gadget's internal infrastructure, the endpoint the API client is configured with is the public HTTP URL for the app + "/api/graphql". This means we can use the endpoint to infer the public HTTP URL for the app, which we use for building a fully qualified URL when someone does `api.fetch("/some/relative/path")` or `useFetch("/some/relative/path")`.

Sadly, this breaks inside Gadget's infrastructure, where we set up api clients with a custom GraphQL endpoint that uses our internal sandbox<->api connection for performance. The endpoint inside the sandbox is `api.gadget-services.net`, which means `api.fetch("/foo"` becomes a request to `https://api.gadget-services.net/foo`, which isn't right at all. It's weird that folks would be doing `api.fetch` within the sandbox to make an API call back to the sandbox they are already in, but stranger things have happened, and I think it should still work despite being roundabout.

So, we need a way to build the correct absolute URLs for relative inputs from within the sandbox. To KISS, this just adds a new option that we can pass to the api client in the sandbox. It's backwards and forwards compatible in that old api clients wont break when passed this, and new api clients don't need it to be passed. API clients constructed outside the sandbox by anyone other than us can use the public API endpoint for inference and don't need to pass this option either.

## PR Checklist

### DIFF
--- a/packages/api-client-core/spec/GadgetConnection-suite.ts
+++ b/packages/api-client-core/spec/GadgetConnection-suite.ts
@@ -699,6 +699,21 @@ export const GadgetConnectionSharedSuite = (queryExtra = "") => {
       expect(await result.text()).toEqual("hello");
     });
 
+    test("fetch can pass relative string paths when used with a baseRouteURL", async () => {
+      const fetch = jest.fn().mockResolvedValue(new Response("hello")) as any;
+      const connection = new GadgetConnection({
+        endpoint: "https://api.internal.net/api/graphql?operation=meta",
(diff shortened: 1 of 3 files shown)


infiton (approved):

airhorns:
Here's what the gadget side will look like once this is published: https://github.com/gadget-inc/gadget/compare/internal-base-route-url?expand=1

jasong689 (approved):
👍

PR #242, opened 2023-08-02
extends useFetch to provide helpers to know when a response is in the…

infiton:
While building a chat template with `useFetch` @jasong689 ran into issues being able to know when the `data` with `streaming: string` was in the process of being updated from a stream. This PR adds to helpers for this

- In the FetchResult you can do `[{data, streaming}] = useFetch(...)` which will be true while `data` is being updated from a stream and `false` otherwise.
- You can also provide a `onStreamComplete` callback that will run when a stream successfully completes

## PR Checklist

### DIFF
--- a/packages/react/src/useFetch.ts
+++ b/packages/react/src/useFetch.ts
@@ -9,6 +9,7 @@ export interface FetchHookState<T> {
   response?: Response;
   error?: ErrorWrapper;
   fetching: boolean;
+  streaming: boolean;
   options: FetchHookOptions;
 }
 
@@ -16,14 +17,20 @@ export type FetchHookResult<T, U = T> = [FetchHookState<T>, (opts?: Partial<Fetc
 
 type FetchAction<T> =
   | { type: "fetching" }
+  | { type: "streaming" }
   | { type: "fetched"; payload: T }
+  | { type: "streamed" }
   | { type: "update"; payload: T }
   | { type: "error"; payload: ErrorWrapper };
 
 const reducer = <T>(state: FetchHookState<T>, action: FetchAction<T>): FetchHookState<T> => {
   switch (action.type) {
     case "fetching":
-      return { ...state, fetching: true, error: undefined };
+      return { ...state, fetching: true, streaming: false, error: undefined };
+    case "streaming":
+      return { ...state, streaming: true };
+    case "streamed":
+      return { ...state, streaming: false };
     case "fetched":
       return { ...state, fetching: false, data: action.payload, error: undefined };
     case "update":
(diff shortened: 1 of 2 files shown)


packages/react/src/useFetch.ts
```diff
@@ -112,7 +124,8 @@ export function useFetch(path: string, options?: FetchHookOptions): FetchHookRes
 export function useFetch<T = string>(path: string, options?: FetchHookOptions): FetchHookResult<T> {
   // Used to prevent state update if the component is unmounted
   const mounted = useRef<boolean>(true);
-  const memoizedOptions = useStructuralMemo<FetchHookOptions>(options ?? {});
+  const { onStreamComplete, ...optionsToMemoize } = options ?? {};
```
infiton, line 127: need to not pass this into useStructuralMemo, since otherwise will lead to a new value on every render. Is there a pattern to make sure this function is memoized and if not memoize it?

thegedge: So the purpose of `useStructuralMemo` is to _not_ produce a new identity if the new value is structurally the same as the previous. This should avoid re-renders.

I don't think the issue is actually `useStructuralMemo`, which I recommend you keep, but with the caller passing in a new instance of `onStreamComplete` every time. A couple of thoughts:

1. This is an issue in most libraries, where you're expected to `useCallback` when passing things into library components.
2. Find out if there's a way we can safely compare two functions (doubtful it's perfect, but we could compare their `toString()`s.

---

And just a little side note on why we introduced `useStructuralMemo`: to make things avoid re-rendering in the usual case of how users will likely write this code:

```js
// Users are likely to write this
useFetch("...", { ...options })

// But to avoid re-renders without (structural memo) they'd have to do something like this
const options = useMemo(() => { ...options}, [...deps])
useFetch("...", options)
```

packages/react/src/useFetch.ts
```diff
@@ -190,11 +206,17 @@ export function useFetch<T = string>(path: string, options?: FetchHookOptions):
                 }
               }
             }
-          })().catch((error) => {
-            if (!abortContoller.signal.aborted) {
-              dispatchError(mounted, dispatch, error, response);
-            }
-          });
+
+            mergedOptions.onStreamComplete?.();
```
thegedge, line 210: Another approach here is for the user to do this:

```js
const previousStreaming = useRef(streaming);

useEffect(() => {
  if(previousStreaming.current && !streaming) {
    // on stream complete
  }
  previousStreaming.current = streaming;
}, [streaming])
```

Which would avoid your callback issue, but kind of annoying for users to have to write. I guess it doesn't deal with aborts either, especially the ones that would be ignored by placing it here inside of the `try`

(scrap the idea, but leaving the comment here in case it sparks anything)

airhorns (approved):

thegedge (approved):

packages/react/src/useFetch.ts
thegedge: Oh, I completely missed that right below this comment was a `useStructuralMemo` 🤦

PR #251, opened 2023-08-16
Fix typecheck in all packages after switch to ESM

airhorns:
When I switched js-clients to start publishing both CJS and ESM, I had to mess with the tsconfigs and add separate ones for each build mode. This works fine for the `src` directory of each project, but I missed the fact that because those tsconfigs only apply to the src dirs, running a build only typechecks the `src` dirs. Nothing in CI was typechecking the `spec` dirs or the test bundless, whoops!

This adds an explicit new CI step to typecheck all the files in every package that has a `typecheck` script defined, and fixes a few errors that crept in.

### DIFF
--- a/packages/test-bundles/build.ts
+++ b/packages/test-bundles/build.ts
@@ -2,7 +2,7 @@ import react from "@vitejs/plugin-react-swc";
 import fs from "fs";
 import { join, parse } from "path";
 import { visualizer } from "rollup-plugin-visualizer";
-import { fileURLToPath } from "url";
+import { URL, fileURLToPath } from "url";
 import { build } from "vite";
 
 const bundleDir = fileURLToPath(new URL("bundles", import.meta.url));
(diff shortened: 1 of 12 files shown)


packages/tiny-graphql-query-compiler/src/index.ts
```diff
@@ -86,7 +86,7 @@ export interface VariableOptions {
   type: string;
   name?: string;
   value?: any;
-  required?: string;
+  required?: boolean;
```
airhorns, line 89: This was a mistake -- when I re-added this property to support the metadata coming from the client generator I just got this type wrong, it should be boolean.

packages/test-bundles/build.ts
```diff
@@ -2,7 +2,7 @@ import react from "@vitejs/plugin-react-swc";
 import fs from "fs";
 import { join, parse } from "path";
 import { visualizer } from "rollup-plugin-visualizer";
-import { fileURLToPath } from "url";
+import { URL, fileURLToPath } from "url";
```
thegedge, line 5: Surprised this one isn't coming in from the DOM bit of `"lib": ["es2020", "DOM"]`, but maybe I'm misunderstanding that option!

thegedge (approved):
🚀

packages/test-bundles/build.ts
airhorns: Same, i think this is just how the node types are, the global `URL` is actually a different, incompatible type from this one

PR #252, opened 2023-08-16
Allow passing no mutation variables to the react hook

airhorns:
Antoine and I both created param-less actions recently, and went to run them using `mutate()` as returned by the react hook. Before this change, that fails with an internal error inside the hook, where we assumed that the variables were present. If you pass `mutate({})`, things work fine, but that seems superfluous. The types should have caught this if we were working in TS, but since we were in the Gadget editor, the type check is off.

So, this changes the types and the runtime code to allow passing nothing at all to the mutate function returned by the hook if there are no required variables. We're still just the client here, so the server will still validate that any really required params are present and return an error if they aren't passed.

### DIFF
--- a/packages/react/spec/useAction.spec.tsx
+++ b/packages/react/spec/useAction.spec.tsx
@@ -13,7 +13,7 @@ import { MockClientWrapper, createMockUrqlCient, mockUrqlClient } from "./testWr
 
 describe("useAction", () => {
   // these functions are typechecked but never run to avoid actually making API calls
-  const TestUseActionCanRunActionsWithVariables = () => {
+  const TestUseActionCanRunUpdateActionsWithVariables = () => {
     const [_, mutate] = useAction(relatedProductsApi.user.update);
 
     // can call with variables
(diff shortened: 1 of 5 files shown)


packages/react/src/useAction.ts
```diff
@@ -75,7 +75,8 @@ export const useAction = <
   return [
     transformedResult,
     useCallback(
-      async (variables, context) => {
+      async (variables: F["variablesType"], context?: Partial<OperationContext>) => {
```
airhorns, line 78: I think I freaked out typescript too much by changing the return type of this outer function to require an inference, this just fixes internal type errors within these functions but doesn't change the output type

packages/react/spec/useAction.spec.tsx
```diff
@@ -32,6 +32,22 @@ describe("useAction", () => {
...
+
+    // can call with no model variables
+    void mutate({});
+
+    // can call with no variables at all
+    void mutate();
+
+    // @ts-expect-error can't call with variables that don't belong to the model
+    void mutate({ foo: "123" });
+  };
```
thegedge, line 49: 👌

airhorns:
ping @jasong689

airhorns:
ping @jasong689

PR #255, opened 2023-08-22
Support custom return types from actions

jasong689:
We need to update the action runner and builders to take `hasReturnType` into consideration when generating graphQL and selecting the result. This also updates the hooks to do the same. To release this we should first release api-client-core, then update gadget, and then react

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/operationBuilders.ts
+++ b/packages/api-client-core/src/operationBuilders.ts
@@ -131,7 +131,9 @@ export const actionOperation = (
   modelSelectionField: string,
   variables: VariablesOptions,
   options?: SelectionOptions | null,
-  namespace?: string | null
+  namespace?: string | null,
+  isBulkAction?: boolean | null,
+  hasReturnType?: boolean | null
 ) => {
   const selection = options?.select || defaultSelection;
 
@@ -139,7 +141,8 @@ export const actionOperation = (
     [operation]: Call(variableOptionsToVariables(variables), {
       success: true,
       errors: ErrorsSelection,
-      [modelSelectionField]: selection ? fieldSelectionToQueryCompilerFields(selection, true) : false,
+      [modelSelectionField]: selection && !hasReturnType ? fieldSelectionToQueryCompilerFields(selection, true) : false,
+      [isBulkAction ? "results" : "result"]: !!hasReturnType,
     }),
   };
 
--- a/packages/api-client-core/src/operationRunners.ts
+++ b/packages/api-client-core/src/operationRunners.ts
@@ -86,6 +86,32 @@ export const findManyRunner = async <Shape extends RecordShape = any>(
 };
 
 export interface ActionRunner {
+  (
+    modelManager: { connection: GadgetConnection },
+    operation: string,
+    defaultSelection: FieldSelection | null,
+    modelApiIdentifier: string,
+    modelSelectionField: string,
+    isBulkAction: false,
+    variables: VariablesOptions,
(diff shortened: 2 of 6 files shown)


packages/api-client-core/src/operationBuilders.ts
```diff
@@ -131,15 +131,18 @@ export const actionOperation = (
   modelSelectionField: string,
   variables: VariablesOptions,
   options?: SelectionOptions | null,
-  namespace?: string | null
+  namespace?: string | null,
+  isBulkAction?: boolean | null,
```
jasong689, line 135: This was added since bulk operations have a `results` field instead of `result`. We could avoid this if we generated a different default selection based on whether or not an action has a return type. Not sure if that's needed since the `actionRunner` already takes an `isBulkAction` arg already.

kristianpd:
`api-client-core` : 
https://codeload.github.com/gadget-inc/js-clients/tar.gz/refs/tags/%40gadgetinc/api-client-core-v0.15.3-gitpkg-4f5bb62

`react`: 
https://codeload.github.com/gadget-inc/js-clients/tar.gz/refs/tags/%40gadgetinc/react-v0.14.2-gitpkg-4f5bb62

packages/react/src/useAction.ts
```diff
@@ -47,7 +47,11 @@ export const useAction = <
   action: F,
   options?: LimitToKnownKeys<Options, F["optionsType"]>
 ): ActionHookResult<
-  GadgetRecord<Select<Exclude<F["schemaType"], null | undefined>, DefaultSelection<F["selectionType"], Options, F["defaultSelection"]>>>,
+  F["hasReturnType"] extends true
+    ? any
+    : GadgetRecord<
+        Select<Exclude<F["schemaType"], null | undefined>, DefaultSelection<F["selectionType"], Options, F["defaultSelection"]>>
+      >,
```
kristianpd, line 54: in my 🎩 i am currently seeing `Promise<any>` here, maybe it's a version bump problem

packages/api-client-core/src/operationRunners.ts
```diff
@@ -86,6 +86,32 @@ export const findManyRunner = async <Shape extends RecordShape = any>(
...
+    operation: string,
+    defaultSelection: FieldSelection | null,
+    modelApiIdentifier: string,
+    modelSelectionField: string,
+    isBulkAction: false,
+    variables: VariablesOptions,
+    options?: SelectionOptions | null,
+    namespace?: string | null,
+    hasReturnType?: true
+  ): Promise<any>;
```
kristianpd, line 100: ```Promise<Shape extends void ? void : GadgetRecord<Shape>>;```?

packages/api-client-core/src/operationRunners.ts
```diff
@@ -109,6 +135,32 @@ export interface ActionRunner {
...
+    operation: string,
+    defaultSelection: FieldSelection | null,
+    modelApiIdentifier: string,
+    modelSelectionField: string,
+    isBulkAction: true,
+    variables: VariablesOptions,
+    options?: SelectionOptions | null,
+    namespace?: string | null,
+    hasReturnType?: true
+  ): Promise<any[]>;
```
kristianpd, line 150: ```Promise<Shape extends void ? void : GadgetRecord<Shape>>;```?

kristianpd:
The types error we think is due to the `cdn.types.dev` being unable to resolve `0.15.4` let's publish it and test once more before bumping anything in userlandependencies

PR #258, opened 2023-09-05
Add `exchanges` option

scott-rc:
This adds an `exchanges` parameter to our client constructor so we can add custom ones in the sandbox. I decided to go with 3 hooks (beforeAll, beforeAsync, afterAll) based on the docs below:

https://formidable.com/open-source/urql/docs/advanced/authoring-exchanges/#synchronous-first-asynchronous-last

This API doesn't give us complete control over the exchanges (e.g. re-arrange, remove, replace), so if you think we will want to do any of that we should add it now.

## PR Checklist


airhorns (approved):
I think this is grand and I think this is the right level of control for now.

PR #260, opened 2023-09-05
Improve bulk error handling by throwing an `ErrorGroup`

airhorns:
While working on adding support for bulk actions that return many results and/or many errors, I noticed our bulk action error handling totally sucked. We didn't ever return any errors regardless of if they occurred. Instead, I think if any errors are returned, we should throw them to the caller so they know it happened. If any results do come back, they might want to know that too, some stuff worked and some stuff didn't. So, we decorate the error group with a `.results` property that has those bulk action results on it as well.

This is technically a breaking change, but I think its best considered a bug fix -- errors were just being suppressed before, which was a mistake from a couple years ago. I think that if things are erroring in prod, they should throw, as is already the case for single actions.

## PR Checklist

- [x] Any immediate changes are slated for release in Gadget via a generated package dependency bump


scott-rc (approved):
:shipit:

packages/api-client-core/src/operationRunners.ts
```diff
@@ -191,19 +193,29 @@ export const actionRunner: ActionRunner = async <Shape extends RecordShape = any
...
-  // if there's nothing at `mutationResult[modelSelectionField]`, but the caller isn't expecting a return (void).
-  if (defaultSelection == null) {
-    return;
-  }
-
-  // todo this does not support pagination params right now, we'll need to add it to bulk action Results
-  if (isBulkAction) {
-    return !hasReturnType ? hydrateRecordArray<Shape>(response, mutationResult[modelSelectionField]) : mutationResult.results;
+  if (!isBulkAction) {
+    const mutationTriple = assertMutationSuccess(response, dataPath);
```
jasong689, line 198: What is a mutation triple?

jasong689 (approved):

packages/api-client-core/src/operationRunners.ts
airhorns: success / errors / result

PR #265, opened 2023-09-18
Fix bulk action input processing to deal with ambiguous identifiers and nested params

airhorns:
Finally figured out a way to mock these tests to test these hooks!

### DIFF
--- a/packages/react/spec/useBulkAction.spec.ts
+++ b/packages/react/spec/useBulkAction.spec.ts
@@ -58,7 +58,7 @@ describe("useBulkAction", () => {
     expect(result.current[0].error).toBeFalsy();
   });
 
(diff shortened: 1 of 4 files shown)


packages/react/spec/useBulkAction.spec.ts
```diff
@@ -102,6 +102,255 @@ describe("useBulkAction", () => {
...
+
+    await act(async () => {
+      const promiseResult = await mutationPromise;
+      expect(promiseResult.data!.length).toEqual(2);
+      expect(promiseResult.data![0].id).toEqual("123");
+      expect(promiseResult.data![1].id).toEqual("124");
+    });
+
+    expect(result.current[0].data!.length).toEqual(2);
+    expect(result.current[0].data![0].id).toEqual("123");
```
MillanWangGadget, line 348: Would it be possible to add some checks for the updated field values here?

PR #267, opened 2023-09-25
Fix for useAction incorrectly mapping non nested auth action inputs

jcao49:
`useAction` was nesting variables under the action model api identifier, but the graphql api for these actions were expecting them to be non-nested (`api.user.signUp({email..})` is not nested under `user` like it would be in `api.user.signUp({user: {email:...}`). This caused us to pass in the variables to graphql in a shape it was not expecting (and could not parse the data for)!

## PR Checklist

### DIFF
--- a/packages/react/spec/useAction.spec.tsx
+++ b/packages/react/spec/useAction.spec.tsx
@@ -417,7 +417,7 @@ describe("useAction", () => {
     `);
   });
 
-  test("generates correct mutation and variables for fat trigger", async () => {
+  test("generates correct mutation and variables for trigger that contributes inputs", async () => {
     let variables: AnyVariables;
 
     const client = createMockUrqlClient({
--- a/packages/react/spec/useAction.spec.tsx
+++ b/packages/react/spec/useAction.spec.tsx
@@ -8,7 +8,7 @@ import type { AnyVariables } from "urql";
 import { Provider } from "../src/GadgetProvider.js";
 import { useAction } from "../src/index.js";
 import type { ErrorWrapper } from "../src/utils.js";
-import { relatedProductsApi } from "./apis.js";
+import { relatedProductsApi, superAuthApi } from "./apis.js";
 import { MockClientWrapper, createMockUrqlClient, mockUrqlClient } from "./testWrappers.js";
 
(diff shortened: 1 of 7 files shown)


packages/react/src/utils.ts
```diff
@@ -327,9 +327,10 @@ export const disambiguateActionVariables = (
 
   if (action.acceptsModelInput || action.hasCreateOrUpdateEffect) {
     if (
-      action.modelApiIdentifier in variables &&
-      typeof variables[action.modelApiIdentifier] === "object" &&
-      variables[action.modelApiIdentifier] !== null
+      (action.modelApiIdentifier in variables &&
+        typeof variables[action.modelApiIdentifier] === "object" &&
+        variables[action.modelApiIdentifier] !== null) ||
+      !action.variables[action.modelApiIdentifier]
```
infiton, line 333: is there a test we can add for this?

jasong689: I was thinking we could add a test post release of the email pw feature since those are the only triggers that currently have params that aren't either model input or exported params. We would need to 🎩 this with the trigger though

infiton: :+1:

jasong689: Now that I'm thinking about it a bit more, maybe it's better to include a test for this now. @jcao49 maybe we can add the signUp trigger to one of the existing app's that we have a client for and use that as a test case

packages/react/spec/useAction.spec.tsx
```diff
@@ -417,6 +417,54 @@ describe("useAction", () => {
     `);
   });
 
+  test("generates correct mutation and variables for fat trigger", async () => {
```
infiton, line 420: nit: since this is open source, may be worth using a more descriptive name (most folks won't know what a "fat" trigger is)

infiton (approved):

packages/react/spec/useAction.spec.tsx
jasong689: "trigger that contributes inputs"

jasong689 (approved):
👍 you can either bump the version in this PR on or main after merging

PR #270, opened 2023-09-27
Bump graphql dependency to latest

airhorns:
Superseeds https://github.com/gadget-inc/js-clients/pull/266, resolves a security advisory when installing the client


airhorns:
Integration tests passed gadget side, merging

PR #271, opened 2023-09-27
Fix react bulk actions with flat params

airhorns:
Support for invoking bulk actions with flattened params in react was broken, and https://github.com/gadget-inc/js-clients/pull/267 accidentally broke it further.

It was broken since inception because the tests were asserting incorrectly that the produced `input` params weren't nested -- they should have been, oops! Updated the tests and stopped using snapshot tests to describe this so no one is tempted to break it.

Then, to fix the issue and properly disambiguate the params, I had to change the behaviour of the disambiguator a little bit. To preserve the behaviour for the change in https://github.com/gadget-inc/js-clients/pull/267, we need to respect the idea that if an action does not accept model input, we shouldn't do any nesting of variables. This does this by giving the new `action.acceptsModelInput` value precedence over `action.hasCreateOrUpdateEffect`, which was the ancient way of asking this question. That old way keeps support for old API clients, but the new way is more trustworthy. For things like the signUp action, `action.acceptsModelInput` is false, but `action.hasCreateOrUpdateEffect` is true. So, we stop `||`-ing them, and instead give the new thing precedence. This way, for actions with hardcoded param sets like `user.signUp`, we never nest.

This fixes the over-eager nesting issue for bulk actions as well.

### DIFF
--- a/packages/react/src/utils.ts
+++ b/packages/react/src/utils.ts
@@ -323,14 +323,12 @@ export const disambiguateActionVariables = (
   }
 
   let newVariables: Record<string, any>;
-  const idVariable = Object.entries(action.variables).find(([key, value]) => key === "id" && value.type === "GadgetID");
 
-  if (action.acceptsModelInput || action.hasCreateOrUpdateEffect) {
+  if (action.acceptsModelInput ?? action.hasCreateOrUpdateEffect) {
     if (
-      (action.modelApiIdentifier in variables &&
-        typeof variables[action.modelApiIdentifier] === "object" &&
-        variables[action.modelApiIdentifier] !== null) ||
-      !action.variables[action.modelApiIdentifier]
+      action.modelApiIdentifier in variables &&
+      typeof variables[action.modelApiIdentifier] === "object" &&
+      variables[action.modelApiIdentifier] != null
     ) {
       newVariables = variables;
     } else {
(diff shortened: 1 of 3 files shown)


airhorns:
ping @jcao49 @jasong689

packages/react/src/utils.ts
```diff
@@ -323,14 +323,12 @@ export const disambiguateActionVariables = (
   }
 
   let newVariables: Record<string, any>;
-  const idVariable = Object.entries(action.variables).find(([key, value]) => key === "id" && value.type === "GadgetID");
 
-  if (action.acceptsModelInput || action.hasCreateOrUpdateEffect) {
+  if (action.acceptsModelInput ?? action.hasCreateOrUpdateEffect) {
```
jcao49, line 327: 🧠

jcao49 (approved):
the power of nullish coalescing 😖💯

packages/react/src/utils.ts
jasong689, line 327: Hm now that I think about this a bit more, this makes less sense. `??` is nullish coalescing so `acceptsModelInput` needs to return `null` or `undefined` but it never will

jasong689: I think this will hit old apps on effect cards that have a newly generated client which has `acceptsModelInput` defined

airhorns: I think `acceptsModelInput` will be defined for newly generated api clients, but it will be undefined for old ones that are locked in time in the package.json. So, the check for `hasCreateOrUpdateEffect` is a fallback for clients that haven't been generated recently. In more recent clients though, `acceptsModelInput` is trustworthy, it correctly reports for AAC actions and the new `actionType: "custom"` stuff, as well as bulk actions, whereas the old property doesn't.

jasong689 (approved):

PR #274, opened 2023-09-28
Ensure hook pausing with urql works

airhorns:
urql hooks have a `pause` option that allows preventing sending the initial request. We should respect this option and pass it through to urql and avoid doing anything that requires never pausing a request.

## PR Checklist

### DIFF
--- a/packages/react/spec/useFindFirst.spec.ts
+++ b/packages/react/spec/useFindFirst.spec.ts
@@ -170,4 +170,16 @@ describe("useFindFirst", () => {
     rerender();
     expect(result.current[0]).toBe(beforeObject);
   });
+
+  test("doesn't issue a request if paused", async () => {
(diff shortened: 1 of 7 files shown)


infiton:
I'm confused: pause works right? Do the tests you added fail without the changes to the src?

It looks like the src changes are to have pause play nicely with some error handling, should there be tests that cover that?

airhorns:
Sorry not a good description! Before this change, if you passed pause to a hook, we'd pass it down to URQL, and it'd return a data object with `data: null, fetching: false, error: null`, which our stuff has never seen before. It would interpret that as "we're not fetching and there's no data, so something must have gone wrong", even though there was no error. The tests I added do indeed fail without the changes to teach the error handler to not throw if the hook is paused.

infiton:
gotcha; and I guess you wouldn't see this if you didn't destructure error

infiton (approved):

PR #284, opened 2023-10-11
Fix support for POST requests run in strict mode from useFetch

airhorns:
This fixes a very subtle bug in `useFetch`. When running a fetch request with `shouldSendImmediately: false` and `json: true`, users would get an error looking like this:

```
Uncaught (in promise) ErrorWrapper: [GraphQL] The user aborted a request.
    at _ErrorWrapper.forClientSideError (https://fetch-testing--development.gadget.app/node_modules/.vite/deps/@gadgetinc_react.js?v=19130701:484:12)
    at dispatchError (https://fetch-testing--development.gadget.app/node_modules/.vite/deps/@gadgetinc_react.js?v=19130701:966:32)
    at https://fetch-testing--development.gadget.app/node_modules/.vite/deps/@gadgetinc_react.js?v=19130701:1048:25
```

This is because the POST request was immediately, erroneously being aborted as soon as it was run. We don't want to abort the request, we want to run it!

We have a bunch of existing code to abort requests on component unmount for good hygiene. We don't need the response for a component that is no longer on screen and we don't want to charge users (or make expensive openai requests) that take a long time that no one wants anymore. So, we use a useEffect teardown to abort defunct requests. This works good, but, the bug stemmed from the dependencies to that useEffect changing at the wrong time. We were accidentally mutating one of the options objects that gets passed to that useEffect as a dependency on request start, which had the effect of triggering the useEffect to do its unmount/remount thing immediately on the next render. Simply running the request should not ever cause the dependencies to change.

Most confusingly, this mutate-the-options-and-trigger-a-depenendcies-change only happened when you passed a different option in to the hook (`json: true`) which I think is why this bug lurked so long.

The fix is to just not mutate the options object we're using as a dependency in the dependencies array, which means no dependencies change and the useEffect doesn't abort the request. We keep it around though for the actual unmount case.

This also showed a bit of a test hole, which I fixed. useEffects are also pains in the ass for react strict mode which I initially thought was the culprit. I kept those tests around, and added a test that passes in `json` and failed before these changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "devDependencies": {
     "@gadget-client/app-with-no-user-model": "^1.5.0",
     "@gadget-client/bulk-actions-test": "^1.104.0",
-    "@gadget-client/related-products-example": "^1.853.0",
+    "@gadget-client/related-products-example": "^1.854.0",
     "@gadget-client/super-auth": "^1.37.0",
     "@gadgetinc/api-client-core": "workspace:*",
     "@gadgetinc/eslint-config": "^0.6.1",
--- a/packages/api-client-core/spec/mockUrqlClient.ts
+++ b/packages/api-client-core/spec/mockUrqlClient.ts
@@ -41,6 +41,7 @@ export type MockOperationFn = jest.Mock & {
 export type MockFetchFn = jest.Mock & {
   requests: { args: any[]; resolve: (response: Response) => void; reject: (error: Error) => void }[];
   pushResponse: (response: Response) => Promise<void>;
+  reportAbort: () => Promise<void>;
 };
 
 export interface MockUrqlClient extends Client {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
   "devDependencies": {
     "@gadget-client/app-with-no-user-model": "^1.5.0",
     "@gadget-client/bulk-actions-test": "^1.104.0",
-    "@gadget-client/related-products-example": "^1.853.0",
+    "@gadget-client/related-products-example": "^1.854.0",
```
airhorns, line 26: Was proving it was still an issue on the most recent version of the generated clients

packages/api-client-core/spec/mockUrqlClient.ts
```diff
@@ -134,9 +135,30 @@ const newMockFetchFn = () => {
...
+      const signal = request.args[1]?.signal;
+      if (!signal) {
+        throw new Error("no signal on request, can't report an abort that has happened");
+      }
+      if (!signal.aborted) {
+        throw new Error("signal on request has not been aborted, can't report an abort that has happened");
+      }
+      await request.reject(new Error("AbortError: The user aborted a request."));
+    });
+  };
```
airhorns, line 161: Because we're mocking the browser fetch, we're responsible for actually implementing aborts if the abort signal fires. This allows us to do that.

airhorns:
Fixes GGT-4990

infiton (approved):
bet that was fun to debug 🙏

PR #285, opened 2023-10-12
Add support for sending the shopify session token via websocket requests in the React provider

airhorns:
The Gadget client already knows how to talk websockets to the Gadget backend, but the custom auth mode we use in the React provider didn't have support for passing the Shopify JWT for this. Now it does!


airhorns:
🎩 'd this locally, shipping

PR #286, opened 2023-10-12
introduces useActionForm

infiton:
This PR introduces `useActionForm` that will allow us to use `react-form-hook` together with `useAction`; this hook will help us create smaller starting files for our web apps. @airhorns this is mostly the code you wrote with a few changes:

- I'm not returning `modelManager` from the hook; let's leave it out for now and if we need it later add it in
- the hook returns `actionData`
- the promise returned when calling `submit` waits until the action has resolved 

## PR Checklist

### DIFF
--- a/packages/react/src/useActionForm.ts
+++ b/packages/react/src/useActionForm.ts
@@ -1,0 +1,332 @@
+import type {
+  ActionFunction,
+  AnyModelManager,
+  DefaultSelection,
+  GadgetRecord,
+  GlobalActionFunction,
+  Select,
+} from "@gadgetinc/api-client-core";
+import { camelize } from "@gadgetinc/api-client-core";
+import { useCallback, useEffect, useRef } from "react";
+import type { DeepPartial, FieldErrors, FieldValues, UseFormProps, UseFormReturn } from "react-hook-form";
+import { useForm } from "react-hook-form";
+import { useApi } from "./GadgetProvider.js";
+import { useAction } from "./useAction.js";
+import { useFindBy } from "./useFindBy.js";
+import { useFindOne } from "./useFindOne.js";
+import { useGlobalAction } from "./useGlobalAction.js";
+import type { ActionHookState, ErrorWrapper, OptionsType } from "./utils.js";
+import { get, set } from "./utils.js";
+
+export * from "react-hook-form";
+
+/** Finds a given record from the backend database by either id or a `{[field: string]: value}` slug */
+const useFindExistingRecord = (
+  modelManager: AnyModelManager | undefined,
+  findBy: string | { [key: string]: any },
+  options: { select?: Record<string, any>; pause?: boolean }
+): [{ data?: GadgetRecord<any>; fetching: boolean; error?: ErrorWrapper }, () => void] => {
+  // for simplicity, we conditionally call either the findBy or findOne hook. this violates the rules of hooks, but is a LOT simpler than mounting and pausing both hooks and massaging the results of each together. this means we don't support the same form toggling from being for a record found by id and then later a record found by some other criteria, but that's a very rare use case. you can workaround this by adding a `key` prop to the component calling `useActionForm`, and having the value change when the find method changes, which will give different instance of the component and avoid the hook order changing.
+  if (modelManager && typeof findBy === "string") {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    return useFindOne(modelManager as any, findBy, options);
+  } else if (modelManager) {
+    const [findByKey, findByValue] = Object.entries(findBy)[0];
+    const finder = (modelManager as Record<string, any>)[`findBy${camelize(findByKey)}`];
+
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    return (useFindBy as any)(finder, findByValue, options);
+  } else {
+    return [
+      { fetching: false },
+      () => {
+        // noop
+      },
+    ];
+  }
+};
+
+const OmittedKeys = ["__typename", "id", "createdAt", "updatedAt"] as const;
+type OmittedKey = (typeof OmittedKeys)[number];
+
+const omitKeys = (data: any) => {
+  const newData = { ...data };
+  for (const key of OmittedKeys) {
+    delete newData[key];
+  }
+  return newData;
+};
+
+const toDefaultValues = (modelApiIdentifier: string | undefined, data: any) => {
+  data = omitKeys(data);
+
+  if (modelApiIdentifier && data[modelApiIdentifier] && typeof data[modelApiIdentifier] === "object") {
+    data[modelApiIdentifier] = omitKeys(data[modelApiIdentifier]);
+  }
+
+  return data;
+};
+
+/**
+ * The identity of a record to build a form for
+ *
+ * - Pass the ID of a record as a string to look up a record by id
+ * - Pass a {<fieldValue>: <value>} object to look up a record by some other field. __Note__: a `findBy` function must exist for the field you wish to find the record by.
+ *
+ * @example
+ * // find the user record with ID=123
+ * const { submit } = useActionForm(api.user.update, { record: "123" });
+ *
+ * @example
+ * // find a record by email. `api.user.findByEmail` must exist
+ * const { submit } = useActionForm(api.user.update, { record: { email: "user@gadget.app" } });
+ */
+export type RecordIdentifier = string | { [key: string]: any };
+
+type UseActionFormHookStateData<F extends ActionFunction<any, any, any, any, any> | GlobalActionFunction<any>> = F extends ActionFunction<
+  any,
+  any,
+  any,
+  any,
+  any
+>
+  ? F["hasReturnType"] extends true
+    ? any
+    : GadgetRecord<
+        Select<Exclude<F["schemaType"], null | undefined>, DefaultSelection<F["selectionType"], F["optionsType"], F["defaultSelection"]>>
+      >
+  : any;
+
+type UseActionFormHookState<F extends ActionFunction<any, any, any, any, any> | GlobalActionFunction<any>> = ActionHookState<
+  UseActionFormHookStateData<F>,
+  Exclude<F["variablesType"], null | undefined>
+>;
+
+type UseActionFormSubmit<F extends ActionFunction<any, any, any, any, any> | GlobalActionFunction<any>> = (
+  event?: React.BaseSyntheticEvent<object, any, any> | undefined
+) => Promise<UseActionFormHookState<F>>;
+
+export type UseActionFormResult<
+  GivenOptions extends OptionsType,
+  SchemaT,
+  ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any> | GlobalActionFunction<any>,
+  FormVariables extends FieldValues,
+  FormContext = any
+> = Omit<UseFormReturn<FormVariables, FormContext>, "handleSubmit"> & {
+  /**
+   * Any error that occurred during initial data fetching or action submission
+   */
+  error?: ErrorWrapper | Error | null;
+  /**
+   * Function to call to submit the form
+   */
+  submit: UseActionFormSubmit<ActionFunc>;
+  /**
+   * The data resulting from running the action
+   */
+  actionData?: ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any>
+    ? ReturnType<typeof useAction<GivenOptions, SchemaT, ActionFunc, any>>[0]["data"]
+    : ReturnType<typeof useGlobalAction<any>>[0]["data"];
+};
+
+/**
+ * React hook to manage state for a form that calls a Gadget action. `useActionForm` must be passed an action function from an instance of your generated API client library, like `api.user.create`, `api.blogPost.publish` or `api.someGlobalAction`. `useActionForm` returns a `Form` object from `react-hook-form` which can be used to build great form experiences.
+ *
+ * `useActionForm` manages the Gadget API calls for the form, and handling validation errors returned by the Gadget backend. If your form is for an existing record, `useActionForm` will also fetch the record from the backend and populate the form with the existing values.
+ *
+ * `useActionForm` doesn't run the action when called. To actually submit the form, call the `submit` function on the returned `Form` object.
+ *
+ * @see https://react-hook-form.com/
+ *
+ * @param action an action function from a model manager in your application's client, like `api.user.create`
+ * @param options action options, like selecting the fields in the result
+ */
+export const useActionForm = <
+  GivenOptions extends OptionsType,
+  SchemaT,
+  ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any> | GlobalActionFunction<any>,
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  ExtraFormVariables = {},
+  FormContext = any,
+  FormVariables extends FieldValues = ActionFunc["variablesType"] & ExtraFormVariables,
+  ActionResultData = UseActionFormHookStateData<ActionFunc>
+>(
+  action: ActionFunc,
+  options?: Omit<UseFormProps<FormVariables, FormContext>, "defaultValues"> & {
+    defaultValues?: DeepPartial<FormVariables> & { [key in OmittedKey]?: any };
+    /**
+     * The record identifier to run this action on, if it already exists.
+     * Should be undefined for create actions, or a record ID (or finder) for update / etc actions
+     **/
+    findBy?: RecordIdentifier;
+    /**
+     * Which fields to select on from the record when retrieving it from the backend.
+     */
+    select?: ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any> ? ActionFunc["optionsType"]["select"] : never;
+    /**
+     * Which fields to send from the form's values when sending it from the backend.
+     */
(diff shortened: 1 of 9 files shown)


packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,261 @@
...
+import { get, set } from "./utils.js";
+
+/** Finds a given record from the backend database by either id or a `{[field: string]: value}` slug */
+const useFindExistingRecord = (
+  modelManager: AnyModelManager,
+  findBy: string | { [key: string]: any },
+  options: { select?: Record<string, any>; pause?: boolean }
+): [{ data?: GadgetRecord<any>; fetching: boolean; error?: ErrorWrapper }, () => void] => {
+  // for simplicity, we conditionally call either the findBy or findOne hook. this violates the rules of hooks, but is a LOT simpler than mounting and pausing both hooks and massaging the results of each together. this means we don't support the same form toggling from being for a record found by id and then later a record found by some other criteria, but that's a very rare use case. you can workaround this by adding a `key` prop to the component calling `useActionForm`, and having the value change when the find method changes, which will give different instance of the component and avoid the hook order changing.
+  if (typeof findBy === "string") {
```
jasong689, line 20: Could you get around this by always using `useFindBy` and use `id` as the field api identifier? `id` is a system field that can't be renamed

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,261 @@
...
+>(
+  action: ActionFunc,
+  options?: UseFormProps<ActionFunc["variablesType"], FormContext> & {
+    /**
+     * The record identifier to run this action on, if it already exists.
+     * Should be undefined for create actions, or a record ID (or finder) for update / etc actions
+     **/
+    findBy?: RecordIdentifier;
+    /**
+     *
```
airhorns, line 95: needs a comment here

infiton: :+1:

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,263 @@
...
+  ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any>,
+  FormContext = any
+>(
+  action: ActionFunc,
+  options?: UseFormProps<ActionFunc["variablesType"], FormContext> & {
+    /**
+     * The record identifier to run this action on, if it already exists.
+     * Should be undefined for create actions, or a record ID (or finder) for update / etc actions
+     **/
+    findBy?: RecordIdentifier;
```
jasong689, line 93: I wonder if it would be more clear with something like
```typescript
& { findBy?: RecordIdentifier; record: never } | { record?: ActionFunc["hasReturnType"] extends true ...., findBy: never }
```
not sure if that would typecheck but makes it obvious it's one or the other. Another suggestion is maybe only include these if `operatesWithRecordIentity` is true

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,263 @@
...
+    onError?: (error: Error | FieldErrors<ActionFunc["variablesType"]>) => void;
+  }
+): UseActionFormResult<GivenOptions, SchemaT, ActionFunc, FormContext> => {
+  const api = useApi();
+  const findExistingRecord = !!options?.findBy;
+  const hasSetInitialValues = useRef<boolean>(!findExistingRecord);
+
+  // find the existing record if there is one
+  const modelManager = (api as any)[action.modelApiIdentifier] as AnyModelManager;
+  const [findResult] = useFindExistingRecord(modelManager, options?.findBy || "1", {
```
jasong689, line 137: Is "1" the correct default we want for this?

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  const submit = useCallback(
+    async (event?: React.BaseSyntheticEvent<object, any, any> | undefined) => {
+      let result: any;
+
+      formHook.clearErrors();
+
+      await handleSubmit(
+        async (data) => {
+          let modelInput = data[action.modelApiIdentifier];
```
jasong689, line 200: Hard to tell but how does this case arise? How do you set properties on the data such that the inputs are set on the model api identifier? I guess how would you use `useActionForm` with a model that has ambiguous identifiers.

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,263 @@
...
+      )(event);
+
+      if (result?.error) {
+        handleSubmissionError(result.error);
+      }
+    },
+    [handleSubmit, formHook, handleSubmissionError, existingRecord, options, action.modelApiIdentifier, runAction]
+  );
+
+  const proxiedFormState = new Proxy(formState, {
```
airhorns, line 242: Is there a way to avoid this proxy? It seems kinda complicated and I find throws newer developers for a loop when you console.log it or see it in the debugger. Whats it for?

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,261 @@
...
+    /**
+     * The record identifier to run this action on, if it already exists.
+     * Should be undefined for create actions, or a record ID (or finder) for update / etc actions
+     **/
+    findBy?: RecordIdentifier;
+    /**
+     *
+     */
+    record?: ActionFunc["hasReturnType"] extends true
+      ? any
```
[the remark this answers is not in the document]
infiton: ah I think I was thinking that the input should match the shape of the data that gets returned from the action

packages/react/src/useActionForm.ts
infiton: the formState from useForm does some weird proxy stuff, from https://react-hook-form.com/docs/useform/formstate

```
formState is wrapped with a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to improve render performance and skip extra logic if specific state is not subscribed to. Therefore make sure you invoke or read it before a render in order to enable the state update.```

I was initially thinking I would create my own state from the useForm state but I didn't want to mess with any optimizations they were doing

would be open to any suggestions on better ways to do this derived state though

Comment (airhorns) [reply packages/react/src/useActionForm.ts:242]: Ah, read a bit more about that and I think you're right, we need to keep it as a proxy to take advantage of their nice only-re-render-if-you-touch-the-thing stuff. Fancy!

File: packages/react/src/useActionForm.ts (hunk 1 above)
Comment (infiton) [reply packages/react/src/useActionForm.ts]: yah I think so, but now with global actions we gotta do a conditional hook anyways

File: packages/react/src/useActionForm.ts (hunk 3 above)
Comment (infiton) [reply packages/react/src/useActionForm.ts:131]: got rid of the record so this should be more clear now

File: packages/react/src/useActionForm.ts (hunk 4 above)
Comment (infiton) [reply packages/react/src/useActionForm.ts:160]: I think this is a hack since we always need to pass a value in here; if there is no `findBy` then the hook is always paused

File: packages/react/src/useActionForm.ts (hunk 5 above)
Comment (infiton) [reply packages/react/src/useActionForm.ts]: added tests for this 👍

File: packages/react/src/useActionForm.ts (hunk 4 above)
Comment (airhorns) [reply packages/react/src/useActionForm.ts:160]: A hack indeed, maybe we should just make the value optional and that'd be clearer?

File: packages/react/src/useActionForm.ts [hunk 8]
```diff
@@ -0,0 +1,293 @@
...
+import { useForm } from "react-hook-form";
+import { useApi } from "./GadgetProvider.js";
+import { useAction } from "./useAction.js";
+import { useFindBy } from "./useFindBy.js";
+import { useFindOne } from "./useFindOne.js";
+import { useGlobalAction } from "./useGlobalAction.js";
+import type { ErrorWrapper, OptionsType } from "./utils.js";
+import { get, set } from "./utils.js";
+
+export * from "react-hook-form";
```
airhorns, line 14: 🫡

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,293 @@
...
+  FormContext = any
+> = Omit<UseFormReturn<FormVariables, FormContext>, "handleSubmit"> & {
+  /**
+   * Any error that occurred during initial data fetching or action submission
+   */
+  error?: ErrorWrapper | Error | null;
+  /**
+   * Function to call to submit the form
+   */
+  submit: () => Promise<void>;
```
airhorns, line 94: If submit is gonna return stuff, can we type it here? 

I think returning stuff is important so that folks can naturally do an `const record = await submit(); redirectTo(record.id` sort of thing

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,293 @@
...
+     */
+    send?: string[];
+    /**
+     * Called when the form submits
+     */
+    onSubmit?: () => void;
+    /**
+     * Called when the action completes successfully on the backend
+     */
+    onSuccess?: (actionResult: any) => void;
```
airhorns, line 147: Can we type the result? I think we type it elsewhere too

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,293 @@
...
+                  : validationError.apiIdentifier;
+              formHook.setError(errorKey as any, {
+                message: validationError.message,
+              });
+            }
+          } else {
+            const codeToReplace = "code" in executionError ? `${executionError.code}: ` : "";
+            const message = executionError.message.replace(codeToReplace, "");
+
+            formHook.setError("root", { message });
```
airhorns, line 220: I like this pattern, didn't notice you added this! I think we should put transport errors in here too if we're gonna do it like this, like if the network is disconnected or what have you. Can you verify that's working ok?

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,293 @@
...
+      } else {
+        return (target as any)[prop];
+      }
+    },
+  });
+
+  return {
+    ...formHook,
+    formState: proxiedFormState,
+    error: findResult.error || actionResult.error,
```
airhorns, line 289: Do we need to return this if we're going to put errors at the root? Kind of annoying to make the user check two places?

airhorns (commented):
This is gonna be so good.

packages/react/src/useActionForm.ts
infiton: whoops yup!

infiton: in the docs it looks like you expected it to return `result.data` from `useAction`? Or should we return the same as the useAction mutation?

packages/react/src/useActionForm.ts
infiton: :+1:

packages/react/src/useActionForm.ts
infiton: yah this was a big win for the auth templates; transport here meaning no `code`?

packages/react/src/useActionForm.ts
infiton: I think since we only show the message on the formState (and also I haven't put the `findResult.error` there) it would be useful to have the full errors as well

packages/react/src/useActionForm.ts
airhorns: Hm, I guess being consistent would be best, I am down for returning the whole result instead of just the data

packages/react/src/useActionForm.ts
airhorns: Yes no `code`, but also I think the if statement at the top here concerning `"executionErrors" in error` might be ignoring errors where the network doesn't work at all. I think that "executionErrors" is an `urql` thing, and there can also be a `networkError` that comes when there's no connection made at all or a timeout or whatever. I think the move would just be to add a test and see what happens. I took a look but I couldn't find a case where we do this kind of transport error already :(

packages/react/src/useActionForm.ts
infiton: :+1: putting me through TS paces today; I've typed these

packages/react/src/useActionForm.ts
infiton: done

packages/react/src/useActionForm.ts
infiton: yah was looking for that; I think `executionErrors` is our thing?

```/**
   * A list of errors encountered by the backend when processing a request. Populated if the client successfully communicated with the backend, but the backend was unable to process the request and rejected it with an error.
   * Includes GraphQL syntax errors, missing or invalid argument errors, data validation errors, or unexpected errors encountered when running backend logic.
   **/
```

I think maybe we don't want the network errors in the `formState`? I think this would be the reason to have the top level `error` as well... i.e. you watch that to put up a something went wrong toast, but you don't show the network error as a form server side validation error

infiton: I found a `mockNetworkError`

infiton: ok added a test asserting that 

```
expect((result.current.error as any).message).toMatchInlineSnapshot(`"[Network] Network error"`);
expect(result.current.formState.errors).toEqual({});
```

I think that's right to me, showing network error as a form error seems wrong, but I don't feel strongly about that

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,324 @@
...
+
+export * from "react-hook-form";
+
+/** Finds a given record from the backend database by either id or a `{[field: string]: value}` slug */
+const useFindExistingRecord = (
+  modelManager: AnyModelManager | undefined,
+  findBy: string | { [key: string]: any },
+  options: { select?: Record<string, any>; pause?: boolean }
+): [{ data?: GadgetRecord<any>; fetching: boolean; error?: ErrorWrapper }, () => void] => {
+  // for simplicity, we conditionally call either the findBy or findOne hook. this violates the rules of hooks, but is a LOT simpler than mounting and pausing both hooks and massaging the results of each together. this means we don't support the same form toggling from being for a record found by id and then later a record found by some other criteria, but that's a very rare use case. you can workaround this by adding a `key` prop to the component calling `useActionForm`, and having the value change when the find method changes, which will give different instance of the component and avoid the hook order changing.
```
jasong689, line 29: Maybe a note in the docs about this would be useful

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,324 @@
...
+  if (modelManager && typeof findBy === "string") {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    return useFindOne(modelManager as any, findBy, options);
+  } else if (modelManager) {
+    const findByKey = Object.keys(findBy)[0];
```
jasong689, line 34: A note about this would be good too

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,324 @@
...
+    const findByValue = Object.values(findBy)[0];
```
jasong689, line 35: nit: I'm not 100% sure if iteration order is guaranteed across browsers so maybe better to use `Object.entries` to get both key and value?

packages/react/src/useActionForm.ts
```diff
@@ -0,0 +1,324 @@
...
+     **/
+    findBy?: RecordIdentifier;
+    /**
+     * Which fields to select on from the record when retrieving it from the backend.
+     */
+    select?: ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any> ? ActionFunc["optionsType"]["select"] : never;
+    /**
+     * Which fields to send from the form's values when sending it from the backend.
+     */
+    send?: string[];
```
jasong689, line 170: Should these be typed against available fields?

packages/react/src/useActionForm.ts
jasong689: hm yea. the nice thing about this is it works if a user goes from not passing in a string to passing one in. won't work for `{ [key: string]: any }` findBys though

jasong689 (commented):
🎩 :
- basic model with create, update, delete ✅ 
- global action with params ✅ 
- model with ambiguous field 🔴 
  - created a model named `ambiguous`
  - created 2 string fields named `ambiguous` and `string`
```typescript
import { api } from "../api";
import { useActionForm } from "@gadgetinc/react";

export default function () {
  const {
    submit,
    register,
    formState: { errors }
  } = useActionForm(api.ambiguous.create);

  return (<form onSubmit={submit}>
    <input {...register("ambiguous.ambiguous")} />
    <input {...register("ambiguous.string")} />
    <button type="submit">Submit</button>
  </form>)
}```
results in `Invalid arguments found in variables. Did you mean to use ({ ambiguous: { ... } })`

infiton:
> model with ambiguous field

fixed this and added a test for it

packages/react/src/useActionForm.ts
infiton: adding a note in https://github.com/gadget-inc/gadget/pull/8598

jasong689:
😭 one more ambiguous issue when using `useActionForm(api.ambiguous.update, { findBy: "1"})`. my default value for the ambiguous field gets populated with `[object Object]`

Screen Shot 2023-10-17 at 4 55 30 PM

packages/react/src/useActionForm.ts
infiton: ok added the network errors to the root

jasong689 (approved):
looks like ambiguous models are fixed. my other comments/questions aren't blocking. i think we can work on a follow up pr to help address casting number fields automatically.

airhorns (approved):

PR #288, opened 2023-10-18
better typing for useActionForm

infiton:
once https://github.com/gadget-inc/js-clients/pull/286 we found that the typing in our email password template had issues; this was mainly because we didn't add the server side validation types to `formState: {errors}` this PR adds that typing (I'm sorry it's so confusing 😢 ) to say that the server side validation errors can be any of the field values of the model or any of the variable keys.

There is one issue in that for reset password for the typing to work we need to have:

```
useActionForm<
      (typeof fullAuthApi.user.resetPassword)["optionsType"],
      (typeof fullAuthApi.user.resetPassword)["schemaType"],
      typeof fullAuthApi.user.resetPassword,
      { confirmPassword?: string }
    >(fullAuthApi.user.resetPassword, {
      defaultValues: { code: "abc123" },
    });
```

I'm not sure how we can do this typing in the js file?

## PR Checklist

### DIFF
--- a/packages/react/src/useActionForm.ts
+++ b/packages/react/src/useActionForm.ts
@@ -106,13 +106,41 @@ type UseActionFormSubmit<F extends ActionFunction<any, any, any, any, any> | Glo
   event?: React.BaseSyntheticEvent<object, any, any> | undefined
 ) => Promise<UseActionFormHookState<F>>;
 
+type ExcludeUndefined<T> = T extends undefined ? never : T;
+
+type ServerSideError<F extends ActionFunction<any, any, any, any, any> | GlobalActionFunction<any>> = F extends ActionFunction<
+  any,
+  any,
+  any,
+  any,
+  any
+>
+  ? {
+      [key in F["modelApiIdentifier"]]?: {
+        [key in
+          | keyof F["selectionType"]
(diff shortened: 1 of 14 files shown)


packages/react/src/useActionForm.ts
```diff
@@ -106,13 +106,41 @@ type UseActionFormSubmit<F extends ActionFunction<any, any, any, any, any> | Glo
...
+};
+
 export type UseActionFormResult<
   GivenOptions extends OptionsType,
   SchemaT,
   ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any> | GlobalActionFunction<any>,
   FormVariables extends FieldValues,
   FormContext = any
-> = Omit<UseFormReturn<FormVariables, FormContext>, "handleSubmit"> & {
+> = Omit<Omit<UseFormReturn<FormVariables, FormContext>, "handleSubmit">, "formState"> & {
```
airhorns, line 142: Can you avoid the double omit with `Omit<xyz, "a" | "b">`

airhorns (approved):

packages/react/src/useActionForm.ts
infiton: :pray:

PR #293, opened 2023-10-23
Allow a custom navigate to be provided to the Provider

jasong689:
By default our components that control navigation `SignInOrRedirect`/`SignOutOrRedirect` use `window.location`. This causes window reloads and flashes of unstyled content. We can't make assumptions about an app's routing system, so we should allow consumers to pass in a custom navigate function that we'll use.

## PR Checklist


infiton:
what would the provider setup look like in this case?

infiton (approved):
great

PR #296, opened 2023-10-27
Fix updated spec params and add a comment for new useNavigate prop

jasong689:
I merged 2 PRs however one updated the signature of one of the test utils. This fixes that and adds a comment to the `useNavigate` prop. Also bump the version to `0.14.14`

## PR Checklist


infiton (approved):
small comment suggestion then 🚀

PR #297, opened 2023-10-27
update useSignout to also use navigate prop if possible

jasong689:
Missed `useSignout` when updating components to use the `navigate` prop

## PR Checklist

### DIFF
--- a/packages/react/src/auth/useSignOut.ts
+++ b/packages/react/src/auth/useSignOut.ts
@@ -1,7 +1,8 @@
-import { useCallback, useContext } from "react";
+import { useCallback, useContext, useEffect } from "react";
 import { GadgetConfigurationContext, useApi } from "../GadgetProvider.js";
 import { useAction } from "../useAction.js";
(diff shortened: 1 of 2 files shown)


packages/react/src/auth/useSignOut.ts
```diff
@@ -23,10 +24,14 @@ export const useSignOut = (opts?: { redirectOnSuccess?: boolean; redirectToPath?
 
     if (error) throw error;
 
-    if (redirectOnSuccess && data) {
-      const redirectUrl = new URL(redirectToPath ?? signInPath, window.location.origin);
-      window.location.assign(redirectUrl.toString());
-    }
+    // eslint-disable-next-line
+    useEffect(() => {
```
jasong689, line 28: This is needed because using something like `react-router-dom`s navigate means react re-rendering a different component while this one updates which results in
```Cannot update a component while rendering a different component warning```

infiton (approved):

PR #298, opened 2023-11-08
support flat fields in useActionForm when findBy is specifiec

infiton:
a user noticed (ref https://github.com/gadget-inc/gadget/pull/8834) that `useActionForm` didn't work when registering flat fields if they also loaded a record with a `findBy`; this is because we were flattening the findBy result into `defaultValues` AND when disambiguating fields if there is a value nested under the models apiIdentifier it chooses that over the flat ones.

To solve this when a find result comes back we set the default values at both the top level and the nested level; however this means that when the form is submitted we need to see for each field key if the submitted value is different from the default value; this is handled in `disambiguateDefaultValues`

## PR Checklist


packages/react/src/useActionForm.ts
```diff
@@ -67,6 +67,37 @@ const toDefaultValues = (modelApiIdentifier: string | undefined, data: any) => {
...
+  );
+
+  const modelData = { ...data[action.modelApiIdentifier] };
+
+  for (const key of Object.keys(modelData)) {
+    const initialValue = initialData[key];
+    const flatValue = data[key];
+    const nestedValue = data[action.modelApiIdentifier]?.[key];
+
+    if (nestedValue && initialValue !== nestedValue) {
```
jasong689, line 90: This might need to be a `nestedValue !== undefined && initialValue !== nestedValue` so that `0` is a valid value. Or maybe even a `key in data[action.modelApiIdentifier]`

jasong689 (commented):
I'm cool with this approach but another way we could do this is by wrapping register and inspecting which properties the consumer is registering to. This way we don't need to care about the values

infiton:
`wrapping register` that was the original approach I took, however you can still change the state of the form with `setValue`; we could also look at the dirty tracking, but again the user could change the data that is submitted without updating the dirty state

I'm not sure there is another way to know for sure what has changed without looking at what came in from the findBy

jasong689:
Hm yea, could we check the `formState`? It looks like there's a `touchedFields` property for anything that's been changed

packages/react/src/useActionForm.ts
infiton: :+1: added

packages/react/src/useActionForm.ts
```diff
@@ -67,6 +67,37 @@ const toDefaultValues = (modelApiIdentifier: string | undefined, data: any) => {
...
+  );
+
+  const modelData = { ...data[action.modelApiIdentifier] };
+
+  for (const key of Object.keys(modelData)) {
+    const initialValue = initialData[key];
+    const flatValue = data[key];
+    const nestedValue = data[action.modelApiIdentifier]?.[key];
+
+    if (key in data[action.modelApiIdentifier] && initialValue !== nestedValue) {
```
jasong689, line 90: Sorry I think there's one more check in that `data[action.modelApiIdentifier]` must be an object to use the `in` operator. I think `data` we know is an object but not `data[action.modelApiIdentifier]`

infiton: yup! good call

packages/react/src/useActionForm.ts
```diff
@@ -67,6 +67,39 @@ const toDefaultValues = (modelApiIdentifier: string | undefined, data: any) => {
...
+    })
+  );
+
+  const modelData = { ...data[action.modelApiIdentifier] };
+
+  for (const key of Object.keys(modelData)) {
+    const initialValue = initialData[key];
+
+    if (
+      typeof data[action.modelApiIdentifier] === "object" &&
```
jasong689, line 89: `typeof null` is `object`, thanks javascript. this is what lodash does
https://github.com/lodash/lodash/blob/3.0.2-npm-packages/lodash.isobject/index.js

infiton: :heart:

jasong689 (approved):

infiton:
@airhorns mind taking a look at this when you have a moment? Not urgent though

airhorns (commented):
This makes sense -- is complicated but if the default values don't get sent when we stash them in both places it will work! I think for anyone debugging looking at the form values it is pretty confusing though. What do you think about an option passed to `useActionForm` instead, like `paramsStyle: "flat" | "nested"` and we default to flat?

infiton:
> anyone debugging looking at the form values it is pretty confusing though

you mean because they will see default values at both the flat and nested level? I guess so, but I don't know if I find it more surprising than the fact that we set so many default values even if you haven't registered them with the form

> What do you think about an option passed to useActionForm instead, like paramsStyle: "flat" | "nested" and we default to flat

imo this is more likely to lead to confusion, not strongly opposed but not sure how devs will learn this option exists and know when to choose what

infiton:
going to merge this as @pistachiobaby is going to building on top of it

PR #302, opened 2023-12-05
add: rate limit error handling

MikePresman:
This change implements handling rate limit errors on the GadgetConnection clientside communicating to the core Gadget API over websockets.  The  server sends a rate limit code and closes the websocket over the wire.

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -24,6 +24,10 @@ import {
 } from "./support.js";
 
 export type TransactionRun<T> = (transaction: GadgetTransaction) => Promise<T>;
+
+export enum GadgetGraphQLCloseCode {
+  TooManyRequests = 4294,
+}
 export interface GadgetSubscriptionClientOptions extends Partial<SubscriptionClientOptions> {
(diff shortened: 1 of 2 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -496,6 +502,10 @@ export class GadgetConnection {
 
       const retryOnClose = (event: unknown) => {
         if (isCloseEvent(event)) {
+          if (event.code == 4429) {
+            return wrappedReject(new GadgetTooManyRequestsError(event.reason));
```
angelini, line 506: I think we want to do like below and reset listeners:

```
resetListeners()
return wrappedReject(new GadgetTooManyRequestsError(event.reason));
```

angelini (approved):

packages/api-client-core/src/GadgetConnection.ts
MikePresman: good point, makes more sense to just clearListeners.

thegedge (approved):
🙌 

I'm guessing you 🎩-ed this change in the Gadget codebase?

PR #304, opened 2023-12-11
pin react-hook-form to a version that supports node 16

infiton:
This releases a new version of `@gadgetinc/react` that locks `react-hook-form` down to a version that supports node 16. We will allow users on Gadget version `0.1.0` to continue to install `@gadgetinc/react`

## PR Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.14.15",
+  "version": "0.14.16",
   "files": [
     "README.md",
     "dist/**/*"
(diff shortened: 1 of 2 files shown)


packages/react/package.json
```diff
@@ -30,7 +30,7 @@
   "dependencies": {
     "@gadgetinc/api-client-core": "^0.15.11",
     "react-fast-compare": "^3.2.2",
-    "react-hook-form": "^7.46.1",
+    "react-hook-form": "~7.46.1",
```
jasong689, line 33: Is it worth pinning this to `~7.48.2` to get latest before the node `v18` change?

jasong689 (approved):

jasong689:
Looks like the lockfile also needs to be updated

packages/react/package.json
infiton: :+1:

PR #306, opened 2023-12-18
Add support to useActionForm to handle nested relationships

pistachiobaby:
One of the big drawbacks of `useActionForm` is that it didn't support nested relationships. Hopefully with this PR we can fix that.

We achieve this in a few steps:

When fetching data, unwind the edges and node to adhere to a simpler more usable structure for `react-hook-forms`

Afterwards on submit:

1. fetch the relationship(s) meta data that exists on the client
2. Create a map of updates that existed on the default values
3. Traverse through the data structure, transform anything to match our GraphQL api.

For 🎩:

Make sure you replace `gadget-react` in your package json with: `"@gadgetinc/react": "github:gadget-inc/js-clients#@gadgetinc/react-v0.15.0-gitpkg-b07ca84"`

Best thing to do for tophatting would be to try to build a mini app with complex nested relationships. `HasMany -> HasMany -> BelongsTo`, `HasOne -> HasOne -> HasMany` anything you can really think of.

Incase there's a need for a working example: https://zxcv-product-quiz-v5.gadget.app/

## PR Checklist

### DIFF
--- a/packages/react/src/useActionForm.ts
+++ b/packages/react/src/useActionForm.ts
@@ -1,12 +1,13 @@
 import type {
   ActionFunction,
+  AnyClient,
   AnyModelManager,
   DefaultSelection,
   GadgetRecord,
   GlobalActionFunction,
   Select,
 } from "@gadgetinc/api-client-core";
-import { camelize } from "@gadgetinc/api-client-core";
+import { $modelRelationships, camelize } from "@gadgetinc/api-client-core";
 import { useCallback, useEffect, useRef } from "react";
 import type { DeepPartial, FieldErrors, FieldValues, UseFormProps, UseFormReturn } from "react-hook-form";
 import { useForm } from "react-hook-form";
@@ -46,7 +47,7 @@ const useFindExistingRecord = (
   }
 };
 
-const OmittedKeys = ["__typename", "id", "createdAt", "updatedAt"] as const;
+const OmittedKeys = ["id", "createdAt", "updatedAt", "__typename"] as const;
 type OmittedKey = (typeof OmittedKeys)[number];
 
 const omitKeys = (data: any) => {
@@ -64,7 +65,29 @@ const toDefaultValues = (modelApiIdentifier: string | undefined, data: any) => {
     data[modelApiIdentifier] = omitKeys(data[modelApiIdentifier]);
   }
 
-  return data;
+  return unwindEdges(data);
+};
+
+const unwindEdges = (input: any): any => {
+  if (input !== null || (input !== undefined && typeof input === "object")) {
+    if (Array.isArray(input)) {
+      return input.map(unwindEdges);
+    }
+
+    if (input !== null && typeof input === "object") {
+      if (input.edges && Array.isArray(input.edges)) {
+        return input.edges.map((edge: any) => unwindEdges(edge.node));
+      }
+
+      const result: any = {};
+      for (const key of Object.keys(input)) {
+        result[key] = unwindEdges(input[key]);
+      }
+      return result;
+    }
+
+    return input;
+  }
 };
 
 const disambiguateDefaultValues = (data: any, initialData: any, action: any) => {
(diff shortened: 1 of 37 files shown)


packages/react/src/useActionForm.ts
```diff
@@ -336,14 +575,23 @@ export const useActionForm = <
...
         async (data) => {
+          if (isModelAction) {
+            try {
+              if (!action.hasAmbiguousIdentifier && findResult.data) {
+                data = disambiguateDefaultValues(data, findResult.data, action);
+              }
+
+              data = await reshapeDataForGraphqlApi(api, defaultValues, data);
+            } catch (e: any) {
+              handleSubmissionError(e);
```
pistachiobaby, line 586: Somewhat related to the code around here, I've caught my self a few times missing important errors that are produced by useActionForm since I didn't hook into the `onError` function.

If there's ever issues with the `reshapeDataForGraphqlApi` it'll propagate the errors to the `onError` (I kind of assumed it's reserved for errors related to the API) hook. Maybe this isn't the best exeperience for dev's and instead we should just `console.error`?

packages/react/src/useActionForm.ts
```diff
@@ -101,6 +128,198 @@ const disambiguateDefaultValues = (data: any, initialData: any, action: any) =>
   return result;
 };
 
+const reshapeDataForGraphqlApi = async (client: AnyClient, defaultValues: any, data: any) => {
```
pistachiobaby, line 131: I'm not married to `reshapreDataForGraphqlApi` it's just the name I settled on. If there's any alternatives I'm open to them.

packages/react/src/useActionForm.ts
```diff
@@ -241,8 +476,12 @@ export const useActionForm = <
      * Called when the form submission errors before sending, during the API call, or if the API call returns an error.
      */
     onError?: (error: Error | FieldErrors<ActionFunc["variablesType"]>) => void;
+    /**
```
infiton, line 479: still needed?

pistachiobaby: 🔥

packages/react/src/useActionForm.ts
```diff
@@ -64,7 +65,33 @@ const toDefaultValues = (modelApiIdentifier: string | undefined, data: any) => {
...
+};
+
+const unwindEdges = (input: any): any => {
+  if (Array.isArray(input)) {
+    return input.map(unwindEdges);
+  }
+
+  if (input !== null || (input !== undefined && typeof input === "object")) {
+    if (Array.isArray(input)) {
+      return input.map(unwindEdges);
```
infiton, line 78: duplicated from a few lines above?

packages/react/src/useActionForm.ts
```diff
@@ -101,6 +128,198 @@ const disambiguateDefaultValues = (data: any, initialData: any, action: any) =>
...
+
+            if (!item) {
+              // if the item is not found, we need to delete it from the updates object as well as anything that references it
+              const updateEntries = Object.entries(updates); // grab all the entries from the updates object
+              const updateEntry = updateEntries.find(([key, _ids]) => key.includes(path! + "." + nodeIndex));
+              if (updateEntry) {
+                // if we find an entry that matches the path, we need to delete it from the updates object
+                const { 0: updatePath, 1: _ } = updateEntry;
+                delete updates[updatePath];
+              } // TODO - What if we don't find an entry? Should we throw an error?
```
infiton, line 207: unresolved TODO?

packages/react/src/useActionForm.ts
infiton: why catch at all? this feels like something that should throw

packages/react/src/useActionForm.ts
```diff
@@ -101,6 +128,198 @@ const disambiguateDefaultValues = (data: any, initialData: any, action: any) =>
...
 };
 
+const reshapeDataForGraphqlApi = async (client: AnyClient, defaultValues: any, data: any) => {
+  const referencedTypes = client[$modelRelationships];
+
+  if (!referencedTypes) {
+    throw new Error("No referenced types found");
+  }
+
+  function getUpdates(data: any): Record<string, number[]> {
```
airhorns, line 138: nit: do these functions need to be defined inline in the parent function? if not, can we move them out ot the module scope? redefining them each invocation consumes more memory and slows it down

packages/react/src/useActionForm.ts
```diff
@@ -101,6 +128,198 @@ const disambiguateDefaultValues = (data: any, initialData: any, action: any) =>
   return result;
 };
 
+const reshapeDataForGraphqlApi = async (client: AnyClient, defaultValues: any, data: any) => {
+  const referencedTypes = client[$modelRelationships];
+
+  if (!referencedTypes) {
+    throw new Error("No referenced types found");
```
airhorns, line 135:
```suggestion
    throw new Error("No Gadget model metadata found -- please ensure you are using the latest version of the API client for your app");
```

packages/react/src/useActionForm.ts
```diff
@@ -101,6 +128,198 @@ const disambiguateDefaultValues = (data: any, initialData: any, action: any) =>
...
+      } else if (input !== undefined && input !== null && typeof input === "object") {
+        const result: any = {};
+
+        for (const key of Object.keys(input)) {
+          const currentPath = path ? `${path}.${key}` : key;
+          result[key] = traverse(input[key], currentPath, depth + 1);
+        }
+
+        if (depth > 1) {
+          const newPath = path?.substring(0, path.length - 2);
```
airhorns, line 156: What's the substring for?

airhorns (approved):
Can't find any major issues! Built a little app to test HasMany and it worked great! We're gonna need some serious docs on `useFieldArray`

packages/react/src/useActionForm.ts
pistachiobaby: It's to create a path name that excludes the index, since I just end up building an array of ids.

packages/react/src/useActionForm.ts
pistachiobaby: Yeah for sure, I inlined the functions since It felt like reshapeDataForGraphAPI is the only one that's gonna use it. I'm starting to wonder if maybe I should just create a "reshaper" class or something to encapsulate it all.

packages/react/src/useActionForm.ts
pistachiobaby: For example: `questions.0.answers.0` -> `questions.0.answers = [1]`

pistachiobaby: I realize the above might be a little confusing haha, in that case the `1` is the id in `questions.0.answers.0`

infiton (approved):
a lot of work went in to this 🏆 as far as I can follow this makes sense to me and I think your test coverage is very strong

PR #311, opened 2024-01-18
Update generated JS clients used in tests

airhorns:
- can now use the real implementations from a generated client for testing bulk actions
 - need to mock old clients missing the new model relationships metadata

## PR Checklist


airhorns:
Blessings @jasong689

PR #317, opened 2024-02-12
Use a lite version of @repeater/repeaterjs

maneetgoyal:
## PR Checklist

```

Test Suites: 44 passed, 44 total
Tests:       446 passed, 446 total
Snapshots:   204 passed, 204 total
Time:        6.371 s
```

Live Testing:

https://github.com/gadget-inc/js-clients/assets/25290212/13091fa7-7606-4039-8107-16ecf7e073a3


maneetgoyal:
@airhorns Can you please take a look?

packages/api-client-core/src/graphql-live-query-utils/index.ts
```diff
@@ -0,0 +1,5 @@
+import { createApplyLiveQueryPatch } from "./createApplyLivequeryPatch.js";
+import { applyJSONDiffPatch } from "@n1ru4l/graphql-live-query-patch-jsondiffpatch";
+
+export const applyLiveQueryJSONDiffPatch =
+    createApplyLiveQueryPatch(applyJSONDiffPatch);
```
airhorns, line 5: Can we import some of the tests from the original package and test the imported code here?

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
airhorns, line 1: Would you mind re-running with an up to date pnpm version so this doesn't regress?

packages/api-client-core/package.json
```diff
@@ -28,6 +28,7 @@
     "prerelease": "gitpkg publish"
   },
   "dependencies": {
+    "@manufac/repeater": "^4.0.6",
```
airhorns, line 31: Can you share the source for this package with us? I'm fine either way if you want to keep it under your github account or move it to ours, just want to make sure we can patch it further once the engagement is done

airhorns (commented):
This is the

airhorns:
I think if you rebase or merge main, you should get github actions running 👍

packages/api-client-core/package.json
maneetgoyal: Added you to the repo. Planning to add all that code into this same monorepo in a subsequent PRs. They are using a different version of TS which was causing some compilation errors with the TS config used here.

PR #318, opened 2024-02-13
don't throw if useActionForm doesn't have a fieldType but interpret t…

infiton:
useActionForm throws if it gets to a field like `user.roles` that is an array of objects because it assumes that this should represent a field of relationships; if it can't find a field type it throws. I don't think we should throw here (it's breaking our change password form on the web template)

@pistachiobaby is there a reason we must throw here?

## PR Checklist


pistachiobaby:
@infiton You're right, that was probably me being a little overcautious with the field type and assuming that every thing should have a relationship.

pistachiobaby (approved):

PR #319, opened 2024-02-15
Adopt new property names for background job options

airhorns:


packages/api-client-core/src/types.ts
```diff
@@ -811,45 +835,22 @@ export type EnqueueBackgroundActionOptions<Action extends AnyActionFunction> = {
...
+   * Setting `retries` to an object allows configuring the count as well as the schedule retries will be attempted on, @see BackgroundActionRetryPolicy
    *
    * @example
    * // retry up to 3 more times after failure (4 attempts total)
-   * retry: 3
+   * retries: 3
    *
    * @example
-   * retry: { retries: 5, minTimeout: 1000, maxTimeout: 60000, randomize: true }
+   * retries: { retries: 5, minTimeout: 1000, maxTimeout: 60000, randomize: true }
```
thegedge, line 849: `retries` → `retryCount` (similarly for the rest).

thegedge (approved):

PR #321, opened 2024-02-27
Shopify app bridge v4 support

jasong689:
This updates `react-shopify-app-bridge` to support `@shopify/app-bridge-react` v4. Will run this through CI over in gadget before merging

## PR Checklist

### DIFF
--- a/packages/react-shopify-app-bridge/spec/Provider.spec.tsx
+++ b/packages/react-shopify-app-bridge/spec/Provider.spec.tsx
@@ -1,7 +1,6 @@
 import type { AnyClient } from "@gadgetinc/api-client-core";
 import { GadgetConnection } from "@gadgetinc/api-client-core";
-import { AppBridgeContext } from "@shopify/app-bridge-react/context.js";
-import { Redirect } from "@shopify/app-bridge/actions/index.js";
+import * as AppBridgeReact from "@shopify/app-bridge-react";
 import "@testing-library/jest-dom";
 import { render } from "@testing-library/react";
 import React from "react";
@@ -12,7 +11,9 @@ describe("GadgetProvider", () => {
   let mockApiClient: AnyClient;
   const { location } = window;
   const mockNavigate = jest.fn();
+  const mockOpen = jest.fn();
   const mockApiKey = "some-api-key";
+  let useAppBridgeMock: jest.SpyInstance;
 
   describe.each([true, false])("as install request: %s", (isInstallRequest) => {
     beforeAll(() => {
@@ -27,15 +28,17 @@ describe("GadgetProvider", () => {
         search: isInstallRequest ? "?shop=example.myshopify.com&hmac=abcdefg&host=abcdfg" : "",
       };
 
-      const originalUseContext = React.useContext;
+      window.open = mockOpen;
 
-      jest.spyOn(React, "useContext").mockImplementation((context) => {
-        if (context === AppBridgeContext) {
-          return {};
-        } else {
-          return originalUseContext(context);
-        }
-      });
+      window.shopify = {
+        // @ts-expect-error mock
+        environment: {
+          embedded: false,
+          mobile: false,
+        },
+      };
+
+      useAppBridgeMock = jest.spyOn(AppBridgeReact, "useAppBridge").mockImplementation(() => window.shopify);
     });
 
     beforeEach(() => {
@@ -50,6 +53,7 @@ describe("GadgetProvider", () => {
 
     afterEach(() => {
       mockNavigate.mockClear();
+      useAppBridgeMock.mockClear();
     });
 
     afterAll(() => {
@@ -93,19 +97,19 @@ describe("GadgetProvider", () => {
       });
 
       if (isInstallRequest) {
-        expect(mockNavigate).toHaveBeenCalledWith(
-          "https://test-app.gadget.app/api/connections/auth/shopify?shop=example.myshopify.com&hmac=abcdefg&host=abcdfg"
+        expect(mockOpen).toHaveBeenCalledWith(
+          "https://test-app.gadget.app/api/connections/auth/shopify?shop=example.myshopify.com&hmac=abcdefg&host=abcdfg",
+          "_top"
         );
+        expect(mockNavigate).not.toHaveBeenCalled();
       } else {
         expect(mockNavigate).not.toHaveBeenCalled();
+        expect(mockOpen).not.toHaveBeenCalled();
         expect(container.outerHTML).toMatchInlineSnapshot(`"<div><span>hello world</span></div>"`);
       }
     });
 
     test("can render a standalone app type in embedded context", () => {
-      // @ts-expect-error mock
-      jest.spyOn(window, "self", "get").mockImplementation(() => ({}));
-
       const { container } = render(
         <Provider api={mockApiClient} shopifyApiKey={mockApiKey} type={AppType.Standalone}>
           <span>hello world</span>
@@ -125,9 +129,7 @@ describe("GadgetProvider", () => {
     });
 
     test("can render a standalone app type in mobile context", () => {
-      window.MobileWebView = {
-        postMessage: jest.fn(),
-      };
+      window.shopify.environment.mobile = true;
 
       const { container } = render(
         <Provider api={mockApiClient} shopifyApiKey={mockApiKey} type={AppType.Standalone}>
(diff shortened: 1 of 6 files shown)


packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -1,14 +1,13 @@
...
-import { getSessionToken } from "@shopify/app-bridge-utils";
-import { Redirect } from "@shopify/app-bridge/actions/index.js";
-import React, { memo, useContext, useEffect, useMemo, useState } from "react";
+import { useAppBridge } from "@shopify/app-bridge-react";
+import React, { memo, useEffect, useMemo, useState } from "react";
 import type { GadgetAuthContextValue } from "./index.js";
 import { GadgetAuthContext } from "./index.js";
 
+// if we haven't properly set up the shopify global then skip anything that requires app bridge
+const shopifyGlobalDefined = !!(globalThis && globalThis.shopify);
```
jasong689, line 9: we don't inject the shopify script tag if you don't have an api key or `@shopify/app-bridge-react` v4 in gadget. this check avoids trying to use the `useAppBridge` hook which throws whenever the shopify global hasn't been set up properly

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -17,14 +16,12 @@ export enum AppType {
...
   children: JSX.Element | JSX.Element[];
-  forceRedirect: boolean;
-  isEmbedded: boolean;
   gadgetAppUrl: string;
   originalQueryParams?: URLSearchParams;
   api: AnyClient;
-  isRootFrameRequest: boolean;
-  isInstallRequest: boolean;
   type: AppType;
+  host: string | undefined;
```
infiton, line 23: this change in props  will mean upgrade steps existing apps right?

jasong689: sorry these are just the props for the `InternalGadgetProvider` not the publicly accessible ones. although i did remove the `router` prop on the `GadgetProvider` since the recommendation is now to simply use `history`

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -219,51 +186,39 @@ export const Provider = ({
...
-      >
-        {app}
-      </AppBridgeReact.Provider>
-    );
-  }
-
-  return app;
+  return (
+    <GadgetUrqlProvider api={api}>
+      {shopifyGlobalDefined ? (
```
infiton, line 197: will this handle the case where the app type is standalone? I think we need to check that here

jasong689: ah yea, standalone app check should be added here too

jasong689: i updated this to render the plain children if you have a standalone app. it seems like the old way was to try and redirect you to gadget which would then send you to shopify and then you'd end up in a redirect loop which seemed wrong

airhorns (commented):
I feel like its really hard to know if we've broken standalone support with this change, its intense and I know the way the existing code works is really sensitive. I think we need to add at least a couple smoke tests that prove the thing works roughly the way it did before, or that we understand how we're changing the behaviour well. I think its long overdue for this package but I feel like I can't really tell you if this is right or not without tests.

packages/react-shopify-app-bridge/spec/Provider.spec.tsx
```diff
@@ -103,9 +104,6 @@ describe("GadgetProvider", () => {
     });
 
     test("can render a standalone app type in embedded context", () => {
-      // @ts-expect-error mock
-      jest.spyOn(window, "self", "get").mockImplementation(() => ({}));
```
jasong689, line 107: instead of relying on `window.self === window.top` app bridge can now tell us if we're embedded

packages/react-shopify-app-bridge/spec/Provider.spec.tsx
```diff
@@ -27,15 +27,15 @@ describe("GadgetProvider", () => {
...
       };
 
-      const originalUseContext = React.useContext;
+      window.shopify = {
+        // @ts-expect-error mock
+        environment: {
+          embedded: false,
+          mobile: false,
+        },
+      };
```
airhorns, line 36: This is way less terrible than i thought, nice!

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -161,27 +147,66 @@ const InnerGadgetProvider = memo(
...
+        canAuth: false,
+        loading: false,
+        appBridge: null,
+        isRootFrameRequest: isRootFrameRequest,
+      }}
+    >
+      {children}
+    </GadgetAuthContext.Provider>
+  );
+};
```
airhorns, line 200: Extracting this was a great call 👍 We could potentially make it a different top level export so that you use a different provider from our package when embedded vs standalone, and then it would treeshake out a bit nicer. Don't have to make this change rn but something to aim for long term methinks

packages/react-shopify-app-bridge/spec/Provider.spec.tsx
```diff
@@ -218,9 +196,10 @@ describe("GadgetProvider", () => {
         hasNext: false,
       });
 
-      expect(mockDispatch).not.toHaveBeenCalled();
       expect(container.outerHTML).toMatchInlineSnapshot(`"<div><span>hello world</span></div>"`);
       expect(mockNavigate).not.toHaveBeenCalled();
+
+      window.shopify.environment.embedded = false;
     });
```
airhorns, line 203: Do you want to add any tests / error messages around the shopify global being absent?

airhorns (approved):
LGTM after comments

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -132,12 +121,9 @@ const InnerGadgetProvider = memo(
         gadgetAppUrl,
       });
 
-      if (isEmbedded && appBridge) {
-        Redirect.create(appBridge).dispatch(Redirect.Action.REMOTE, redirectURLWithOAuthParams);
-      } else {
-        window.location.assign(redirectURLWithOAuthParams);
-      }
-    }, [appBridge, gadgetAppUrl, isEmbedded, isRootFrameRequest, originalQueryParams, runningShopifyAuth]);
+      open(redirectURLWithOAuthParams, "_top");
```
jasong689, line 124: Changed this to ensure it's a top level redirect when embedded and it's an embedded app type

packages/react-shopify-app-bridge/spec/Provider.spec.tsx
jasong689: will take a look at this

jasong689:
going to run this through ci over in gadget

packages/react-shopify-app-bridge/spec/Provider.spec.tsx
infiton: woooo!

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -220,51 +228,31 @@ export const Provider = ({
...
-      >
-        {app}
-      </AppBridgeReact.Provider>
-    );
-  }
-
-  return app;
+  return (
+    <GadgetUrqlProvider api={api}>
+      {shopifyGlobalDefined && coalescedType === AppType.Embedded ? (
```
infiton, line 239: yah I think if app type is Embedded and there is no shopify Global we should fail loudly

jasong689: should we actually throw in those cases? i think it might make sense to set something on the `error` property of the `useGadget` hook. if we throw consumers will have to use an `ErrorBoundary` to handle errors

airhorns: i think this is the kind of error that only the developer can fix -- it isn't a result of the user's actions where their wifi is disconnected or whatever, it's busted code, so I think full throwing is fine?

infiton: what would the experience be right now if you tried to load this in the admin and for some reason the app bridge script tag wasn't there (I'm thinking if someone upgrades the gadget package but doesn't get the script tag)

As long as they can get the message about what they need to do that's all I'm worried about

infiton (approved):

PR #322, opened 2024-02-27
Add an runner to get action results

garymardell:
Uses a subscription endpoint on gadget (https://github.com/gadget-inc/gadget/pull/10364) to get the background action result. 

We are waiting on the subscription for the first pushed result that either has an outcome set or contains an error (for example permission error) and processes the responses in the same way as the "api" version of the actions. 

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/operationRunners.ts
+++ b/packages/api-client-core/src/operationRunners.ts
@@ -293,7 +293,7 @@ export const actionResultRunner = async <Action extends AnyActionFunction, Optio
   // If we have an outcome we should process the result accordingly
   if (backgroundAction.outcome) {
     // Check for errors in the action result and throw
-    assertResponseSuccess(backgroundAction);
+    assertResponseSuccess(backgroundAction.result);
 
     switch (action.type) {
       case "action": {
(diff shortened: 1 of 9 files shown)


packages/api-client-core/src/operationRunners.ts
```diff
@@ -257,11 +271,50 @@ export const enqueueActionRunner = async <Action extends AnyActionFunction>(
...
+): Promise<BackgroundActionResult> => {
+  const plan = actionResultOperation(id, action, options);
+  const subscription = connection.currentClient.subscription(plan.query, plan.variables);
+
+  const response = await pipe(
+    subscription,
+    filter((operation) => operation.error || operation.data?.backgroundAction?.outcome),
+    take(1),
+    toPromise
+  );
```
airhorns, line 297: hnoice

airhorns (approved):
No notes! Nice!

airhorns:
I am down to merge this and test it all out together, but one other thing we should test is reconnection. I think urql will just handle it for us, but if you start this subscription, and then your websocket dies, it should get re-established and the subscription should re-start automatically I think?

PR #324, opened 2024-02-29
Use liter version of repeaterjs

maneetgoyal:
Used a liter version of https://github.com/repeaterjs/repeater.

---

`pnpm i` is failing because of:

```
 WARN  Package name mismatch found while reading {"integrity":"sha1-AmNJ4SOUtpCatG/+uQxrx+xTkKg=","tarball":"https://registry.gadget.dev/npm/_/tarball/1268/1362/6391"} from the store. This means that the lockfile is broken. Expected package: @gadget-client/related-products-example@1.860.0-development.859. Actual package in the store by the given integrity: @gadget-client/related-products-example@1.859.0.
 ERR_PNPM_UNEXPECTED_PKG_CONTENT_IN_STORE  The lockfile is broken! A full installation will be performed in an attempt to fix it.
/Users/maneet/Documents/js-clients-1/packages/blog-example:
 ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/@gadget-client%2Fblog: Not Found - 404

This error happened while installing a direct dependency of /Users/maneet/Documents/js-clients-1/packages/blog-example

@gadget-client/blog is not in the npm registry, or you have no permission to fetch it.
```

---

## PR Checklist


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
maneetgoyal, line 1: @airhorns I tried downgrapding `pnpm` but this kept happening. Which `pnpm` version are you on?

maneetgoyal: I'm on `8.15.4`.

packages/api-client-core/package.json
```diff
@@ -28,8 +28,9 @@
     "prerelease": "gitpkg publish"
   },
   "dependencies": {
+    "@manufac/repeater": "^4.0.6",
```
maneetgoyal, line 31: @airhorns You have been added to this repo. I'll try to move all that code to this monorepo itself via subsequent PRs.

pnpm-lock.yaml
airhorns: Weird. I updated `main` to lock us all to use the same version of pnpm via corepack. If you rebase on top of `main` and then re-run pnpm, I'll take whatever lockfile that produces!

airhorns:
Ah, that PNPM store issue happened to me locally and is a result of a gadget bug. To fix it, run this:

```
rm -rf node_modules
pnpm store prune
pnpm install
```

and you should be back in business, it isn't actually the lock file that is busted but a bug in what version of a package tarball gadget served in the past.

PR #326, opened 2024-03-01
Use liter version of `@repeaterjs/repeater`

maneetgoyal:

## PR Checklist


maneetgoyal:
@airhorns WDYT?

airhorns:
This looks good, but as I said in the other PR, I think we need some enhanced tests for this code that we now own. If someone goes to modify this, how will they know if they broke it or not? Can you import any existing tests from the original package, and add some more?

maneetgoyal:
> This looks good, but as I said in the other PR, I think we need some enhanced tests for this code that we now own. If someone goes to modify this, how will they know if they broke it or not? Can you import any existing tests from the original package, and add some more?

Makes sense, adding tests soon.

PR #327, opened 2024-03-04
resolves issues with useActionForm and JSON fields

infiton:
This PR resolves a customer issue where they are trying to use `useActionForm` to update a JSON array field and it is treating it like a `HasMany` field leading to strange results. We shouldn't assume when input data is coming into the useAction payload that just because its an array we should treat it as a has many field. The fix is simple: check that the field type is in fact a hasMany

## PR Checklist


airhorns (approved):
LGTM!

PR #332, opened 2024-03-12
Fix result processing

garymardell:
Caught a case that I had broken with https://github.com/gadget-inc/js-clients/pull/322 where the result was not properly being accessed. Unfortunately this case where `hasReturnType` being true had no tests so was not caught during refactoring. 

I don't know if its possible to re-release the version; whether this happens automatically or I need to bump the version again.

## PR Checklist

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.15.17",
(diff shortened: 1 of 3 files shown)


airhorns:
Ah, this makes sense! No, once the version is released it can't be modified, so this'll have to be a new version! There's a way to use gitpkg to vendor a prerelease in Gadget if you want?

PR #333, opened 2024-03-18
Support startAt parameter to BackgroundActionOptions

garymardell:
Adds support for `startAt` option to enqueuing background actions to allow them to be scheduled into the future. This is already supported by the graphql api inside gadget. For convenience I am allowing both ISO strings and `Date` objects that will automatically be converted. 

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/types.ts
+++ b/packages/api-client-core/src/types.ts
@@ -1,4 +1,4 @@
-import { OperationContext } from "@urql/core";
+import type { OperationContext } from "@urql/core";
 import type { VariableOptions } from "tiny-graphql-query-compiler";
 import type { FieldSelection } from "./FieldSelection.js";
 import type { ActionFunction, AnyActionFunction, BulkActionFunction, GlobalActionFunction } from "./GadgetFunctions.js";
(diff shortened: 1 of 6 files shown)


packages/api-client-core/src/types.ts
```diff
@@ -858,6 +858,18 @@ export type EnqueueBackgroundActionOptions<Action extends AnyActionFunction> = {
...
+
+  /**
+   * Schedules the background action to run in the future after this datetime.
+   *
+   * @example
+   * startAt: "2024-03-18T18:14:08.257Z"
+   *
+   * @example
+   * // Start in 60 seconds from now
+   * startAt: new Date(new Date().getTime() + 60 * 1000)
```
thegedge, line 870: Dontcha just love doing date/time math in JS 🙈

thegedge (approved):
🚀 

From what I can tell, https://github.com/gadget-inc/js-clients/blob/main/packages/react/src/useEnqueue.ts should just work with these changes.

> For convenience I am allowing both ISO strings and Date objects that will automatically be converted.

🙇 I think this is critical, so thanks for adding!

garymardell:
> From what I can tell, https://github.com/gadget-inc/js-clients/blob/main/packages/react/src/useEnqueue.ts should just work with these changes.

Yes and I have added a test case that passes through the option in `useEnqueue` to confirm.

PR #335, opened 2024-03-20
Test produced packages with the are-the-types-wrong type export verifier package

airhorns:
This was a freaking journey, but this fixes all the packages in js-clients to pass the `arethetypeswrong` package publish lint.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "preinstall": "npx only-allow pnpm"
   },
   "devDependencies": {
+    "@arethetypeswrong/cli": "^0.15.1",
     "@gadget-client/app-with-no-user-model": "^1.8.0",
     "@gadget-client/bulk-actions-test": "^1.110.0",
     "@gadget-client/full-auth": "^1.7.0",
@@ -37,6 +38,7 @@
     "@types/node": "^16.11.7",
     "@types/semver": "^7.5.8",
     "eslint": "^8.57.0",
+    "execa": "^5.1.1",
     "gitpkg": "^1.0.0-beta.4",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
--- a/package.json
+++ b/package.json
@@ -22,12 +22,12 @@
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.15.1",
-    "@gadget-client/app-with-no-user-model": "^1.8.0",
-    "@gadget-client/bulk-actions-test": "^1.110.0",
-    "@gadget-client/full-auth": "^1.7.0",
-    "@gadget-client/related-products-example": "^1.862.0",
-    "@gadget-client/zxcv-deeply-nested": "^1.210.0",
-    "@gadget-client/zxcv-simple-relationship": "^1.20.0",
+    "@gadget-client/app-with-no-user-model": "^1.10.0",
+    "@gadget-client/bulk-actions-test": "^1.113.0",
+    "@gadget-client/full-auth": "^1.9.0",
+    "@gadget-client/related-products-example": "^1.865.0",
+    "@gadget-client/zxcv-deeply-nested": "^1.212.0",
+    "@gadget-client/zxcv-simple-relationship": "^1.23.0",
     "@gadgetinc/api-client-core": "workspace:*",
     "@gadgetinc/eslint-config": "^0.6.1",
     "@gadgetinc/prettier-config": "^0.4.0",
@@ -55,6 +55,9 @@
     "overrides": {
       "@gadgetinc/api-client-core": "workspace:*",
       "@gadgetinc/react": "workspace:*"
+    },
+    "patchedDependencies": {
+      "@n1ru4l/push-pull-async-iterable-iterator@3.2.0": "patches/@n1ru4l__push-pull-async-iterable-iterator@3.2.0.patch"
     }
   },
   "packageManager": "pnpm@8.15.4+sha256.cea6d0bdf2de3a0549582da3983c70c92ffc577ff4410cbf190817ddc35137c2"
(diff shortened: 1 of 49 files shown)


package.json
```diff
@@ -21,12 +21,13 @@
...
-    "@gadget-client/related-products-example": "^1.862.0",
-    "@gadget-client/zxcv-deeply-nested": "^1.210.0",
-    "@gadget-client/zxcv-simple-relationship": "^1.20.0",
+    "@arethetypeswrong/cli": "^0.15.1",
+    "@gadget-client/app-with-no-user-model": "^1.10.0",
+    "@gadget-client/bulk-actions-test": "^1.113.0",
+    "@gadget-client/full-auth": "^1.9.0",
+    "@gadget-client/related-products-example": "^1.865.0",
+    "@gadget-client/zxcv-deeply-nested": "^1.212.0",
+    "@gadget-client/zxcv-simple-relationship": "^1.23.0",
```
airhorns, line 30: Had to get all these clients working first so there were no types mismatches in the tests here 👍

package.json
```diff
@@ -53,6 +55,9 @@
     "overrides": {
       "@gadgetinc/api-client-core": "workspace:*",
       "@gadgetinc/react": "workspace:*"
+    },
+    "patchedDependencies": {
+      "@n1ru4l/push-pull-async-iterable-iterator@3.2.0": "patches/@n1ru4l__push-pull-async-iterable-iterator@3.2.0.patch"
```
airhorns, line 60: Fun fact: the types are wrong in this package and it causes issues in ours in attw. This fixes it locally so our tests pass.

packages/react/spec/auth/useSession.spec.ts
```diff
@@ -44,7 +45,6 @@ describe("useSession", () => {
             firstName
             googleImageUrl
             googleProfileId
-            id
```
airhorns, line 47: New versions of the clients sort these a bit differently, but its still deterministic.

packages/api-client-core/tsconfig.cjs.json
```diff
@@ -2,7 +2,8 @@
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./dist/cjs",
-    "module": "CommonJS"
+    "module": "CommonJS",
+    "moduleResolution": "node"
```
airhorns, line 6: attw insists that module and moduleResolution match so that you aren't lying to yourself about how node will actually do things -- all these updates make that the case, which required fixing some of the busted import statements.

packages/react/spec/__recordings__/useActionFormNested_1974165259/with-polly_2612993081/can-create-a-single-HasMany-HasMany-HasOne-BelongsTo-relationship-with-support-for-Id-pre_1709343723/recording.har
```diff
@@ -8,7 +8,7 @@
     },
     "entries": [
       {
-        "_id": "31f099dad6e92eaa85f9b95594fbd941",
+        "_id": "93610691231b267f112e91822785676b",
```
airhorns, line 11: New client version caused changes in the gql queries which caused all these to need to be re-recorded

airhorns:
ping sweet dear @thegedge sorry about the high volume of reviews, we want to get a release of the react shopify app bridge package out today

packages/api-client-core/spec/GadgetConnection-suite.ts
```diff
@@ -1,9 +1,10 @@
 import { Response } from "cross-fetch";
-import gql from "gql-tag";
+import gqlTag from "gql-tag";
 import nock from "nock";
 import { AuthenticationMode, BrowserSessionStorageType, GadgetConnection } from "../src/index.js";
 import { base64 } from "./helpers.js";
 
+const gql = gqlTag as any;
```
thegedge, line 7: nit: Maybe we could type it as `(literals: TemplateStringsArray, ...values: any[]) => string`?

packages/api-client-core/src/exchanges/urlParamExchange.ts
```diff
@@ -1,6 +1,7 @@
+import type { Exchange } from "@urql/core";
 import { mapExchange } from "@urql/core";
 
-export const urlParamExchange = mapExchange({
+export const urlParamExchange: Exchange = mapExchange({
```
thegedge, line 4: Curious: why this one? I would have assumed `mapExchange` would be typed to return an `Exchange`.

packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -15,7 +15,7 @@ import {
   hydrateRecord,
   hydrateRecordArray,
 } from "./support.js";
-import type { InternalFieldSelection, InternalFindListOptions, InternalFindManyOptions, InternalFindOneOptions } from "./types";
+import type { InternalFieldSelection, InternalFindListOptions, InternalFindManyOptions, InternalFindOneOptions } from "./types.js";
```
thegedge, line 18: Feels odd. Curious as to why this one was necessary.

thegedge (approved):
Great to have this in place here! Looks like a bunch of unfortunate things you had to do 🙈 🚀

packages/api-client-core/src/exchanges/urlParamExchange.ts
airhorns: It was for a portable type thing -- typescript asked to put an explicit annotation to make it portable

packages/api-client-core/src/InternalModelManager.ts
airhorns: All esm files gotta have the extension!

thegedge: I guess the thing I find surprising is that `tsc` wouldn't deal with this, eh? Just given it knows the target.

PR #336, opened 2024-03-21
Allow pagination parameters to be passed through select option

garymardell:
A problem faced is with pagination of `has many` relationships. The root nodes are paginated using a cursor based paginator where as the nested `has many` relationships are paginated with an offset paginator. This means passing the cursor from a has many connection back to a root query will raise an error. 

## Example

We want to paginate through all the comments for a given post. 

```graphql
query PostWithComments($id: GadgetID!) {
  post(id: $id) {
    id
    comments {
      pageInfo {
        endCursor
        hasNextPage
      }
      edges {
        node {
          id
        }
      }
    }
  }
}
```

Passing the `endCursor` to a `findMany` will fail. Instead we need to get the next page of comments for this given post. 

```graphql
query PostWithComments($id: GadgetID!) {
  post(id: $id) {
    id
    comments(after: "cursor") {
      pageInfo {
        endCursor
        hasNextPage
      }
      edges {
        node {
          id
        }
      }
    }
  }
}
```

## Solution

Technically this is already possible to do today with the API however you need to use the `tiny-query-graph-compiler` directly with the `Call`. This change introduces some specific keys that if present are automatically extracted into the `Call`.  

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/FieldSelection.ts
+++ b/packages/api-client-core/src/FieldSelection.ts
@@ -22,7 +22,7 @@ export interface EdgeFieldSelection {
 }
 
 function isEdgeConnection(selection: FieldSelection | ConnectionFieldSelection): boolean {
-  return ("edges" in selection && typeof selection["edges"] == "object");
+  return "edges" in selection && typeof selection["edges"] == "object";
 }
 
 export function isFieldSelection(selection: FieldSelectionType): selection is FieldSelection {
--- a/packages/api-client-core/src/FieldSelection.ts
+++ b/packages/api-client-core/src/FieldSelection.ts
@@ -3,5 +3,28 @@
  * Example: `{ id: true, name: false, richText: { markdown: true, html: false } }`
  **/
 export interface FieldSelection {
-  [key: string]: boolean | null | undefined | FieldSelection;
+  [key: string]: FieldSelectionType;
+}
+
+export type FieldSelectionType = boolean | null | undefined | FieldSelection | ConnectionFieldSelection;
+
+export interface ConnectionFieldSelection {
+  before?: string;
+  after?: string;
+  first?: number;
+  last?: number;
+  pageInfo?: FieldSelection;
+  edges: EdgeFieldSelection;
+}
+
+export interface EdgeFieldSelection {
+  node: FieldSelection;
+}
+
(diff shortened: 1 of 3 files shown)


packages/api-client-core/src/FieldSelection.ts
```diff
@@ -3,5 +3,28 @@
  * Example: `{ id: true, name: false, richText: { markdown: true, html: false } }`
  **/
 export interface FieldSelection {
-  [key: string]: boolean | null | undefined | FieldSelection;
+  [key: string]: FieldSelectionType;
```
garymardell, line 6: In order to support `after` and `first` parameters we need to allow the user to pass through both strings and numbers values. Adding them directly to this type means that all fields would accept them which is not ideal. In order to mitigate this I have introduced a specific `ConnectionFieldSelection` that has the just pagination parameters and edges.

packages/api-client-core/src/FieldSelection.ts
```diff
@@ -3,5 +3,28 @@
...
+  pageInfo?: FieldSelection;
+  edges: EdgeFieldSelection;
+}
+
+export interface EdgeFieldSelection {
+  node: FieldSelection;
+}
+
+export function isFieldSelection(selection: FieldSelectionType): selection is FieldSelection {
+  return !!(selection && typeof selection == "object" && !("edges" in selection));
```
angelini, line 25: Will this cause us to break `compileSelect` if they add a model field called `edges`?

angelini: Which we already have 2 models in prod which would have hit this case.

garymardell: Good catch, this broke a specific case when a `BelongsTo` has a node called edges. I have added additional check that also tests that if there is a field called `edges` in the selection that it is also an object (indicating that it is selecting a connection). I have also added a bunch of additional tests for each of the cases.

angelini (approved):
Awesome :+1:

airhorns:
I think this approach needs a bit more thinking, it means that you can't have a field named `after` when you have a pagination argument named `after` also. I think we need a different spot for the args to go.

garymardell:
Hmm I think almost all routes here with it being part of `select` lead to issues with keys being excluded or the types being incredibly awkward.

PR #340, opened 2024-04-03
Remove Repeater dep

maneetgoyal:
The repeater-lite code is moved to js-clients monorepo (api-client-core package) itself for easier maintenance.

Test cases passing:

image

Test Video with a prerelease version:

https://github.com/gadget-inc/js-clients/assets/25290212/ea6cd6bb-d0a0-4010-8d73-e148dd0075a3

## PR Checklist

### DIFF
--- a/packages/api-client-core/spec/repeater.spec.ts
+++ b/packages/api-client-core/spec/repeater.spec.ts
@@ -1,3 +1,7 @@
+/**
+ * Tests adopted from https://github.com/repeaterjs/repeater/blob/master/packages/repeater/src/__tests__/repeater.ts.
+ */
+
 import {
     Repeater,
     RepeaterOverflowError,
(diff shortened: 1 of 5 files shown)


maneetgoyal:
cc @airhorns

packages/api-client-core/spec/repeater.spec.ts
```diff
@@ -0,0 +1,1626 @@
...
+    RepeaterOverflowError,
+    DroppingBuffer,
+    FixedBuffer,
+    MAX_QUEUE_LENGTH,
+    SlidingBuffer,
+} from "../src/repeater/index.js";
+import { delayPromise } from "./helpers.js";
+
+
+describe("RepeaterBuffer", () => {
```
airhorns, line 12: Could you add a little comment here and in the source itself explaining where this code all came from just so the next person who reads it knows that there's an upstream this is copied from?

airhorns (approved):
LGTM after comments

packages/api-client-core/spec/repeater.spec.ts
maneetgoyal: @airhorns Done

PR #342, opened 2024-04-07
Add support for enqueuing actions in the background in bulk

airhorns:
This updates our operation planners and runners to work with background enqueued bulk actions, which return an array of handles instead of just one.

### DIFF
--- a/packages/api-client-core/src/GadgetFunctions.ts
+++ b/packages/api-client-core/src/GadgetFunctions.ts
@@ -108,11 +108,12 @@ export interface ActionFunctionMetadata<OptionsT, VariablesT, SelectionT, Schema
   variablesType: VariablesT;
   isBulk: IsBulk;
   hasAmbiguousIdentifier?: boolean;
-  /** @deprecated */
-  hasCreateOrUpdateEffect?: boolean;
   acceptsModelInput?: boolean;
   paramOnlyVariables?: readonly string[];
   hasReturnType?: boolean;
+  singleActionFunctionName?: string;
+  /** @deprecated */
+  hasCreateOrUpdateEffect?: boolean;
 }
 
 export type ActionFunction<OptionsT, VariablesT, SelectionT, SchemaT, DefaultsT> = ActionFunctionMetadata<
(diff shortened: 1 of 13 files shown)


packages/api-client-core/src/operationBuilders.ts
```diff
@@ -175,31 +167,32 @@ export const actionResultOperation = <Action extends AnyActionFunction, Options
...
-        [`... on ${camelize(action.operationName)}Result`]: globalActionFieldSelection(),
+        [`... on ${camelize(operationName)}Result`]: globalActionFieldSelection(),
       };
     }
   }
 
   const actionResultOperation: BuilderOperation = {
     type: "subscription",
-    name: action.operationName,
+    name: capitalizeIdentifier(operationName) + "BackgroundResult",
```
thegedge, line 195: nit: This is in a function called `actionResultOperation`, but this change implies it's a background action. Given other stuff in here, does seems like this change is the right one, so perhaps we rename the function to be the more explicit `backgroundActionResultOperation`?

packages/api-client-core/src/operationRunners.ts
```diff
@@ -307,7 +322,7 @@ export const actionResultRunner = async <Action extends AnyActionFunction, Optio
         action.defaultSelection,
         response.data,
         backgroundAction.result,
-        action.modelSelectionField,
+        action.isBulk ? action.modelApiIdentifier : action.modelSelectionField,
```
thegedge, line 325: Unfortunate discrepancy, but makes sense since backgrounding a bulk action turns it into a set of individual actions.

packages/api-client-core/src/GadgetFunctions.ts
```diff
@@ -108,11 +108,12 @@ export interface ActionFunctionMetadata<OptionsT, VariablesT, SelectionT, Schema
   variablesType: VariablesT;
   isBulk: IsBulk;
   hasAmbiguousIdentifier?: boolean;
-  /** @deprecated */
-  hasCreateOrUpdateEffect?: boolean;
   acceptsModelInput?: boolean;
   paramOnlyVariables?: readonly string[];
   hasReturnType?: boolean;
+  singleActionFunctionName?: string;
```
thegedge, line 114: I don't see this one used anywhere in this PR, but it's indeed in there, so game for fully specifying the shape of this.

thegedge (approved):
🚀

PR #353, opened 2024-04-12
Namespaced model support

airhorns:
## PR Checklist


packages/api-client-core/src/GadgetFunctions.ts
```diff
@@ -98,7 +103,7 @@ interface BulkActionWithInputs<OptionsT, VariablesT> {
 export interface ActionFunctionMetadata<OptionsT, VariablesT, SelectionT, SchemaT, DefaultsT, IsBulk> {
   type: "action";
   operationName: string;
-  namespace: string | null;
+  namespace: string | string[] | null;
```
airhorns, line 106: This is why we have to use this annoying type -- because existing generated clients with api-client-cores expect the namespace to be a string in the past, we can't just always pass a string array.

packages/api-client-core/src/operationBuilders.ts
```diff
@@ -5,10 +5,14 @@ import type { AnyActionFunction } from "./index.js";
...
-const hydrationOptions = (modelApiIdentifier: string): BuilderFieldSelection => {
+const hydrationOptions = (modelApiIdentifier: string, namespace?: string | string[] | null): BuilderFieldSelection => {
+  const fullyQualifiedIdentifier = namespace
+    ? [...(Array.isArray(namespace) ? namespace : [namespace]), modelApiIdentifier].join(".")
+    : modelApiIdentifier;
+
   return {
     gadgetMeta: {
-      [`hydrations(modelName: "${modelApiIdentifier}")`]: true,
+      [`hydrations(modelName: "${fullyQualifiedIdentifier}")`]: true,
```
alanko0511, line 15: Have we handled the dot notation style of "namespace + API identifier" on the Gadget side?

alanko0511 (approved):

PR #373, opened 2024-04-21
handle shopify managed installation

infiton:
To support shopify managed installs I am changing our Shopify app bridge provider to do initial setup with an app by calling a mutation that may have as a side effect that a shop is installed as part of loading the app for the first time. We need to do this because with Shopify managed installation there is no prior communication with a Gadget app during the installation process and the first request to the Gadget backend will come the first time the app is loaded in an embedded context. This mutation will detect this scenario and will install the app Gadget side using a session token exchange

## PR Checklist

### DIFF
--- a/packages/api-client-core/spec/mockUrqlClient.ts
+++ b/packages/api-client-core/spec/mockUrqlClient.ts
@@ -2,7 +2,8 @@ import type { Client, GraphQLRequest, OperationContext, OperationResult, Operati
 import { createRequest, makeErrorResult } from "@urql/core";
 import type { DocumentNode, ExecutionResult, OperationDefinitionNode } from "graphql";
 import type { SubscribePayload, Client as SubscriptionClient, Sink as SubscriptionSink } from "graphql-ws";
-import { find, findLast } from "lodash";
+import { defaults, find, findLast } from "lodash";
+import pRetry from "p-retry";
 import { act } from "react-dom/test-utils";
 import type { Sink, Source, Subject } from "wonka";
 import { filter, makeSubject, pipe, subscribe, take, toPromise } from "wonka";
@@ -36,6 +37,11 @@ export type MockOperationFn = jest.Mock & {
    * One should ensure the appropriate `executeXYZ` call has been made by urql, then call this function.
    */
   pushResponse: (key: string, response: Omit<OperationResult, "operation">) => void;
+  /**
+   *
+   * Waits for a subject to be created for a given key. This is useful for ensuring waiting in a test for a query or mutation to be run
+   */
+  waitForSubject: (key: string) => Promise<void>;
 };
 
 export type MockFetchFn = jest.Mock & {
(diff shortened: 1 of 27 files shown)


packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -84,30 +79,42 @@ const InnerGadgetProvider = memo(
...
+      appBridge
+        .idToken()
+        .then((shopifySessionToken) => {
+          console.debug("[gadget-rsab] fetching or installing shop");
+          fetchOrInstallShop({ shopifySessionToken }).catch((err) => {
+            console.error({ err }, "[gadget-rsab] error fetching or installing shop");
+          });
+        })
+        .catch(() => {
+          console.error("[gadget-rsab] failed to get id token from app bridge");
```
airhorns, line 111: Can we return this error to the caller somehow instead of just swallowing it? Without that, someone can't show a toast or something if there's a network connectivity error or a 500 or whatever (or even know that it happened)

packages/api-client-core/spec/mockUrqlClient.ts
```diff
@@ -115,6 +121,22 @@ const newMockOperationFn = (assertions?: (request: GraphQLRequest) => void) => {
     });
   };
 
+  fn.waitForSubject = async (key: string, options?: pRetry.Options) => {
```
airhorns, line 124: 🤘

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -23,16 +23,11 @@ type GadgetProviderProps = {
...
-      id
-      shop {
-        id
-      }
-    }
-    shopifyConnection {
-      requiresReauthentication
+const FetchOrInstallShopMutation = `
+  mutation ShopifyFetchOrInstallShop($shopifySessionToken: String!) {
+    shopifyFetchOrInstallShop(shopifySessionToken: $shopifySessionToken) {
```
airhorns, line 28: Does it make sense to nest this mutation under the shopifyConnection doodad like previously so as to not pollute their GraphQL namespace?

airhorns (commented):
LMK when the backend bit is up and I can tophat if you like!

packages/react-shopify-app-bridge/src/Provider.tsx
infiton: yah definitely

packages/react-shopify-app-bridge/src/Provider.tsx
infiton: ok made a change to support this; lmk what you think of the approach (I'm not in love with it)

infiton:
@airhorns instructions for 🎩 in https://github.com/gadget-inc/gadget/pull/11154

infiton:
@airhorns @jasong689 this one look ready to go?

airhorns (approved):

jasong689 (approved):
lgtm. is this breaking change (since it's a minor version bump). just wondering we need any changelog for the gadget managed package upgrade modal

infiton:
> lgtm. is this breaking change (since it's a minor version bump)?

no should work backwards compatible as long as the apps graphql schema gets regenerated (which it will when you update the packages)

PR #374, opened 2024-04-22
Add live query support to the imperative operation runners

airhorns:
We shipped all the backend support for live queries and the frontend support to the react client only. This adds support to the helper functions we use within the imperative, non-react JS client for live queries also!

I chose to use async iterators for this api. Calls will look like this:

```typescript
for await (const widget of api.widget.findOne("123")) {
  console.log(widget);
}
```

`widget` there is a fully formed `GadgetRecord` instance, and each time it changed on the backend, we push a whole new instance out the async iterator.

This async iterator approach is the smallest amount of code, and doesn't couple our external interface to any one library's interface for observables or event emitters or whatever. It just uses JS builtins (async iterators), and those can be adapted outside of the client to anything you want. This means a minimal bundle size increase, and we already depend on async iterator support within the client for live queries, so no change in required browser features.

## PR Checklist

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -40,6 +40,7 @@
     "tiny-graphql-query-compiler": "^0.2.2",
     "tslib": "^2.6.2",
     "type-fest": "^3.3.0",
+    "wonka": "^6.3.2",
     "ws": "^8.17.0"
   },
   "devDependencies": {
@@ -51,7 +52,6 @@
     "p-retry": "^4.5.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "tiny-graphql-query-compiler": "workspace:*",
     "typescript": "5.4.5"
   }
 }
--- a/packages/api-client-core/spec/helpers.ts
+++ b/packages/api-client-core/spec/helpers.ts
@@ -1,4 +1,6 @@
 import { parse } from "graphql";
+import { defaults } from "lodash";
+import pRetry from "p-retry";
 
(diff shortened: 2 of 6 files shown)


airhorns:
ping sweet, dear @scott-rc

scott-rc (approved):
Sorry for the delay!

LGTM 👍

packages/api-client-core/package.json
```diff
@@ -40,6 +40,7 @@
     "tiny-graphql-query-compiler": "^0.2.2",
     "tslib": "^2.6.2",
     "type-fest": "^3.3.0",
+    "wonka": "^6.3.2",
```
scott-rc, line 43: I'm guessing we always had this transitively so it doesn't affect bundle size?

airhorns: Correct!

airhorns:
* **#386** <a href="https://app.graphite.dev/github/pr/gadget-inc/js-clients/386?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#374** <a href="https://app.graphite.dev/github/pr/gadget-inc/js-clients/374?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @airhorns and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

PR #379, opened 2024-04-22
Add support for using the `fetch` global from node, and for importing cross-fetch in ESM contexts

airhorns:
A user on discord reported that the way we import cross-fetch doesn't work in a native ESM context:
CleanShot 2024-04-22 at 18 06 49@2x

I can't reproduce it -- when I import `cross-fetch` in a native ESM context locally, I get a module with `fetch` and `default` defined, but I don't really care -- we need to work in next.js. This fixes that twice: first, by using `globalThis.fetch` for our fetch implementation if it is found, which in recent nodes, it is. We used to only check for `window.fetch`, but `globalThis.fetch` is a thing in node (and uses fancy caching smarts in next.js by default too). The second fix is to make the fallback to cross-fetch work if what is returned is a naked ESM module with just the `default` prop that we need to access. 
Yucky.

Annoyingly, the change to start using `globalThis.fetch` breaks our nock-ing setup, because it starts using node's built in fetch, which nock doesn't mock, instead of falling back to `cross-fetch`, which nock does mock as it uses raw `node:http` under the hood. So, to keep knock working, we patch `globalThis.fetch` in the jest env to still use cross-fetch, which gets mocked as before. Nock has forthcoming support for fetch, which I tried, it does not work well yet, so I feel ok about this in the medium term until we can use nock proper.

Extra yucky.

## PR Checklist

### DIFF
--- a/packages/api-client-core/spec/imports/cjs-import.js
+++ b/packages/api-client-core/spec/imports/cjs-import.js
@@ -1,3 +1,6 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { GadgetConnection } = require("../..");
-new GadgetConnection({ endpoint: "/test" });
+const connection = new GadgetConnection({ endpoint: "/test" });
+void connection.fetch("https://gadget.dev").then(() => {
+  console.log("ok");
+});
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -74,6 +74,8 @@ export enum AuthenticationMode {
   Custom = "custom",
 }
 
+const objectForGlobals = typeof globalThis != "undefined" ? globalThis : typeof window != "undefined" ? window : undefined;
+
 /**
  * Root level database connection that Actions can use to mutate data in a Gadget database.
  * Manages transactions and the connection to a Gadget API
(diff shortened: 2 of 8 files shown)


packages/api-client-core/spec/imports/cjs-import.js
```diff
@@ -1,3 +1,6 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { GadgetConnection } = require("../..");
-new GadgetConnection({ endpoint: "/test" });
+const connection = new GadgetConnection({ endpoint: "/test" });
+void connection.fetch("https://gadget.dev").then(() => {
+  console.log("ok");
```
infiton, line 5: these for debugging or should they have an impact on the test?

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -109,11 +111,17 @@ export class GadgetConnection {
...
+      this._fetchImplementation = objectForGlobals.fetch.bind(objectForGlobals);
     } else {
       this._fetchImplementation = async (...args: [any]) => {
-        const { fetch } = await import("cross-fetch");
+        // lazily import cross-fetch to avoid bundling it in the client
+        let fetch: typeof globalThis.fetch = (await import("cross-fetch")) as any;
+
+        // when compiled to CJS, the dynamic import above returns the default export right away. in ESM though, we need to access it ourselves
+        if ((fetch as any).default) {
+          fetch = (fetch as any).default;
```
infiton, line 123: we were previously using essentially `fetch.fetch`; is the idea that `fetch.default` will be present in both contexts and be the right thing? or is the difference here moving to cross-fetch 4?

airhorns: Good question, and to be completely honest, I don't super understand. If I run this script (raw JS, using real ESM syntax):

```javascript
import fetch from "cross-fetch";
console.log(fetch)
```

I get this:
```
❯ node packages/api-client-core/spec/imports/test.mjs
<ref *1> [Function: fetch] {
  ponyfill: true,
  fetch: [Circular *1],
  Headers: [class Headers],
  Request: [class Request],
  Response: [class Response],
  default: [Circular *1]
}
```

the imported value has a `fetch` prop *and* a `default` prop. That's true for cross-fetch v3 and v4. But, in the discord user's `next.js` setup, the imported value obviously doesn't have a `fetch` key, and just had this default key. So, I am unsure what piece of the build chain is different or not, but, I know that the default prop will point to the right thing if it does exist, so I think it is safe to just always use it.

infiton: :+1:

infiton (approved):

PR #385, opened 2024-04-29
Background action result types

garymardell:
Adds typing to background action handles that determines the result type based on the Action being passed in. This means that actions such as `MockWidgetCreateAction` will be typed as a `GadgetRecord`. I had to follow a similar pattern to what is in `useActionForm` and pass through a `SchemaT` type otherwise the typescript compiler complains when inferring the `GadgetRecord` shape. This type also has to be propagated back up into the `useEnqueue` react package in order to properly construct the correct `BackgroundHandle` and result type. 

I added additional typing to the `MockWidgetCreateAction` to better reflect the types that an actual action will have when constructed by gadget.

Note: I still need to test it with gadget codebase but that has been causing me some difficulties...

## PR Checklist

### DIFF
--- a/packages/api-client-core/package.json
+++ b/packages/api-client-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.15.23",
+  "version": "0.15.24",
   "files": [
     "Readme.md",
     "dist/**/*"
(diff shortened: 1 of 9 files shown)


airhorns:
You did it! You brave, brave soul! I think it makes sense that you had to do do the wild amount of generics in exactly the same pattern to get it to work.

garymardell:
I still haven't got this quite right as the `select` options are not being applied against the `GadgetRecord` so unselected fields are not omitted. Figuring this out now.

garymardell:
Update: Just pushed a commit and finally figured it out 🎉 will do another integration against gadget to double check

PR #386, opened 2024-04-30
Dynamic import the live query implementation for smaller initial bundle size

airhorns:
Most Gadget users don't execute live queries, and the code to power them is kinda big, it's about 10% of our bundle size. To try to help folks keep their LCPs down, I think we should defer loading it until someone executes the first live query. It's a quick request for that JS at that time, and live queries are already set up to stream results, so the deferred loading can be managed pretty easily. 

While working on this, I also noticed that the live query utils were importing all of `graphql`! `urql` uses a slimmed down version of `graphql` internally called `@0no-co/graphql` that has a much smaller bundle size, and we were accidentally including both it and the mainline graphql via `graphql-live-query`. We were only using one very small function from the latter library, so I just vendored it into Gadget and now we can avoid that dep at runtime altogether. Woop woop!

The result: 25% reduction in initial bundle size for the plain api-client, 21% reduction overall if you do load a live query! Woo!

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { isLiveQueryOperationDefinitionNode } from "@n1ru4l/graphql-live-query";
-import { applyAsyncIterableIteratorToSink, makeAsyncIterableIteratorFromSink } from "@n1ru4l/push-pull-async-iterable-iterator";
 import type { ClientOptions, RequestPolicy } from "@urql/core";
 import { Client, cacheExchange, fetchExchange, subscriptionExchange } from "@urql/core";
 import type { ExecutionResult } from "graphql";
@@ -14,7 +13,6 @@ import type { BrowserStorage } from "./InMemoryStorage.js";
 import { InMemoryStorage } from "./InMemoryStorage.js";
 import { operationNameExchange } from "./exchanges/operationNameExchange.js";
 import { urlParamExchange } from "./exchanges/urlParamExchange.js";
-import { applyLiveQueryJSONDiffPatch } from "./graphql-live-query-utils/index.js";
 import {
   GadgetTooManyRequestsError,
   GadgetUnexpectedCloseError,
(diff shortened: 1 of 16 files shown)


airhorns:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/gadget-inc/js-clients/386?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#386** <a href="https://app.graphite.dev/github/pr/gadget-inc/js-clients/386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#374** <a href="https://app.graphite.dev/github/pr/gadget-inc/js-clients/374?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @airhorns and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -384,21 +383,37 @@ export class GadgetConnection {
...
-                      this.getBaseSubscriptionClient().subscribe(input, sink as Sink<ExecutionResult>)
-                    )
-                  ),
-                  sink
-                ),
+                unsubscribe: () => {
+                  if (unsubscribe) {
+                    unsubscribe();
+                  } else {
+                    void loadAndSubscribe.then((unsubscribe) => unsubscribe());
```
scott-rc, line 414: Should we attach a `.catch` to this?

scott-rc (approved):
Nice! 🎉

airhorns:
ping @danroberts

danroberts (approved):
This seems more like a case for tree shaking the bundle, but 20% reduction seems good in the meantime!

airhorns:
The problem with tree shaking it is only works on module level imports, and we've made live-ness an option to everything which makes it much more convenient but much less treeshakeable.

PR #388, opened 2024-05-01
refresh session token on every request

infiton:
made a big mistake in https://github.com/gadget-inc/js-clients/pull/373 where we only request a session token from the app bridge once when the provider mounts and then reuse that across it's lifetime; only issue: the token has a 60 second expiry :(

We need to request. a new token each time the client makes a request

## PR Checklist

### DIFF
--- a/packages/react-shopify-app-bridge/package.json
+++ b/packages/react-shopify-app-bridge/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react-shopify-app-bridge",
-  "version": "0.15.0",
+  "version": "0.15.1",
   "files": [
     "README.md",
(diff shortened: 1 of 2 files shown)


airhorns (approved):
LGTM! I think it makes sense to ship this ASAP, but can we add a test after?

jcao49 (approved):

infiton:
> LGTM! I think it makes sense to ship this ASAP, but can we add a test after?

was trying to add a test quick; unfortunately we bypass all the fetch goodies with the mockUrql client so I think we need to do some work to have the mockUrqlClient go through the real clients fetch setup

infiton:
but yah; will make sure to back through and test this

PR #390, opened 2024-05-04
we need the react-shopify-app-bridge Provider to show a loading state…

infiton:
@DevAOC noticed that when using `0.15.1` there is an annoying flash in our default Shopify template where we show the unauthenticated app very quickly. This happens when `useGadget` returns `{isAuthenticated: false, loading: false}`

On closer inspection it looks like this flash has been with us for awhile; here are the state updates:

`0.14.2`:

Screen Shot 2024-05-03 at 9 05 18 PM

`0.15.1`:

Screen Shot 2024-05-03 at 9 08 42 PM

Notice that in `0.14.1` it is only there for one render; in `0.15.1` for two renders, this is because there are two reasons for this:

1. `loading` could never be true if `isRootFrameRequest` is true; in both versions on first render `isRootFrameRequest` is true because in the outer provider we haven't yet `setLocation`; when the inner provider renders its `useEffect`s run first and trigger a state update with `isRootFrameRequest: true`, which leads to a render with `loading: false` 😢 
2. In `0.15` we moved to authentication happening with `useMutation` and not `useQuery`; as a result when the provider first mounts the mutation to do auth isn't running; we need to use a ref to keep us in a loading state until the first time that the auth mutation runs; blerg

I don't really see why `isRootframeRequest` stopped us from being in a loading state, I think it could be a hold over from before we split the providers out. In any case the `InnerGadgetProvider` is only used is the shopify app bridge is defined and the app type is embedded so I think it makes good sense that we are in a loading state until the auth mutation resolves.

## PR Checklist

### DIFF
--- a/packages/react-shopify-app-bridge/spec/Provider.spec.tsx
+++ b/packages/react-shopify-app-bridge/spec/Provider.spec.tsx
@@ -4,6 +4,7 @@ import * as AppBridgeReact from "@shopify/app-bridge-react";
 import "@testing-library/jest-dom";
 import { render } from "@testing-library/react";
 import React from "react";
+import { act } from "react-dom/test-utils";
 import { mockUrqlClient } from "../../api-client-core/spec/mockUrqlClient.js";
 import { AppType, Provider } from "../src/Provider.js";
 
@@ -14,6 +15,7 @@ describe("GadgetProvider", () => {
   const mockOpen = jest.fn();
   const mockApiKey = "some-api-key";
   let useAppBridgeMock: jest.SpyInstance;
+  let resolveIdToken: (value: string) => void;
 
   describe.each([true, false])("as install request: %s", (isInstallRequest) => {
     beforeAll(() => {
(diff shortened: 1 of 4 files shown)


packages/react-shopify-app-bridge/spec/Provider.spec.tsx
```diff
@@ -92,7 +97,10 @@ describe("GadgetProvider", () => {
         </Provider>
       );
 
-      await mockUrqlClient.executeMutation.waitForSubject("ShopifyFetchOrInstallShop");
+      await act(async () => {
```
infiton, line 100: added these to get rid of the logs saying that state was being updated outside of an `act` block; we call a mutation when the appBridge.idToken resolves so in order to have that state update be in an act block we need to resolve that promise like this

packages/react-shopify-app-bridge/spec/useGadget.spec.tsx
```diff
@@ -196,4 +207,64 @@ describe("useGadget", () => {
     // standalone apps are always falsy
     expect(result.current.isRootFrameRequest).toBeFalsy();
   });
+
+  test("has loading true for embedded apps until ShopifyFetchOrInstallShop resolves", async () => {
```
infiton, line 211: this test would fail on `0.14.2` and `0.15.1`

airhorns (approved):
Dang, that little ref is sneaky, but this makes good sense to me! So glad we have tests that let you prove this works, nice!!

PR #400, opened 2024-05-11
React hooks for namespaced models / global actions tests

airhorns:
Adds tests for the functionality added to the react hooks in https://github.com/gadget-inc/js-clients/pull/353 that didn't have tests

## PR Checklist

### DIFF
--- a/packages/react/spec/useAction.spec.tsx
+++ b/packages/react/spec/useAction.spec.tsx
@@ -143,7 +143,16 @@ describe("useAction", () => {
   });
 
   test("returns no data, fetching=true, and no error when the mutation is run, and then the successful data if the mutation succeeds", async () => {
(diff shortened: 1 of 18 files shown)


packages/react/spec/useAction.spec.tsx
```diff
@@ -254,6 +326,153 @@ describe("useAction", () => {
...
+    });
+
+    expect(result.current[0].data).toBeFalsy();
+    expect(result.current[0].fetching).toBe(true);
+    expect(result.current[0].error).toBeFalsy();
+
+    expect(query).toMatchInlineSnapshot(`
+      "mutation updatePlayer($id: GadgetID!, $player: UpdateGamePlayerInput) {
+        game {
+          updatePlayer(id: $id, player: $player) {
```
alanko0511, line 353: The `pushResponse` function below could not catch the issue of not including the model namespace in these react hooks because the GraphQL errors come from the server.

I also added these snapshots to ensure we are sending the expected queries.

danroberts (approved):
👍

PR #407, opened 2024-05-17
fix(tiny-qql): lift field call subselect call variables

danroberts:
When we compile a FieldCall, we lift the variables up to the top level query inputs.

We handle the case where this happens to a subfield, but we're not handling the case where there is a subfield in the selection of a FieldCall itself which is also a FieldCall.

In that case, we were failing to contribute the variables of the subfield call during `extractVariables`

## PR Checklist


alanko0511 (approved):
:shipit: 👌

airhorns (approved):
Ah ha, good find!

scott-rc (approved):

PR #410, opened 2024-05-21
adds shopify-ui-extensions package

infiton:
This is an alternative to https://github.com/gadget-inc/gadget/pull/11509 where we will make things even easier for extension development; the code for i.e. a customer ui extension ends up being:

```js
import { useApi } from "@shopify/ui-extensions-react/customer-account";
import { Provider } from "@gadgetinc/shopify-ui-extensions/react";
import { api } from "./api";

function MenuActionModelExtensionProvider() {
  const { sessionToken } = useApi();

  return (
    <Provider api={api} sessionToken={sessionToken}>
      <MenuActionModalExtension />
    </Provider>
  );
}
```

and then in the non-react world it looks like:

```js
import { registerShopifySessionTokenAuthentication } from "@gadgetinc/shopify-ui-extensions";
import { api as gadgetApi } from "./api";

export default extension('customer-account.order.action.menu-item.render', (root, api) => {
  const { sessionToken } = api;

  registerShopifySessionTokenAuthentication(gadgetApi, sessionToken);

  root.appendChild(
    root.createComponent(
      Button,
      { to: 'extension:/' },
      i18n.translate('menuItemButton')
    )
  );
});
```

## PR Checklist


infiton:
thoughts on approach/api @airhorns @rdraward

packages/shopify-ui-extensions/src/react/context.ts
```diff
@@ -0,0 +1,16 @@
...
+  /** An instance of the Gadget application's api client set up to use Shopify session token authentication */
+  api: AnyClient;
+};
+
+export const GadgetShopifyUiExtensionContext = createContext<GadgetShopifyUiExtensionContextValue>({
+  ready: false,
+  api: null as unknown as AnyClient,
+});
+
+export const useGadget = () => useContext<GadgetShopifyUiExtensionContextValue>(GadgetShopifyUiExtensionContext);
```
airhorns, line 16: I think it might be good to make this generic so that folks can do

```typescript
useGadget<MyClient>().api // => MyClient
```

```suggestion
export type GadgetShopifyUIExtensionContextValue<Client extends AnyClient = AnyClient> = {
  /** Whether or not the api client is ready to make requests to with Shopify session token authentication */
  ready: boolean;
  /** An instance of your Gadget app's api client, which has been set up to use Shopify session token authentication */
  api: Client;
};

export const GadgetShopifyUIExtensionContext = createContext<GadgetShopifyUIExtensionContextValue>({
  ready: false,
  api: null as unknown as AnyClient,
});

export const useGadget = <Client extends AnyClient = AnyClient>() => useContext<GadgetShopifyUIExtensionContextValue<Client>>(GadgetShopifyUIExtensionContext);
```

airhorns (approved):
LGTM

packages/shopify-ui-extensions/src/react/context.ts
infiton: :+1:

rdraward:
I can dig it, setting via an extension-specific `Provider` works for me

We'll need to document this in our "Working with extensions" page (and update tutorials, samples, templates) as well as write up a Reference guide for this package

PR #415, opened 2024-05-28
Simpler/smaller/more performant GadgetRecord

airhorns:
While poking at the `GadgetRecord` class for some other stuff, I noticed some inefficiency in how the change tracking is implemented. `GadgetRecord` is a strange object because it is a class that has all our business logic wrapped in a proxy that allows us to do the change tracking. The objects we had back into userland are `Proxy` objects that intercept all `get` and `set` calls to do field change tracking. The actual `GadgetRecord` object sits "next" to the proxy, in that it isn't even the proxy target, but instead an object that is closed over by the proxy to provide some of the functionality. 

Previously, `GadgetRecord` stuck all its private state on a `__gadget` object on it. This adds one more object allocation to every created record object and an unnecessary layer of indirection. the `GadgetRecord` object itself is already an object with a different identity than the place where we store all the actual keys/values of the record. I think `__gadget` was from before that was the case. So, this removes it in favour of just storing the bits right on the `GadgetRecord` itself. To keep the proxy logic simple for what is a field and what is a special key, I stored all the bits under symbols, which still makes it pretty clear that it is special.

I also removed one of the special bits as I think it was redundant, we had a list of the keys we were tracking in an array, as well as a map of {[key]: true}, both always set at the same time.

### DIFF
--- a/packages/api-client-core/src/GadgetRecord.ts
+++ b/packages/api-client-core/src/GadgetRecord.ts
@@ -9,66 +9,75 @@ export enum ChangeTracking {
 
 export type RecordShape = Record<string, any> | null | undefined | void;
 
+const kFields = Symbol.for("g/fields");
+const kInstantiatedFields = Symbol.for("g/if");
+const kPersistedFields = Symbol.for("g/pf");
+const kFieldKeys = Symbol.for("g/fk");
+const kTouched = Symbol.for("g/t");
+
 /** Represents one record returned from a high level Gadget API call */
-export class GadgetRecordImplementation<Shape extends RecordShape> {
-  private __gadget = {
-    fields: {} as any,
-    instantiatedFields: {} as any,
-    persistedFields: {} as any,
-    fieldKeys: [] as string[],
-    fieldKeysTracker: {} as Record<string, boolean>,
-    touched: false,
-  };
+export class GadgetRecord_<Shape extends RecordShape> {
+  /** Storage of the actual keys and values of this record */
+  [kFields]: Record<string, any> = {};
+  /** Storage of the keys and values of this record at the time it was instantiated */
+  [kInstantiatedFields]: Record<string, any> = {};
+  /** Storage of the keys and values of this record at the time it was last persisted */
+  [kPersistedFields]: Record<string, any> = {};
+  /** Storage of the keys and values of this record at the time it was last persisted */
+  [kFieldKeys]: Set<string>;
+  [kTouched] = false;
 
   private empty = false;
 
   constructor(data: Shape) {
-    this.__gadget.instantiatedFields = cloneDeep(data);
-    this.__gadget.persistedFields = cloneDeep(data);
-    Object.assign(this.__gadget.fields, data);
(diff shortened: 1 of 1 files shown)


packages/api-client-core/src/GadgetRecord.ts
```diff
@@ -9,66 +9,75 @@ export enum ChangeTracking {
...
+  [kTouched] = false;
 
   private empty = false;
 
   constructor(data: Shape) {
-    this.__gadget.instantiatedFields = cloneDeep(data);
-    this.__gadget.persistedFields = cloneDeep(data);
-    Object.assign(this.__gadget.fields, data);
+    this[kInstantiatedFields] = cloneDeep(data) ?? {};
+    this[kPersistedFields] = cloneDeep(data) ?? {};
```
airhorns, line 34: I tried making this lazy but it breaks our contracts around detecting nested changes, so we still gotta do this eagerly on instantiation :(

packages/api-client-core/src/GadgetRecord.ts
```diff
@@ -9,66 +9,75 @@ export enum ChangeTracking {
 
 export type RecordShape = Record<string, any> | null | undefined | void;
 
+const kFields = Symbol.for("g/fields");
+const kInstantiatedFields = Symbol.for("g/if");
+const kPersistedFields = Symbol.for("g/pf");
+const kFieldKeys = Symbol.for("g/fk");
+const kTouched = Symbol.for("g/t");
```
thegedge, line 16: nit: Any reason to not go for clearer names?

packages/api-client-core/src/GadgetRecord.ts
```diff
@@ -9,66 +9,75 @@ export enum ChangeTracking {
...
     }
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this;
 
     const handler = {
       get: (obj: any, prop: string | symbol) => {
-        if (prop in this) {
-          // if the GadgetRecordImplementation responds to the property or function
-          // prioritize that
-          return (this as any)[prop];
```
thegedge, line 42: `this === handler` here, eh? So it wasn't doing anything useful :o

thegedge (approved):
🚀

packages/api-client-core/src/GadgetRecord.ts
airhorns: Yeah, bundle size -- these strings don't get mangled during minification. EVERY BYTE COUNTS

packages/api-client-core/src/GadgetRecord.ts
airhorns: Yup, or just only doing the copy the first time we need to write something. But, we have tests that say "modifying an array property marks that property as changed", which means we need to either proxy nested stuff in order to catch the write to the nested doodad, or just keep eagerly deepcloning. It... sucks.

airhorns:
Gadget CI passes with this here: https://buildkite.com/gadget/gadget/builds/23302#018fc0c4-ad50-47cc-9398-d80ffb0eda50

PR #418, opened 2024-06-04
Ensure adding more url params to a websocket endpoint correctly appends them

airhorns:
If the base websocket endpoint already had url params, the naive approach we were using before incorrectly appended `?` twice. This uses the browser builtin for adding url params that we actually already have code for to do so.


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -437,7 +437,7 @@ export class GadgetConnection {
     return client;
   }
 
-  private newSubscriptionClient(overrides: GadgetSubscriptionClientOptions) {
+  newSubscriptionClient(overrides?: GadgetSubscriptionClientOptions) {
```
scott-rc, line 440: Did you mean to make `overrides` required?

scott-rc (approved):
LGTM 👍

packages/api-client-core/src/GadgetConnection.ts
airhorns: I think that was an accident from before -- the whole code deals with it being optional already, and i think overrides shouldn't be mandatory

PR #419, opened 2024-06-14
Initial <AutoForm /> implementation

danroberts:
This is the initial up to date implementation of @airhorns original autocomponents PR, trying to establish a feature branch 'feature/autoform' for continued work on the AutoForm component. 

## PR Checklist


packages/react/package.json
```diff
@@ -3,66 +3,135 @@
...
+      "optional": true
+    },
+    "@mui/x-data-grid": {
+      "optional": true
+    },
+    "@mui/x-date-pickers": {
+      "optional": true
+    },
+    "@shopify/polaris": {
+      "optional": true
```
airhorns, line 131: Can you validate that gadget's yarn setup won't auto-install all these optional deps?

packages/react/spec/imports/cjs-import.js
```diff
@@ -1,3 +1,3 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const { Provider } = require("../..");
+const { Provider } = require("@gadgetinc/react");
```
airhorns, line 2: I am not sure this is the same thing -- did you have to make this change?

packages/react/src/auto/AutoFormContext.ts
```diff
@@ -0,0 +1,34 @@
...
+  metadata: ActionMetadata | undefined;
+  submit: UseActionFormSubmit<any>;
+}
+/**
+ * React context that stores an instance of the JS Client for an app (AKA the `api` object)
+ */
+
+export const AutoFormMetadataContext = React.createContext<AutoFormMetadataContext | undefined>(undefined);
+/**
+ * Get the current `api` object from React context
```
airhorns, line 17: needs updating

packages/react/src/auto/AutoFormContext.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { ActionMetadata } from "../metadata.js";
+
+/**  */
+
+export interface AutoFormMetadataContext {
+  metadata: ActionMetadata | undefined;
+  submit: UseActionFormSubmit<any>;
+}
+/**
+ * React context that stores an instance of the JS Client for an app (AKA the `api` object)
```
airhorns, line 12: needs updating

packages/react/src/auto/mui/MUIAutoForm.tsx
```diff
@@ -0,0 +1,83 @@
...
+        </Grid>
+      ))}
+      <Grid item xs={12}>
+        <LoadingButton loading={isLoading} type="submit">
+          {(props.submitLabel as any) ?? "Submit"}
+        </LoadingButton>
+      </Grid>
+    </Grid>
+  );
+};
```
airhorns, line 83: Is this stuff worth landing now do you think? Is it working the same way?

packages/react/src/auto/index.tsx
```diff
@@ -0,0 +1,5 @@
+import type { AutoFormProps } from "./AutoForm.js";
+
+export interface AutoAdapter<T = any, F = any> {
+  AutoForm: React.ComponentType<AutoFormProps<any, any, any, any> & F>;
```
airhorns, line 4: Doesn't this need all the other components in it too?

packages/react/src/auto/polaris/PolarisFormInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+    case FieldType.RoleAssignments: {
+      return <PolarisRolesCombobox label={props.field.name} {...field} />;
+    }
+    case FieldType.HasMany:
+    case FieldType.HasManyThrough: {
+      return <PolarisHasManyInput field={props.field.apiIdentifier} />;
+    }
+    case FieldType.HasOne:
+    case FieldType.BelongsTo: {
+      // TODO: implement relationship field inputs
```
airhorns, line 90:
```suggestion
```

packages/react/src/auto/polaris/PolarisJSONInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+      multiline={4}
+      monospaced
+      value={string}
+      onChange={onStringChange}
+      error={!isFocused && error && `Invalid JSON: ${error.message}`}
+      {...focusProps}
+      {...rest}
+    />
+  );
+};
```
airhorns, line 39: Is there a ticket to track making this nice

airhorns:
CleanShot 2024-06-18 at 15 45 40@2x
tree shaking -- it works

packages/react/src/auto/polaris/inputs/PolarisAutoInput.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { PolarisBelongsToInput } from "./PolarisBelongsToInput.js";
+
+const FieldTypeToInputType: Partial<Record<GadgetFieldType, TextFieldProps["type"]>> = {
+  [FieldType.Number]: "number",
+  [FieldType.Email]: "email",
+  [FieldType.Password]: "password",
+  [FieldType.EncryptedString]: "password",
+};
+
+export const PolarisAutoInput = (props: { field: string }) => {
```
airhorns, line 23: What's the difference between this and PolarisFormInput

packages/react/src/auto/polaris/inputs/PolarisBelongsToInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  const modelManager = assert(
+    (api as any)[config.relatedModel.apiIdentifier] as AnyModelManager,
+    "no model manager found for action function"
+  );
+
+  const [{ data, fetching, error }, _refetch] = useFindMany(modelManager as any, { first: 25 });
+
+  if (fetching || !data) {
+    return <p>Loading...</p>;
```
airhorns, line 48: Doesnt have to be this PR but can we do a spinny on the select instead of a layout shift from `<p/>` => `<Select/>`

packages/react/src/use-action-form/types.ts
```diff
@@ -54,6 +54,8 @@ export type UseActionFormResult<
   actionData?: ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any>
     ? ReturnType<typeof useAction<GivenOptions, SchemaT, ActionFunc, any>>[0]["data"]
     : ReturnType<typeof useGlobalAction<any>>[0]["data"];
+
+  originalFormMethods: UseFormReturn<FormVariables, FormContext>;
```
airhorns, line 58: Did you end up actually needing this?

packages/react/src/auto/mui/MUIFormInput.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { MUIRolesCombobox } from "./MUIRolesCombobox.js";
+
+const FieldTypeToInputType: Partial<Record<GadgetFieldType, TextFieldProps["type"]>> = {
+  [FieldType.Number]: "number",
+  [FieldType.Email]: "email",
+  [FieldType.Password]: "password",
+  [FieldType.EncryptedString]: "password",
+};
+
+export const MUIAutoFormControl = (props: { path: string; control: Control<any>; field: FieldMetadata; children: ReactElement }) => {
```
MillanWangGadget, line 22: `field: FieldMetadata` gets confusing where we have other places where we have `field: string` for the string api id. Consider renaming this to `fieldMetadata: FieldMetadata`

packages/react/src/auto/polaris/PolarisAutoForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <FormLayout>
+            {fields.map(([path, field]) => (
+              <PolarisFormInput key={field.apiIdentifier} path={path} field={field} control={originalFormMethods.control} />
+            ))}
+            <Button
+              loading={isLoading}
+              submit
+              onClick={async () => {
+                submit;
+              }}
```
MillanWangGadget, line 97: This `onClick` doesn't seem to be able to do anything. Nothing is returned from the func because of the `{}`

packages/react/package.json
danroberts: I've been able to validate that manually -- do you mean an automated test for it?

packages/react/src/use-action-form/types.ts
danroberts: Yes, this was required in order to get types to work for a <FormProvider> which seems required for our plan for custom components

packages/react/src/auto/polaris/inputs/PolarisAutoInput.tsx
danroberts: FormInput takes the metadata as a prop, and AutoInput gets the metadata from the context. I'd like to eventually completely replace FormInput with AutoInput, but would prefer to do that in a followup PR

packages/react/src/auto/polaris/PolarisJSONInput.tsx
danroberts: Yes, currently tracked as https://linear.app/gadget-dev/issue/GGT-6572/implement-a-jsoncode-editor but it need additional details and the figma linked.

packages/react/src/auto/index.tsx
danroberts: I'm not sure I actually follow what this interface does other than abstract between the polaris and mui autoform for tests, in which case we only need it for the autoform at the moment.

packages/react/src/auto/mui/MUIAutoForm.tsx
danroberts: I'm not sure I follow -- do you mean only commit the Polaris stuff?

packages/react/spec/imports/cjs-import.js
danroberts: I don't recall making this change, I'll try reverting it.

packages/react/package.json
airhorns: Nope just manual is good for me 👍

packages/react/src/auto/mui/MUIAutoForm.tsx
airhorns: Yeah -- I am easy either way

packages/react/src/auto/polaris/inputs/PolarisBelongsToInput.tsx
danroberts: 100%

packages/react/spec/imports/cjs-import.js
danroberts: Right, so with "../.." the test fails with:

``` Command failed with exit code 1: node cjs-import.js
    node:internal/modules/cjs/loader:1137
      throw err;
      ^

    Error: Cannot find module '../..'
    Require stack:
    - /private/var/folders/4m/12k7sghx0n56pct6fz0zxf100000gn/T/js-clients-react-imports-z0kpAz/cjs-import.js
    ```
    
    However, looking at the other tests in context, I'm pretty sure requiring the main gadgetinc/react package is what this test is attempting to assert is functioning.
    
    I'm not sure I follow what "../.." was meant to represent in this case, but it seems to me that require("@gadgetinc/react") is more correct logically.

airhorns: Ok

alanko0511 (approved):
:shipit: since this is just the base of all the upcoming work on the auto component project, I'm fine with fixing them forward.

MillanWangGadget (approved):
No immediate show stoppers jump out at me for now. I think this is ready to be turned into a feature branch now that CI is green 🫡

PR #424, opened 2024-06-20
add useFieldMetadata hook & context providers to MUI

MillanWangGadget:
- **UPDATE**
  - Added a shared `useFieldMetadata` hook for commonly used process of turning a field api identifier into a path and field metadata
    - This PR is a small subset of https://github.com/gadget-inc/js-clients/pull/423 - If that PR is blocked, this smaller PR can be merged so that other component work can be unblocked
      - This PR before the other one or not at all

## PR Checklist

### DIFF
--- a/packages/react/src/auto/hooks/useFieldMetadata.tsx
+++ b/packages/react/src/auto/hooks/useFieldMetadata.tsx
@@ -1,0 +1,16 @@
+import { useFormFields } from "../AutoForm.js";
+import { useAutoFormMetadata } from "../AutoFormContext.js";
+
+export const useFieldMetadata = (fieldApiIdentifier: string) => {
+  const { metadata: autoFormMetadata } = useAutoFormMetadata();
+  const fields = useFormFields(autoFormMetadata, {});
+  const targetFieldMetadata = fields.find((field) => field[1].apiIdentifier === fieldApiIdentifier);
+
+  if (!targetFieldMetadata) {
+    throw new Error(`Field ${fieldApiIdentifier} not found in metadata`);
+  }
+  const path = targetFieldMetadata[0];
+  const fieldMetadata = targetFieldMetadata[1];
+
+  return { path, fieldMetadata };
+};
--- a/packages/react/src/auto/mui/MUIAutoForm.tsx
+++ b/packages/react/src/auto/mui/MUIAutoForm.tsx
@@ -3,11 +3,13 @@ import { LoadingButton } from "@mui/lab";
 import type { GridProps } from "@mui/material";
 import { Alert, Grid, Skeleton } from "@mui/material";
 import React from "react";
+import { FormProvider } from "react-hook-form";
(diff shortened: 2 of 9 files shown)


packages/react/src/auto/hooks/useFieldMetadata.tsx
```diff
@@ -0,0 +1,16 @@
+import { useFormFields } from "../AutoForm.js";
+import { useAutoFormMetadata } from "../AutoFormContext.js";
+
+export const useFieldMetadata = (fieldApiIdentifier: string) => {
+  const { metadata: autoFormMetadata } = useAutoFormMetadata();
+  const fields = useFormFields(autoFormMetadata, {});
+  const metadataFromUseFormField = fields.find((field) => field[1].apiIdentifier === fieldApiIdentifier);
```
alanko0511, line 7: Nit: `metadataFromUseFormField` -> `targetFieldMetadata`?

alanko0511 (approved):
:shipit:

packages/react/src/auto/hooks/useFieldMetadata.tsx
MillanWangGadget: Renamed

packages/react/src/auto/mui/MUIAutoForm.tsx
```diff
@@ -36,7 +38,7 @@ export const MUIAutoForm = <
 >(
   props: MUIAutoFormProps<GivenOptions, SchemaT, ActionFunc, Options>
 ) => {
-  const { action, include: _include, exclude: _exclude, fields: _fields, submitLabel: _submitLabel, ...rest } = props;
+  const { action, include: _include, exclude: _exclude, fields: _fields, submitLabel: _submitLabel, record, findBy, ...rest } = props;
```
danroberts, line 41: Unless I'm missing something, I don't think we need to destructure these include/exclude props here.

packages/react/src/auto/hooks/useFieldMetadata.tsx
```diff
@@ -0,0 +1,16 @@
+import { useFormFields } from "../AutoForm.js";
+import { useAutoFormMetadata } from "../AutoFormContext.js";
+
+export const useFieldMetadata = (fieldApiIdentifier: string) => {
```
danroberts, line 4: Could you add a test that checks we load the right field from the metadata? Check https://testing-library.com/docs/react-testing-library/api/#renderhook

packages/react/src/auto/mui/MUIAutoForm.tsx
MillanWangGadget: Originally I made this match with the Polaris equivalent component. I removed the unused de-structured properties in both cases.

It seems like this was here before `useFormFields` handled field inclusion/exclusion and it was not removed. Good catch

danroberts (approved):

PR #425, opened 2024-06-20
Implement boolean input for auto component

alanko0511:
This PR adds an AutoInput for a boolean field and a helper function to generate the mock action metadata GraphQL response.

## PR Checklist


packages/react/src/auto/mui/inputs/MUIBooleanInput.tsx
```diff
@@ -0,0 +1,29 @@
...
+    name: path,
+  });
+
+  return (
+    <MUIAutoFormControl field={props.field}>
+      <Checkbox
+        checked={fieldProps.value}
+        onChange={fieldProps.onChange}
+        onBlur={fieldProps.onBlur}
+        disabled={fieldProps.disabled}
```
MillanWangGadget, line 24: Since these 3 have matching property names, you can do {...fieldProps} instead

packages/react/src/auto/polaris/inputs/PolarisBooleanInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  return (
+    <Checkbox
+      label={metadata.name}
+      checked={!!fieldProps.value}
+      onChange={fieldProps.onChange}
+      onBlur={fieldProps.onBlur}
+      disabled={fieldProps.disabled}
+      name={fieldProps.name}
```
MillanWangGadget, line 28: These  can also be {...fieldProps}

`name` also doesn't appear to do anything here `label` takes over

MillanWangGadget (approved):
TOPHAT - With a gitpkg in a real app
- 🍏 The checkbox properly submits with the selected values
- 🍏 Properly loads data from the `findBy` prop

PR #426, opened 2024-06-20
Replace FormInput with AutoInput and refactored useFormFields to return an array of objects

MillanWangGadget:
- **UPDATES**
  - Removed FormInput components and replaced them with AutoInput
    - Polaris and MUI
  - Refactored useFormFields to return an array of objects instead of an array of arrays
    - `{ path:string, fieldMetadata:FieldMetadata }` feels less risky than using an array and trusting ourselves to pass in the proper index

## PR Checklist


packages/react/src/auto/AutoForm.ts
```diff
@@ -50,16 +50,27 @@ export const useValidationResolver = (metadata: ActionMetadata | undefined) => {
 export const useFormFields = (
   metadata: ActionMetadata | undefined | null,
   options: { include?: string[]; exclude?: string[]; fields?: string[]; select?: Record<string, any> }
-): (readonly [string, FieldMetadata])[] => {
+): readonly { path: string; fieldMetadata: FieldMetadata }[] => {
```
alanko0511, line 53: `field.fieldMetadata` seems a bit weird. How about `field.metadata`?

alanko0511 (approved):

packages/react/src/auto/AutoForm.ts
MillanWangGadget: Updated to just be `metadata`

packages/react/src/auto/mui/inputs/MUIAutoInput.tsx
```diff
@@ -19,36 +17,39 @@ const FieldTypeToInputType: Partial<Record<GadgetFieldType, TextFieldProps["type
...
+        <FormControlLabel label={fieldMetadata.name} control={props.children} />
       </FormGroup>
       {error && <FormHelperText>{error?.message}</FormHelperText>}
     </FormControl>
   );
 };
 
-export const MUIFormInput = (props: { path: string; field: FieldMetadata; control: Control<any> }) => {
+export const MUIAutoInput = (props: { field: string }) => {
+  const { path, metadata: fieldMetadata } = useFieldMetadata(props.field);
```
danroberts, line 40: Maybe we can just call this field metadata -- I think it's clear enough that it is the metadata for the field, and we can resolve the field/fieldMetadata ambiguity?

danroberts (approved):

packages/react/src/auto/mui/inputs/MUIAutoInput.tsx
MillanWangGadget: Updated so that this is just `metadata` for all places that use this hook

PR #427, opened 2024-06-21
Add support for models under namespaces

alanko0511:
This PR adds support for namespaced models, such as the `stadium` model under the `game` folder.

The namespaced models don't exist in the API client's top level; instead, they are under multiple namespace classes. For example, to run a `create` action against the `game/stadium` model, the line of code for running the action would be `api.game.stadium.create()`.

To achieve this, we are going to first convert it into a do-notation path, i.e., `game.stadium`, and then use an existing `get` util function to get the target model manager. Inside the `js-client-test` real Gadget app, we also added some namespaced models with some relationships between them for us to test.

For models without a namespace or apps with a framework version that doesn't support a namespace, the namespace array is either `[]` or `null`, and the `namespaceDataPath` function already handles both types.

## PR Checklist


MillanWangGadget (approved):
- MANUAL TESTS
  - 🍏 In a real app, passing in a namespaced create model action properly generates the form and runs the action on submit. 
  - 🍏 Namespaced model update actions load the fields properly. (Proper submission blocked by other ID ticket)
  - 🍏 Custom params on namespaced model actions work as expected
  - 🍏 Field include and exclude work with namespaced models 
  - 🍏 belongsTo, hasMany relations between namespaced models work
- NOTES
  - The usage of `namespace.length>0` models seems be consistent with how `namepace==[]` models works. 
  - Good to see the namespaced model examples used here
  - Tests can run locally after updating dependencies with the new `"@gadget-client/js-clients-test"`

packages/react/cypress/component/auto/form/AutoForm.cy.tsx
```diff
@@ -32,6 +32,14 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
     cy.contains("Saved Widget successfully");
   });
 
+  // TODO: fix the issue of sending a GraphQL mutation request that the hasMany field value is null.
```
infiton, line 35: I think @pistachiobaby may have fixed this now?

alanko0511: It's not fixed at the GraphQL level (related ticket: GGT-6605). The "has many" field input component for MUI is not done, so I can temporarily exclude the `rounds` field first which is a `hasMany` type.

packages/react/src/metadata.tsx
```diff
@@ -32,11 +32,13 @@ const FieldMetadataFragment = graphql(/* GraphQL */ `
       ... on GadgetHasManyConfig {
         relatedModel {
           apiIdentifier
+          namespace
```
infiton, line 35: what happens if you issue this query against a pre 1.1 app?

alanko0511: It returns `[]`

packages/react/cypress/component/auto/form/AutoForm.cy.tsx
alanko0511: I just added a test on the gadget side to prove that
https://github.com/gadget-inc/gadget/pull/12053

infiton (approved):

PR #428, opened 2024-06-22
Added SubmitResultBanner component to <AutoForm>

MillanWangGadget:
- Added SubmitResultBanner
  - Oneliner contextual closable banner used to show success message, submission errors, and actionMetadata errors
  - The individual banners can be completely overridden with all the corresponding properties of the base components
- Unified AutoForm components to use the same wrappers with/without passed in children
- Allowed for MUIAutoForm to be passed in children
- Removed onClick handlers from AutoSubmit buttons since they did not actually get called as a result of having a submit property
 
CleanShot 2024-06-25 at 13 29 07


packages/react/src/auto/polaris/submit/PolarisAutoSubmit.tsx
```diff
@@ -0,0 +1,9 @@
+import { Button } from "@shopify/polaris";
+import React, { ReactNode } from "react";
+
+export const PolarisAutoSubmit = (props: { children?: ReactNode }) => {
+  // const { submit } = useAutoFormMetadata();
+  // The submit function from useAutoFormMetadata can be used instead of the `submit` property on Button
```
alanko0511, line 6: Maybe move these two comments above and make it as a comment block? Like this:
```ts
/**
 * (the comments)
 */
export const PolarisAutoSubmit = (props: { children?: ReactNode }) => {
```

packages/react/src/auto/polaris/submit/PolarisSubmitResultBanner.tsx
```diff
@@ -0,0 +1,46 @@
+import { Banner, BannerProps } from "@shopify/polaris";
+import React, { useEffect, useState } from "react";
+import { useAutoFormMetadata } from "../../AutoFormContext.js";
+
+export const PolarisSubmitResultBanner = (props: { successBannerProps?: BannerProps; errorBannerProps?: BannerProps }) => {
```
alanko0511, line 5: What do you think about `propsOnError` and `propsOnSuccess`? Or even like this:
```
props: {
  bannerProps: (isSuccess: boolean): BannerProps
}
```
Then when users want to modify it based on the state, they can do something like this:
```jsx
<PolarisSubmitResultBanner
  bannerProps={(isSuccess) => {
    if(isSuccess) {
      ...
    } else {
      ...
    }
  }}
/>
```

alanko0511 (approved):
If possible could you add a screenshot in the PR description to show what the banner looks like?

packages/react/src/auto/polaris/submit/PolarisAutoSubmit.tsx
MillanWangGadget: Updated comment position

packages/react/src/auto/polaris/submit/PolarisSubmitResultBanner.tsx
MillanWangGadget: In my opinion, I'd prefer to keep the props as separate objects instead of a boolean controlled one

I feel like a user looking through the suggested props will see something like `bannerProps` and be confused when it is a function instead of an object

PR #430, opened 2024-06-24
Fix namespaced model not working for findBy... operation

alanko0511:
This PR fixes an issue of namespaced model not working when running `findBy...` operations like `findById` or `findBy{unique_field}`.

## PR Checklist


MillanWangGadget (approved):
TESTING
- 🍏 Using the gitpkg'd `@gadgetinc/api-client-core`, `api.name.spaced.model.findByUniqueField(...)` correctly finds the record
- 🍏 No changes in functionality for non-namespaced models

Test makes sense 👍

infiton (approved):

PR #432, opened 2024-06-25
Include record ID when submitting form

alanko0511:
This PR fixes the current issue of not being able to submit a form that runs an "update"/"delete" action by including the record ID when submitting a form that accepts ID as a parameter.

It also includes some new functions and hooks:
- Shared `useAutoForm` for `AutoForm` components.
- Test support functions for generating a GadgetMeta and record queries.

## PR Checklist


packages/react/src/auto/AutoForm.ts
```diff
@@ -73,6 +74,56 @@ export const useFormFields = (
...
+  // setup the form state for the action
+  const {
+    submit,
+    error: formError,
+    formState: { isSubmitSuccessful, isLoading },
+    originalFormMethods,
+  } = useActionForm(action, {
+    defaultValues: {
+      [action.modelApiIdentifier]: record,
+      id: "0", // The ID value will be replaced when sending the form to use the record found by `findBy`
```
alanko0511, line 111: The `id` is marked as a required field in the validator, so I have to add a dummy id value to it.

Inside the `useActionForm`, we will replace the value from the record found when we submit the form:
https://github.com/gadget-inc/js-clients/blob/7acce09540392a06100085a2054813960dd95822/packages/react/src/useActionForm.ts#L186-L188

MillanWangGadget (approved):
MANUAL TESTING
- Regular models
  - 🍏 Update actions are populated with record values and update the record on submit
  - 🍏 Delete action show custom params and properly delete records
  - 🍏 Custom actinos show custom params and properly run
- Namespaced models
  - 🍏 With the other fix to `useActionForm`, the related update/delete/custom actions work as expected 

Code changes look reasonable. Nice new hook 👍 
Let's handle the namespace thing separately

PR #433, opened 2024-06-25
Update Shopify Polaris Icon dependency version

MillanWangGadget:
- **UPDATES**
  - The Shopify Polaris icons version used here was incompatible with the default Shopify Polaris icons version used in Gadget apps. This was due to a breaking change on their end caused by renaming icons
  - References to the old icon names have been updated to the new replacement icons recommended by the Shopify docs
  - Now, in Gadget apps, there is no need to change the default shopify dependencies in the Shopify template when adding the dependencies to use autoforms


alanko0511 (approved):
:shipit: 👌

PR #434, opened 2024-06-25
Fix findBy namespaced model field in useActionForm hook

alanko0511:
This PR fixes an issue fetching a record from a namespaced model when using a `useActionForm` React hook with a `findBy`.

The bug is similar to https://github.com/gadget-inc/js-clients/pull/430, where it only gets the model manager by API identifier but does not include the namespaces.

We found this while working on an upcoming feature, but because this bug can impact all v1.1 apps, we should fix it first.

## PR Checklist


package.json
```diff
@@ -22,12 +22,12 @@
   },
   "devDependencies": {
     "@arethetypeswrong/cli": "^0.15.1",
+    "@gadget-client/app-with-file-fields": "^1.5.0",
     "@gadget-client/app-with-no-user-model": "^1.10.0",
     "@gadget-client/bulk-actions-test": "^1.113.0",
     "@gadget-client/full-auth": "^1.9.0",
-    "@gadget-client/kitchen-sink": "1.4.0-development.184",
+    "@gadget-client/kitchen-sink": "1.5.0-development.200",
```
alanko0511, line 29: I updated this real Gadget app to fix some of the legacy model namespace issues, so I updated the package version.

MillanWangGadget (approved):
Looks good! Tested on the other PR for getting the IDs into the AutoForm

pistachiobaby (approved):

PR #435, opened 2024-06-25
only send query variables if they are defined

infiton:
Working on adding limit filtering on the session model, but `deleteMany` was broken because the session model does not take a `search` argument. I've made a change here that when generating the queries for internal deletes we don't add a variable to the query unless we are actually passing a variable.

I don't think always adding these variables in was intentional cc @airhorns 

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -89,10 +89,10 @@ export const internalFindManyQuery = (apiIdentifier: string, namespace: string[]
         [`list${capitalizedApiIdentifier}`]: Call(
           {
             ...defaultVariables,
-            after: Var({ value: options?.after, type: "String" }),
-            before: Var({ value: options?.before, type: "String" }),
-            first: Var({ value: options?.first, type: "Int" }),
-            last: Var({ value: options?.last, type: "Int" }),
+            after: options?.after ? Var({ value: options.after, type: "String" }) : undefined,
+            before: options?.before ? Var({ value: options?.before, type: "String" }) : undefined,
+            first: options?.first ? Var({ value: options?.first, type: "Int" }) : undefined,
+            last: options?.last ? Var({ value: options?.last, type: "Int" }) : undefined,
           },
           {
             pageInfo: { hasNextPage: true, hasPreviousPage: true, startCursor: true, endCursor: true },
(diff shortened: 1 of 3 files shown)


packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -89,10 +89,10 @@ export const internalFindManyQuery = (apiIdentifier: string, namespace: string[]
...
           {
             ...defaultVariables,
-            after: Var({ value: options?.after, type: "String" }),
-            before: Var({ value: options?.before, type: "String" }),
-            first: Var({ value: options?.first, type: "Int" }),
-            last: Var({ value: options?.last, type: "Int" }),
+            after: options?.after ? Var({ value: options.after, type: "String" }) : undefined,
+            before: options?.before ? Var({ value: options?.before, type: "String" }) : undefined,
+            first: options?.first ? Var({ value: options?.first, type: "Int" }) : undefined,
+            last: options?.last ? Var({ value: options?.last, type: "Int" }) : undefined,
```
thegedge, line 95: 👍 this works, but might be best to fix it at the source.

`tiny-graphql-query-compiler` should be filtering out variables with `undefined` values here https://github.com/gadget-inc/js-clients/blob/24540054e586103a3a00d4d4d30383c44ae02a03/packages/tiny-graphql-query-compiler/src/index.ts#L58-L60

especially since we don't provide a variable value here:

https://github.com/gadget-inc/js-clients/blob/24540054e586103a3a00d4d4d30383c44ae02a03/packages/tiny-graphql-query-compiler/src/index.ts#L127-L129

What do you think?

airhorns (approved):
Agreed with gedge but otherwise LGTM. Generally speaking its totally fine to have a variable specified in the query but not sent, but this seems like an easy way to overcome the fact that some models sneakily don't have all the variables.

packages/api-client-core/src/InternalModelManager.ts
infiton: So it looks to me like we can't fix this here simply since we sometimes pass variables in here that don't have a value at compile time but are passed in at run time (i.e. we can't filter in this line by value.value is undefined)

thegedge: Ah, that makes sense. I was thinking everything was known at compile time 👍

thegedge (approved):

PR #437, opened 2024-06-25
Fix incorrect import path in PolarisAutoForm storybook file

alanko0511:
Fix the import path that is no longer correct after moving the files inside the folder.

## PR Checklist


MillanWangGadget (approved):
Storybook works good. My mistake for not checking this

PR #440, opened 2024-06-26
Implement hidden input for auto component

alanko0511:
This PR introduces `AutoHiddenInput` for the auto component. Underneath is a `<input type="hidden">` for users to debug the value in a browser development tool.

CleanShot 2024-06-26 at 18 07 19@2x

## PR Checklist

### DIFF
--- a/packages/react/spec/auto/inputs/PolarisAutoHiddenInput.spec.tsx
+++ b/packages/react/spec/auto/inputs/PolarisAutoHiddenInput.spec.tsx
@@ -1,0 +1,140 @@
+import { AppProvider } from "@shopify/polaris";
+import translations from "@shopify/polaris/locales/en.json";
+import { act, render } from "@testing-library/react";
+import { userEvent } from "@testing-library/user-event";
+import type { ReactNode } from "react";
+import React from "react";
+import { PolarisAutoForm } from "../../../src/auto/polaris/PolarisAutoForm.js";
+import { PolarisAutoHiddenInput } from "../../../src/auto/polaris/inputs/PolarisAutoHiddenInput.js";
+import { PolarisAutoSubmit } from "../../../src/auto/polaris/submit/PolarisAutoSubmit.js";
+import { testApi as api } from "../../apis.js";
+import { MockClientProvider, mockUrqlClient } from "../../testWrappers.js";
+import { mockWidgetFindBy } from "../support/helper.js";
+import { getWidgetModelMetadata } from "../support/widgetModel.js";
+
+const PolarisMockedProviders = (props: { children: ReactNode }) => {
+  return (
+    <MockClientProvider api={api}>
+      <AppProvider i18n={translations}>{props.children}</AppProvider>
+    </MockClientProvider>
+  );
+};
+
+describe("PolarisAutoHiddenInput", () => {
+  it("should set the value", async () => {
+    const user = userEvent.setup();
+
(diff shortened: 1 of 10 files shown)


packages/react/spec/auto/support/helper.ts
MillanWangGadget, line 1: NIT - Perhaps name this file `mockWidgetFindBy` since it's the only thing in here

packages/react/spec/auto/inputs/PolarisAutoHiddenInput.spec.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const { getByRole } = render(
+      <PolarisAutoForm action={api.widget.create} findBy="42">
+        <PolarisAutoHiddenInput field="name" value="Bob" />
+        <PolarisAutoHiddenInput field="inventoryCount" value={9999} />
+        <PolarisAutoSubmit />
+      </PolarisAutoForm>,
+      { wrapper: PolarisMockedProviders }
+    );
+
+    mockUpdateWidgetFindBy();
```
MillanWangGadget, line 63: - `api.widget.create` and `mockUpdate` are used here. Shouldn't they match?
- Can you add a check that ensures that the GQL mutation includes the same ID that was given in the `findBy`
- Add a check/comment here that mentions that the loaded in value from the findby response is `Alice`

packages/react/spec/auto/inputs/PolarisAutoHiddenInput.spec.tsx
```diff
@@ -0,0 +1,110 @@
...
+    });
+
+    const mutation = mockUrqlClient.executeMutation.mock.calls[0][0];
+    const mutationName = mutation.query.definitions[0].name.value;
+    const variables = mutation.variables.widget;
+
+    expect(mutationName).toEqual("createWidget");
+    expect(variables.inventoryCount).toEqual(9999);
+    expect(variables.name).toEqual("Bob");
+  });
```
MillanWangGadget, line 76: Consider a test where you have this
```
        <PolarisAutoHiddenInput field="name" value="ignored because first" />
        <PolarisAutoHiddenInput field="name" value="Bob" />
```
That shows that when there are multiple hidden inputs on the same field with different values, the value that gets sent is the one that is lowest in the DOM

packages/react/src/auto/hooks/useHiddenInput.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const useHiddenInput = (props: { field: string; value: any; control?: Control<any> }) => {
+  const { field, value, control } = props;
+  const { path, metadata } = useFieldMetadata(field);
+  const form = useFormContext();
+
+  const { field: fieldProperties } = useController({
+    name: path,
+    rules: { required: metadata.requiredArgumentForInput },
+    control,
+  });
```
MillanWangGadget, line 15: Is this hook needed if it's values are never used? Tests still pass if it's removed.

It would provide `name,onBlur,onChange,ref,value,disabled`, but I don't think any of those would be relevant for a hidden input

MillanWangGadget (approved):
MANUAL TESTING
- 🍏 Hidden inputs work to force certain submission values
- 🍏 It throws an exception when given 
  - fields that don't exist in the metadata
  - id field

Nice test helpers 👍
Some code comments but good to go otherwise 👍

PR #441, opened 2024-06-27
Exclude non-dirty fields when submitting an AutoForm

alanko0511:
This PR excludes model fields that are not "dirty" when submitting the form. In the react-hook-form standard, a "dirty field" is a field the user has modified.

Previously, when the field was not dirty, its value was `null`. However, there is a semantic difference between `undefined` and `null`. If the field has a default value set in the Gadget editor and is not required, the backend cannot apply the default value to it.

By only including the dirty fields in the action parameters, excluded fields are considered `undefined`, and the backend can apply the default value to them.

## PR Checklist

### DIFF
--- a/packages/react/src/auto/hooks/useHiddenInput.tsx
+++ b/packages/react/src/auto/hooks/useHiddenInput.tsx
@@ -5,11 +5,13 @@ import { useFieldMetadata } from "./useFieldMetadata.js";
 export const useHiddenInput = (props: { field: string; value: any }) => {
   const { field, value } = props;
   const { path, metadata } = useFieldMetadata(field);
-  const form = useFormContext();
+  const { setValue, formState } = useFormContext();
 
   useEffect(() => {
-    form.setValue(path, value);
-  }, [form, path, value]);
+    setValue(path, value, {
+      shouldDirty: true,
+    });
+  }, [formState.defaultValues, path, setValue, value]);
 
   return {
     value,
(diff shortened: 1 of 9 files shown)


packages/react/src/auto/hooks/useHiddenInput.tsx
```diff
@@ -5,11 +5,13 @@ import { useFieldMetadata } from "./useFieldMetadata.js";
...
-  const form = useFormContext();
+  const { setValue, formState } = useFormContext();
 
   useEffect(() => {
-    form.setValue(path, value);
-  }, [form, path, value]);
+    setValue(path, value, {
+      shouldDirty: true,
+    });
+  }, [formState.defaultValues, path, setValue, value]);
```
alanko0511, line 14: Manually setting a value with `setValue` isn't considered a dirty field, so I have to explicitly set this as a dirty field by setting `shouldDirty: true`.

MillanWangGadget (approved):
MANUAL TESTING - In a real Gadget app
- 🍏 Hidden inputs fields work as expected and properly submit the their `value`
- 🍏 Setting values in the inputs still correctly submits them. Automatic `dirty` setting in existing fields
- 🍏 Untouched form fields lead to the defaut field in the Gadget editor being used
- 🍏 Setting and then clearing a form field leads to an empty string being set in the database (appears as null in the Gadget data viewer though)

Thanks for bringing in `HasOne` 👍 
Good to ship 👍

PR #442, opened 2024-06-27
AutoRelationshipField => Part 1 - BelongsTo fields

MillanWangGadget:
- **UPDATE**
  - Added AutoRelationshipInput component to handle BelongsTo input fields, replacing the previous implementation of the belongsTo input component 
  - Added search and pagination functionality in a new hooks for use in the AutoRelationship field component
    - Eventual plan is to expand the same functionality to HasOne and HasMany. Those will be updated in a follow up update

## PR Checklist


packages/react/src/utils.ts
```diff
@@ -402,6 +402,15 @@ export const set = (obj: any, path: string, value: any) => {
   }, obj);
 };
 
+/**
+ * Based on Lodash uniqBy to ensure all array elements are unique by a property
+ * https://youmightnotneed.com/lodash
```
airhorns, line 407: 🤘

packages/react/src/auto/polaris/inputs/PolarisAutoRelationshipInput.tsx
```diff
@@ -0,0 +1,92 @@
+import { Combobox, Icon, Listbox, Tag, Text } from "@shopify/polaris";
+import { SearchIcon } from "@shopify/polaris-icons";
+import React from "react";
+import { Control } from "react-hook-form";
+import { ErrorWrapper } from "../../../utils.js";
+import { numberOfRecordsToLoad, useRelationshipInputController } from "../../hooks/useRelationshipInputController.js";
+
+export const PolarisAutoRelationshipInput = (props: {
```
airhorns, line 8: I get why this is one component we intend to use for all the relationships internally -- its the same component in multi mode or not multi mode. But, I think it'd be a clearer interface for users of these components to have one per relationship type instead of one for all relationships, since I think we generally represent the relationships as super different externally. Can we name the exports that they use and that we document things like `PolarisAutoBelongsToInput` and use this sucker under the hood?

airhorns (approved):

packages/react/spec/auto/inputs/PolarisAutoRelationshipInputBelongsTo.spec.tsx
```diff
@@ -0,0 +1,253 @@
...
+const mockSectionQueryResponse = (edges?: any[]) => {
+  mockUrqlClient.executeQuery.pushResponse("sections", getSectionsQueryResponse(edges));
+};
+
+const metadata: ActionMetadata = {
+  name: "Widget",
+  action: {
+    name: "Create",
+    apiIdentifier: "create",
+    inputFields: [
```
alanko0511, line 166: Is it possible to use the action metadata from `packages/react/spec/auto/support/widgetModel.ts` ?

alanko0511 (approved):
Do we need to add the relationship inputs into the `polaris/index.ts` & `mui/index.ts` files?

But other than that :shipit: 👌

packages/react/src/auto/polaris/inputs/PolarisAutoRelationshipInput.tsx
MillanWangGadget: Updated such that there are individual fieldType based exports in the `index` file that point to the same `AutoRelationshipInput` component

```jsx
export const PolarisAutoBelongsToInput = (props: AutoRelationshipInputProps) => {
  return <PolarisAutoRelationshipInput {...props} />;
};
```

MillanWangGadget:
> Do we need to add the relationship inputs into the `polaris/index.ts` & `mui/index.ts` files?

Updated to add the corresponding components to bot index files

packages/react/spec/auto/inputs/PolarisAutoRelationshipInputBelongsTo.spec.tsx
MillanWangGadget: Good call. Updated to use the field metadata result from `widgetModel.ts`

PR #444, opened 2024-07-02
Custom Time Picker

gabeb03:
*Background:* [Polaris does not offer a time picker](https://github.com/Shopify/polaris/issues/11223), so this PR adds a custom time picker, pieced together using Polaris components. This PR also XXX

*Testing instructions:* Data models that have a date/time field (and "Include Time" is selected) should have a corresponding autocomponent with a date & time picker. Set a date & time using the autocomponent & make sure the correct value is reflected in the data store. 

## PR Checklist


packages/react/spec/auto/inputs/PolarisDateTimePicker.spec.tsx
```diff
@@ -1,32 +1,64 @@
...
-    expect(onChangeSpy).toHaveBeenCalledWith(new Date("2021-03-05T16:00:00.000Z"));
+    const formSubmitRequest = mockUrqlClient.executeMutation.mock.calls[0][0];
+    const startsAt = formSubmitRequest.variables.widget.startsAt;
+    expect(startsAt).toEqual(today);
+    expect(onChangeSpy).toHaveBeenCalledWith(today);
   });
 });
+
+const today = new Date();
+today.setHours(11);
```
gabeb03, line 48: The original date that is submitted is not converted to UTC I believe this is expected behavior: @MillanWangGadget  or @airhorns - could you confirm?

packages/react/cypress/component/auto/form/PolarisDateTimePicker.cy.tsx
```diff
@@ -56,21 +78,37 @@ describe("PolarisDateTimePicker", () => {
...
+        <PolarisAutoForm action={api.widget.create}>
+          <PolarisDateTimePicker id="test" includeTime onChange={onChangeSpy} field="startsAt" />
+        </PolarisAutoForm>,
+        PolarisWrapper
+      );
       cy.get("#test-date").should("have.value", "");
       cy.get("#test-time").should("have.value", "");
     });
 
+    // TODO: Fix
```
MillanWangGadget, line 91: Is this still relevant?

packages/react/src/auto/mui/inputs/MUIDateTimePicker.tsx
MillanWangGadget, line 1: Add this to `packages/react/src/auto/mui/inputs/MUIAutoInput.tsx` in the same way that it done in the Polaris counterpart

packages/react/src/auto/mui/inputs/MUIDateTimePicker.tsx
```diff
@@ -0,0 +1,18 @@
+import { Box } from "@mui/material";
+import { DatePicker } from "@mui/x-date-pickers";
+import { TimePicker } from "@mui/x-date-pickers/TimePicker";
+import React from "react";
+import { useFieldMetadata } from "../../hooks/useFieldMetadata.js";
+
+export const MUIDateTimePicker = (props: { field: string; value?: Date; onChange: (value: Date) => void; error?: string }) => {
```
MillanWangGadget, line 7: make the onChange optional

packages/react/src/auto/mui/inputs/MUIDateTimePicker.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const { metadata } = useFieldMetadata(props.field);
+
+  return (
+    <Box sx={{ display: "flex" }}>
+      <DatePicker label={metadata.name} />
+      <TimePicker />
```
MillanWangGadget, line 13: These 2 would not be connected to the submit button. They need a `useController` to get the corresponding `value` and `onChange` values that are connected to the form context

packages/react/src/auto/mui/inputs/MUIDateTimePicker.tsx
MillanWangGadget: Also add it to `packages/react/src/auto/mui/index.ts`

packages/react/spec/auto/inputs/PolarisDateTimePicker.spec.tsx
```diff
@@ -16,17 +16,49 @@ describe("PolarisDateTimePicker", () => {
...
+    expect(startsAt).toEqual(today);
+    expect(onChangeSpy).toHaveBeenCalledWith(today);
   });
 });
+
+const today = new Date();
+today.setHours(15);
+today.setMinutes(0);
+today.setSeconds(0);
+today.setMilliseconds(0);
```
MillanWangGadget, line 51: This area ends up getting called before the call to 
```
jest.spyOn(Intl.DateTimeFormat.prototype, "resolvedOptions").mockReturnValue({
```
I don't that makes a difference now considering CI is green, but could you move this area below that statement so that we don't need to think about this in the event we need to revisit this area?

packages/react/src/auto/polaris/PolarisDateTimePicker.tsx
MillanWangGadget, line 1: Add to the polaris index file

packages/react/src/auto/polaris/PolarisDateTimePicker.tsx
```diff
@@ -48,58 +14,120 @@ const copyTime = (to: Date, input: Date) => {
...
-  const localTime = value ? utcToZonedTime(value, localTz) : undefined;
+  const localTime = value ? utcToZonedTime(value, localTz) : fieldProps.value;
   const [{ month, year }, setDate] = useState({ month: (localTime ?? today).getMonth(), year: (localTime ?? today).getFullYear() });
   const [datePopoverActive, setDatePopoverActive] = useState(false);
+  const [timeString, setTimeString] = useState(localTime ? format(localTime, "K:m aa") : "");
+  const [popoverActive, setPopoverActive] = useState(false);
+  const [timeParseError, setTimeParseError] = useState(false);
+  const [hourSelected, setHourSelected] = useState(hours[0]);
+  const [minSelected, setMinSelected] = useState(mins[0]);
+  const [ampmSelected, setAmpmSelected] = useState(ampm[0]);
```
MillanWangGadget, line 69: Can these be changed such that they are parsed from `timeString`? That would prevent the possibility of these values disagreeing with `timeString` which is the real source of truth.  

It could be nice to have a single object represent the whole dateTime to prevent string operations too. Maybe something like this

```js
const  [dateTimeState, setDateTimeState] = useState({
minute,
hour,
ampm,
day,
month,
year

})

const setHours = (hour:number) => {
   setDateTimeState({...dateTimeState, hour})
}
```

packages/react/src/auto/polaris/PolarisDateTimePicker.tsx
```diff
@@ -48,58 +14,120 @@ const copyTime = (to: Date, input: Date) => {
...
-  useEffect(() => {
-    if (value) {
-      setTimeString(getTimeString(value));
+  const toggleDatePopoverActive = useCallback(() => setDatePopoverActive((active) => !active), []);
+  const handleMonthChange = useCallback((month: number, year: number) => {
+    setDate({ month, year });
+  }, []);
+
+  const onTimeStringChange = (value: string) => {
+    console.log("** value: ", value);
```
MillanWangGadget, line 90: leftover log

packages/react/src/auto/polaris/PolarisDateTimePicker.tsx
```diff
@@ -48,58 +14,120 @@ const copyTime = (to: Date, input: Date) => {
...
+
+    parsedDate.setHours(hours, minutes, 0, 0);
+    const date = localTime ? new Date(localTime) : new Date();
+
+    copyTime(date, parsedDate);
+
+    onChange?.(zonedTimeToUtc(date, localTz));
+    fieldProps.onChange(zonedTimeToUtc(date, localTz));
+
+    console.log("** formatted time: ", `${hoursStr}:${paddedMins} ${newAmpm}`);
```
MillanWangGadget, line 127: leftover log

packages/react/src/auto/polaris/PolarisDateTimePicker.tsx
```diff
@@ -48,58 +14,120 @@ const copyTime = (to: Date, input: Date) => {
...
+  });
+
   const { onChange, value } = props;
   const localTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
-  const localTime = value ? utcToZonedTime(value, localTz) : undefined;
+  const localTime = value ? utcToZonedTime(value, localTz) : fieldProps.value;
   const [{ month, year }, setDate] = useState({ month: (localTime ?? today).getMonth(), year: (localTime ?? today).getFullYear() });
   const [datePopoverActive, setDatePopoverActive] = useState(false);
+  const [timeString, setTimeString] = useState(localTime ? format(localTime, "K:m aa") : "");
+  const [popoverActive, setPopoverActive] = useState(false);
```
MillanWangGadget, line 65: rename to timePopoverActive to be more easily differentiable from the date counterpart

packages/react/src/auto/polaris/PolarisDateTimePicker.tsx
```diff
@@ -130,17 +158,46 @@ export const PolarisDateTimePicker = (props: {
...
-          onChange={onTimeStringChange}
-          {...timeInputFocusProps}
-          error={!parseSuccess && !isTimeFocused && `Invalid time, please enter a value like "11:00 AM"`}
-        />
+      {(props.includeTime ?? (config as any).includeTime) && (
+        <Popover
+          active={popoverActive}
+          onClose={() => setPopoverActive(false)}
+          activator={
+            <TextField
```
MillanWangGadget, line 166: Right now, a user has no way of passing in their own custom props to the Date or time inputs. Perhaps some props called `datePickerProps?: Partial<DatePickerProps> ` and `timePickerProps: Partial<TextFieldProps>` could be added so that this would be possible. 


packages/react/src/auto/polaris/submit/PolarisSubmitResultBanner.tsx
^ An example of something similar

packages/react/src/auto/polaris/PolarisDateTimePicker.tsx
```diff
@@ -130,17 +158,46 @@ export const PolarisDateTimePicker = (props: {
...
-          id={props.id ? `${props.id}-time` : undefined}
-          label={props.timeLabel ?? "Time"}
-          prefix={<Icon source={ClockIcon} />}
-          autoComplete="off"
-          value={timeString}
-          onChange={onTimeStringChange}
-          {...timeInputFocusProps}
-          error={!parseSuccess && !isTimeFocused && `Invalid time, please enter a value like "11:00 AM"`}
-        />
+      {(props.includeTime ?? (config as any).includeTime) && (
```
MillanWangGadget, line 161: The time section is fairly large and separate from the date component. Consider moving it to a separate non-exported component

packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
```diff
@@ -0,0 +1,117 @@
...
+            allowRange={false}
+            onChange={onDateChange}
+            onMonthChange={handleMonthChange}
+            selected={props.value}
+            {...props.datePickerProps}
+          />
+        </div>
+      </Popover>
+      {(props.includeTime ?? (config as any).includeTime) && (
+        <PolarisTimePicker
```
MillanWangGadget, line 104: add `{...props.timePickerProps}`

packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
```diff
@@ -0,0 +1,117 @@
...
+  );
+
+  const toggleDatePopoverActive = useCallback(() => setDatePopoverActive((active) => !active), []);
+  const handleMonthChange = useCallback((month: number, year: number) => {
+    setDateTimeState((prevState) => ({
+      ...prevState,
+      month,
+      year,
+    }));
+    // setDateTimeState({ ...dateTimeState, month, year });
```
MillanWangGadget, line 70: Leftover comment

packages/react/src/auto/polaris/inputs/PolarisTimePicker.tsx
```diff
@@ -0,0 +1,178 @@
...
+  minute: string;
+  ampm: string;
+}
+
+const PolarisTimePicker = (props: {
+  field: string;
+  dateTimeState: DateTimeState;
+  setDateTimeState: React.Dispatch<React.SetStateAction<DateTimeState>>;
+  onChange?: (value: Date) => void;
+  localTime?: any;
```
MillanWangGadget, line 51: This can be `: Date` if you force `fieldProps.value as Date` in the date picker

MillanWangGadget (commented):
Right now the time input is not being automatically populated with the value from the record, but the time value is. In storybook, check out AutoForm->UpdateRecord. I just updated that to have a non-null value for startsAt

packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
gabeb03: I added `{...props.timePickerProps}` to the `TextField` activator in `PolarisTimePicker`

packages/react/src/auto/mui/inputs/MUIDateTimePicker.tsx
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <Box sx={{ display: "flex" }}>
+      <DatePicker
+        label={metadata.name}
+        onChange={(newValue: string | number | Date | null) => {
+          props.onChange?.(zonedTimeToUtc(new Date(newValue ?? ""), localTz));
+          fieldProps.onChange(zonedTimeToUtc(new Date(newValue ?? ""), localTz));
+        }}
+      />
+      {(config as any).includeTime && (
```
MillanWangGadget, line 25: `(config as GadgetDateTimeConfig)`

packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+            year={getDateTimeObjectFromDate(zonedTimeToUtc(new Date(fieldProps.value), localTz)).year}
+            allowRange={false}
+            onChange={onDateChange}
+            onMonthChange={handleMonthChange}
+            selected={props.value}
+            {...props.datePickerProps}
+          />
+        </div>
+      </Popover>
+      {(props.includeTime ?? (config as any).includeTime) && (
```
MillanWangGadget, line 119: `(config as GadgetDateTimeConfig)`

packages/react/src/auto/polaris/inputs/PolarisTimePicker.tsx
```diff
@@ -0,0 +1,205 @@
...
+            onFocus={() => setTimePopoverActive(true)}
+            error={timeParseError && `Invalid time format`}
+            {...props.timePickerProps}
+          />
+        }
+      >
+        {!props.hideTimePopover && (
+          <div style={{ overflow: "hidden", padding: "15px 8px" }}>
+            <div style={{ display: "flex" }}>
+              <Scrollable style={{ overflowY: "scroll", height: "250px" }}>
```
MillanWangGadget, line 145: This is a Scrollable while the others are all div

Consider making a component for a single scroller and then do something like this
```jsx
[hoursProps, minutesProps, amPmProps].map((props)=> <ScrollableTimeOptions {...props}/>)
```
That way, we can avoid these `changed one but should have changed all of them` moments

MillanWangGadget (approved):
MANUAL TESTING
- 🍏 The DB record value for the findBy is properly retrieved and set in the component
- 🍏 Updates and creates get the value set in the time component

Some minor code comments but good to go otherwise 👍

packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+  date.setSeconds(0);
+  date.setMilliseconds(0);
+  return date;
+};
+
+export const PolarisDateTimePicker = (props: {
+  field: string;
+  id?: string;
+  value?: Date;
+  timeLabel?: string;
```
MillanWangGadget, line 46: I don't think this is needed since it can already be given from `props.timePickerProps?.label`

alanko0511 (approved):
:shipit: 👌

PR #445, opened 2024-07-02
Merge in changes from main branch

MillanWangGadget:
- There have been many changes on main recently. They are needed here in the feature branch
- The only change here on this branch is that the package version for the reference to the real js-clients app has been updated so that the Cypress tests can continue to work as expected


MillanWangGadget:
Merging since everything on main should be safe already

PR #446, opened 2024-07-02
Updated order of field appearance to match the `include` property

MillanWangGadget:
- **UPDATES**
  - Updated order of field appearance to match the `include` property
  - Removed the `select` and `fields` properties, which seemed to duplicate the behaviour of the `include` and `exclude` properties

## PR Checklist

### DIFF
--- a/packages/react/spec/auto/hooks/useFormFields.spec.tsx
+++ b/packages/react/spec/auto/hooks/useFormFields.spec.tsx
@@ -1,0 +1,112 @@
+import { renderHook } from "@testing-library/react";
+import { useFormFields } from "../../../src/auto/AutoForm.js";
+import { useAutoFormMetadata } from "../../../src/auto/AutoFormContext.js";
+import type { ActionMetadata } from "../../../src/metadata.js";
+import { MockForm } from "../MockForm.js";
+
+describe("useFormFields hook", () => {
+  const getUseFormFieldsResult = (options: { include?: string[]; exclude?: string[] }) => {
+    const { result } = renderHook(
+      () => {
+        const { metadata: autoFormMetadata } = useAutoFormMetadata();
+        return useFormFields(autoFormMetadata, options);
+      },
+      {
+        wrapper: MockForm({ submit: jest.fn(), metadata }),
+      }
+    );
+
+    return result.current;
+  };
+
+  test("Returns the fields in the default order of the action metadata when no include/exclude options are used", () => {
+    const fields = getUseFormFieldsResult({});
+
+    expect(fields.map((field) => field.metadata.apiIdentifier)).toMatchInlineSnapshot(`
+        [
+          "stringField1",
+          "stringField2",
+          "stringField3",
+        ]
+      `);
+  });
+
+  test("Returns the fields in the order of the include options list", () => {
+    const include = ["stringField2", "stringField1", "stringField3"];
+
+    const fields = getUseFormFieldsResult({ include });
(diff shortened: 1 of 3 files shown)


packages/react/src/metadata.tsx
```diff
@@ -206,27 +206,27 @@ export const useActionMetadata = (actionFunction: ActionFunction<any, any, any,
...
   fields: FieldMetadata[] | undefined,
-  options?: { include?: string[]; exclude?: string[]; fields?: string[]; select?: Record<string, any> }
+  options?: { include?: string[]; exclude?: string[] }
 ): FieldMetadata[] => {
   if (!fields) {
     return [];
   }
+
   let subset = fields;
-  if (options?.fields) {
```
MillanWangGadget, line 215: Seems to be the same as `include` but throwing the exception becomes problematic when the action fields metadata is loading

packages/react/src/metadata.tsx
```diff
@@ -206,27 +206,27 @@ export const useActionMetadata = (actionFunction: ActionFunction<any, any, any,
...
+
   if (options?.include) {
+    // When including fields, the order will match the order of the `include` array
+    subset = [];
     const includes = new Set(options.include);
-    subset = subset.filter((field) => includes.has(field.apiIdentifier));
-  }
-  if (options?.select) {
-    const includes = new Set(Object.keys(options.select).filter((key) => options.select![key]));
-    subset = subset.filter((field) => includes.has(field.apiIdentifier));
```
MillanWangGadget, line 228: select seems to be very similar to include where you would do something like 
```jsx
<Autoform ...
  select={{
    fieldToInclude: true,
    ignoredField: false
  }}
/>
```

packages/react/spec/auto/hooks/useFormFields.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const fields = getUseFormFieldsResult({ include });
+    expect(fields.map((field) => field.metadata.apiIdentifier)).toEqual(include);
+  });
+
+  test("Excludes fields in the exclude list even if they are explicitly included", () => {
+    const fields = getUseFormFieldsResult({
+      include: ["stringField2", "stringField1", "stringField3"],
+      exclude: ["stringField2", "stringField3"],
+    });
```
airhorns, line 45: I think we should throw here -- the dev has given us conflicting information, I'd rather just be explicit and make them fix it?

packages/react/spec/auto/hooks/useFormFields.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+    expect(fields.map((field) => field.metadata.apiIdentifier)).toEqual(["stringField1"]);
+  });
+});
```
airhorns, line 48: Is there a default ordering we could add a regression test for?

airhorns (approved):
LGTM after comments

alanko0511 (approved):
:shipit: 👌

packages/react/spec/auto/hooks/useFormFields.spec.tsx
MillanWangGadget: The default ordering is in the same order as the action metadata response. That's the first test in the suite :)

packages/react/spec/auto/hooks/useFormFields.spec.tsx
MillanWangGadget: Updated to have an error if include/exclude have any overlap

PR #447, opened 2024-07-02
Implement json input for auto component without advanced editor

alanko0511:
This PR refactors the existing JSON input to use hooks from the closed https://github.com/gadget-inc/js-clients/pull/439 PR and fixes the issue of not setting the fetched JSON value to the input in the existing JSON input. Note that this input doesn't have advanced features like syntax highlighting and auto-indentation right now. Ticket for syntax highlighting: GGT-6684

The original PR had an issue building the react package related to types. A similar problem was mentioned in their repo (https://github.com/FIameCaster/prism-code-editor/issues/9). However, it is difficult for us to just set the `moduleResolution` to `Bundler` in the react package's `tsconfig.json`, so we could not use that library to achieve the advanced editor result.

## PR Checklist


packages/react/src/auto/hooks/useJSONInputController.tsx
```diff
@@ -0,0 +1,48 @@
...
+        setError(undefined);
+        stringInputController.onChange(newJSON);
+      } catch (error: any) {
+        if (newString !== "") {
+          setError(error);
+        } else {
+          setError(undefined);
+        }
+      }
+    },
```
alanko0511, line 31: Note: right now, setting an error with `setError` doesn't prevent the AutoForm from being sent (ticket: GGT-6663).

airhorns (approved):
This LGTM, but can you make sure there is a ticket around for making nice syntax highlighting be a thing?

alanko0511:
> but can you make sure there is a ticket around for making nice syntax highlighting be a thing?

Here is the ticket: GGT-6684

packages/react/src/auto/mui/inputs/MUIAutoJSONInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+import React from "react";
+import type { Control } from "react-hook-form";
+import { useFocus } from "../../../useFocus.js";
+import { useJSONInputController } from "../../hooks/useJSONInputController.js";
+
+export const MUIAutoJSONInput = (props: {
+  field: string; // The field API identifier
+  control?: Control<any>;
+  style?: CSSProperties;
+}) => {
```
MillanWangGadget, line 12: Could you add `& Partial<TextFieldProps>` to the props here and here pass that into the text field?

packages/react/src/auto/mui/inputs/MUIAutoJSONInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+        maxRows={4}
+        inputProps={{ style: { fontFamily: "monospace" } }}
+        {...originalController}
+        value={stringValue}
+        onChange={(event) => onStringValueChange(event.target.value)}
+        error={inErrorState}
+        helperText={inErrorState && `Invalid JSON: ${error.message}`}
+        {...focusProps}
+      />
+    </>
```
MillanWangGadget, line 31: Fragment wrapper is not needed here

packages/react/src/auto/polaris/inputs/PolarisAutoJsonInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+import React from "react";
+import type { Control } from "react-hook-form";
+import { useFocus } from "../../../useFocus.js";
+import { useJSONInputController } from "../../hooks/useJSONInputController.js";
+
+export const PolarisAutoJSONInput = (props: {
+  field: string; // The field API identifier
+  control?: Control<any>;
+  style?: CSSProperties;
+}) => {
```
MillanWangGadget, line 12: Same here as the MUI variant. Could you add `& Partial<TextFieldProps>` to the props type and pass it in?

packages/react/src/auto/polaris/inputs/PolarisAutoJsonInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <TextField
+        multiline={4}
+        monospaced
+        {...restOfOriginalController}
+        value={stringValue}
+        onChange={onStringValueChange}
+        error={!isFocused && error && `Invalid JSON: ${error.message}`}
+        {...focusProps}
+      />
+    </>
```
MillanWangGadget, line 29: <></> is not needed

MillanWangGadget (approved):
Could you add a spec test similar to the booleanInput test that verifies that the contents of the textbox are properly sent in the GQL request?

Good to go otherwise 👍

PR #448, opened 2024-07-02
Add support for an install state hint passed from the server side to the client that allows the provider to fast-path

airhorns:
In order to lower app LCP, we want to avoid having to do a whole other roundtrip to the server every time the page loads to check the installed state of a shop. We need to do this in some circumstances -- for off-platform apps, or to trigger the install of a shop the very first time a shop using managed installs is seen. So, we can't just remove this request, but, we can avoid blocking on it if the backend says its ok to. The backend already needs to check install state when rendering the page in order to inject the right shopify api key, so we can have it pass a little bit more info about the install state when it does that.

### DIFF
--- a/packages/react-shopify-app-bridge/spec/Provider.spec.tsx
+++ b/packages/react-shopify-app-bridge/spec/Provider.spec.tsx
@@ -10,18 +10,63 @@ import { AppType, Provider } from "../src/Provider.js";
 
 describe("GadgetProvider", () => {
   let mockApiClient: AnyClient;
-  const { location } = window;
   const mockNavigate = jest.fn();
   const mockOpen = jest.fn();
   const mockApiKey = "some-api-key";
   let useAppBridgeMock: jest.SpyInstance;
   let resolveIdToken: (value: string) => void;
 
-  describe.each([true, false])("as install request: %s", (isInstallRequest) => {
-    beforeAll(() => {
-      // @ts-expect-error mock
-      delete window.location;
+  beforeEach(() => {
+    // @ts-expect-error mock
+    delete window.location;
 
+    delete window.gadgetConfig;
+
+    window.open = mockOpen;
+
+    // @ts-expect-error mock
+    window.location = {
+      assign: mockNavigate,
+      origin: "https://test-app.gadget.app",
+      pathname: "/",
+      search: "",
+    };
+
+    window.shopify = {
+      // @ts-expect-error mock
+      environment: {
+        embedded: false,
+        mobile: false,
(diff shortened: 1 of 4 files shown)


packages/react-shopify-app-bridge/spec/Provider.spec.tsx
```diff
@@ -210,103 +213,159 @@ describe("GadgetProvider", () => {
...
-      expect(container.outerHTML).toMatchInlineSnapshot(`"<div><span>hello world</span></div>"`);
-      expect(mockNavigate).not.toHaveBeenCalled();
+    expect(container.outerHTML).toMatchInlineSnapshot(`"<div><span>hello world</span></div>"`);
+    expect(mockNavigate).not.toHaveBeenCalled();
 
-      window.shopify.environment.embedded = false;
+    window.shopify.environment.embedded = false;
+  });
+
+  test("can render an embedded app with an install state hint saying we're installed", async () => {
```
infiton, line 283: should we add a test that covers the case where the mutation comes back with a result that contradicts the hint?

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -152,7 +163,14 @@ const InnerGadgetProvider = memo(
...
 
-    const loading = forceRedirect || redirectToOauth || fetchingOrInstallingShop || !hasStartedFetchingOrInstallingShop.current;
+    // we're loading if:
+    //  - we need to redirect somewhere (forceRedirect || redirectToOauth)
+    //  - if we don't have an install state hint, and we're loading the shop from the backend to check current state
+    //  - if we do have an install state hint, and the hint tells us we need to re-authenticate
+    const loading =
+      forceRedirect ||
+      redirectToOauth ||
+      (hasInstallStateHint ? !isAuthenticated : fetchingOrInstallingShop || !hasStartedFetchingOrInstallingShop.current);
```
infiton, line 173: if there is a hint and the mutation comes back with `isAuthenticated = false` I think this means that we will still be in a loading state?

airhorns: Good call, I will update this to respect the value from the mutation regardless of what the hint says.

packages/react-shopify-app-bridge/spec/Provider.spec.tsx
airhorns: Added in the other file where the results are easier to assert

infiton (approved):
nice!

PR #449, opened 2024-07-03
react-big-commerce

infiton:
This adds a react provider for our BigCommerce connection. The provider has two jobs:

1. Set up the api client to use the session token passed in a query parameter that we will set when rendering big commerce apps using their `/load` callback and load the session to make sure that it is authenticated
2. If we are in an authenticated context initialize the BigCommerce sdk (this will be loaded on the entry html)

## PR Checklist

### DIFF
--- a/packages/react-big-commerce/spec/Provider.spec.tsx
+++ b/packages/react-big-commerce/spec/Provider.spec.tsx
@@ -1,0 +1,144 @@
+import type { AnyClient } from "@gadgetinc/api-client-core";
+import { GadgetConnection } from "@gadgetinc/api-client-core";
+import "@testing-library/jest-dom";
+import { render } from "@testing-library/react";
+import React from "react";
+import { mockUrqlClient } from "../../api-client-core/spec/mockUrqlClient.js";
+import { Provider } from "../src/Provider.js";
+
+describe("GadgetProvider", () => {
+  let mockApiClient: AnyClient;
+  const mockBigCommerceInit = jest.fn();
+  let setAuthenticationModeSpy: jest.SpyInstance;
+  const { location } = window;
+
+  beforeAll(() => {
+    // @ts-expect-error mock
+    delete window.location;
+
+    // @ts-expect-error mock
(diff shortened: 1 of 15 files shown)


pistachiobaby (approved):

packages/react-big-commerce/src/Provider.tsx
```diff
@@ -0,0 +1,123 @@
...
+    if (!sessionToken) {
+      console.log("[gadget-react-big-commerce] no session token, skipping auth setup");
+      return;
+    }
+
+    api.connection.setAuthenticationMode({
+      custom: {
+        async processFetch(_input, init) {
+          const headers = new Headers(init.headers);
+          headers.append("Authorization", `BigCommerceSignedPayload ${sessionToken}`);
```
airhorns, line 45: Requires update here if you decide to change this gadget side

packages/react-big-commerce/src/context.ts
```diff
@@ -0,0 +1,21 @@
+import { createContext, useContext } from "react";
+
+export type GadgetAuthContextValue = {
```
airhorns, line 3: Maybe worth a slightly more specific type name for clarity

packages/react-big-commerce/spec/Provider.spec.tsx
```diff
@@ -0,0 +1,144 @@
...
+      stale: false,
+      hasNext: false,
+    });
+
+    expect(container.outerHTML).toMatchInlineSnapshot(`"<div><span>hello world</span></div>"`);
+    expect(mockBigCommerceInit).not.toHaveBeenCalled();
+  });
+
+  test("when there is a session parameter", () => {
+    window.location.search = "session=test-session-token";
```
airhorns, line 93: Thoughts on keeping the query param the same as the one that BigC sends for clarity? `signed_jwt_token` or whatever it is?

infiton: :+1:

PR #450, opened 2024-07-04
Updated implementations of belongsTo and hasMany auto-inputs

MillanWangGadget:
- **UPDATES**
  - New relationship hooks
    - Individual controller hooks for belongsTo, hasOne, hasMany field types
    - Shared hooks for options retrieval and selected record retrieval
    - Options retrieval has searchability and pagination-on-scroll
  - BelongsTo
    - Updated to retrieve related model record value and show the text label as a deselectable pill
      - Selected record retrieval is done by querying the `currentModel.belongsToField {...}`
  - HasMany
    - Updated to retrieve related model record value and show the text label as a deselectable pill
      -  Selected record retrieval is done with `useFindMany` on the related model, filtering on the current model records ID in the inverse field
    - Deselection is now possible. This is done with `hasManyField: [{update: {id: "1",  inverseField:{_link:null}}}]` in the submitted GQL request. The useActionForm GQL request shaper has been updated to allow for this
  - Initial stubbed support for hasOne fields 
    - Currently disabled until we update the API to forcefully maintain the 1:1 mapping between records on the related models 
    - Hooks are still around for potential future usage
    - Incorporated namespace model `game.city` hasOne `game.stadium` into the testing environment
 - **FUTURE STEPS**
   - Update search to filter non-matching selected values. Currently only filters on un-selected value options
   - Incorporate `dataModel.defaultDisplayField` to get the related model record option labels
   - 

## PR Checklist


packages/react/src/validationSchema.tsx
```diff
@@ -121,7 +121,7 @@ const validatorForField = (field: FieldMetadata) => {
     validator = (validator.nullable() as any).default(null);
   }
 
-  for (const validation of field.configuration.validations) {
+  for (const validation of field.configuration.validations ?? []) {
```
alanko0511, line 124: I don't think this should have `?? []` because the GraphQL schema said it must return the validations. Having this fallback feels like a lie.

If you do this because of the introduction of the `gameCityModel` snapshot and the fields don't need any validations, you can add `validations: []` under the `configuration` object.

MillanWangGadget: Updated such that there is no more `??[]` here and added empty validation arrays to all of the fields in game.city

packages/react/src/auto/hooks/useRelationshipInputController.tsx
```diff
@@ -69,32 +71,85 @@ const useRelatedModelManager = (props: {
...
-    selectedRecordId ?? "",
-    {
-      pause: !selectedRecordId, // Only fetch with a selected records
-    }
+  const [selectedRecordIds, setSelectedRecordIds] = useState<string[]>(
+    isBelongsToField && hasInitialBelongsToSelection ? [fieldProps.value] : []
   );
 
+  // These are needed for HasOne/HasMany fields. To show that a record on the other model is deselected, we need to excluded it from this query
+  const [deselectedRecordIds, setDeselectedRecordIds] = useState<string[]>([]);
```
airhorns, line 92: Can you use the form state for this instead of setStates? It seems like this is going to be annoying to keep in sync with the form state, and would not update if someone did a manual set into the form state from somewhere else

MillanWangGadget: This `deselectedRecordIds` useState is only used as an excluder filter on the the useFindMany call that looks up the selected related model records, because without it, there would be no way to deselect related records on hasOne/hasMany fields. 
This info can't be stored on the current field's path for the form state without disrupting how the field values are converted into GQL requests. I suppose this could be stored on a separate form path, but that is effectively the same as doing a useState with unused extra functionality

MillanWangGadget:
@airhorns @alanko0511 This is ready re re-review 😇

packages/react/src/auto/AutoFormContext.ts
```diff
@@ -12,6 +12,10 @@ export interface AutoFormMetadataContext {
   metadata: ActionMetadata | undefined;
   submit: UseActionFormSubmit<any>;
   submitResult?: AutoFormSubmitResult;
+  model?: {
+    namespace?: string | string[] | null;
+    apiIdentifier?: string;
+  };
```
airhorns, line 18: Is this different than the `metadata`? I think metadata there is actually at the model level already

packages/react/src/auto/hooks/useHasManyInputController.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const { fields, remove, append } = useFieldArray({ name: path });
+  const { options, pagination, search, selected, relatedModel } = useRelatedModelOptions(props);
+  const removeFromFieldsByRecordId = (recordId: string) => {
+    const index = fields.findIndex((entry) => (entry as any).__id === recordId);
+    if (index > -1) {
+      remove(index);
+    }
+  };
+
+  const onSelectRecord = (recordId: string) => {
```
airhorns, line 19: tiny nit but probably wise to use some `useCallback`s around these suckers

packages/react/src/auto/hooks/useHasManyInputController.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const isAlreadySelected = selectedRecordIds.includes(recordId);
```
airhorns, line 20: tiny nit: its a little confusing that this variable is defined below here instead of above here

packages/react/src/auto/hooks/useRelatedModelManager.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return {
+    path,
+    metadata,
+    modelManager,
+    model: {
+      apiIdentifier,
+      namespace,
+      inverseFieldApiIdentifier,
+    },
+  };
```
airhorns, line 38: This is just a style thing but it seems weird that this hook returns all this extra stuff. I think it might be a little easier to understand if it just returned the model manager (do one thing and one thing only)

packages/react/src/auto/hooks/useRelatedModelManager.tsx
```diff
@@ -0,0 +1,39 @@
+import { useApi } from "../../GadgetProvider.js";
+import { GadgetHasManyConfig } from "../../internal/gql/graphql.js";
+import { getModelManager } from "../../utils.js";
+import { useFieldMetadata } from "./useFieldMetadata.js";
+
+export const useRelatedModelManager = (props: {
+  field: string; // Field API identifier
```
airhorns, line 7: I think this API is gonna get dicey if we ever do deep relationship nesting, where you can adjust a model two relationships over because so much of this thing's behavior is determined by the context. What do you think about accepting the metadata as an argument to make it a bit clearer and adjustable?

packages/react/src/auto/hooks/useRelatedModelOptions.tsx
```diff
@@ -0,0 +1,246 @@
...
+  );
+};
+
+/**
+ * For getting the selected record in a BelongsTo relationship
+ * Returns the selected record in an array for interoperability with the HasOne/HasMany hook
+ *
+ * The lookup is done using the `findBy` to lookup on the current model to retrieve the related model record data
+ */
+export const useLinkedChildModelRelatedModelRecords = (props: AutoRelationshipInputProps) => {
```
airhorns, line 89: One thing to think about is that all these hooks would be really useful for someone building their own has many selector component, but right now are hard-coupled to the autoform context and metadata. We don't have to adjust it right now, but I'd love to strive to make these things not need any metadata at all and instead just get given the lowest level stuff they need as input so they could eventually be exposed as headless hooks for folks who want the fine-grained UX control. I really like that we're bottling up the functionality in hooks, just wanna make em maximally reusable.

packages/react/src/auto/hooks/useRelatedModelOptions.tsx
```diff
@@ -0,0 +1,246 @@
...
+}) => {
+  const { field, optionLabel } = props;
+  const optionLabelIsFieldName = typeof optionLabel === "string";
+
+  const {
+    modelManager: relatedModelManager,
+    model: { apiIdentifier, namespace, inverseFieldApiIdentifier },
+  } = useRelatedModelManager({ field });
+
+  const [loadedRecords, setLoadedRecords] = useState<any[]>([]);
```
airhorns, line 179: I think if someone clicks through enough pages this could get really big -- do you wanna keep them all loaded in memory or just the last loaded page? I don't feel strongly

packages/react/src/auto/interfaces/AutoRelationshipInputProps.tsx
```diff
@@ -3,5 +3,10 @@ import { Control } from "react-hook-form";
...
   field: string;
   control?: Control<any>;
-  optionLabel?: string | ((record: Record<string, any>) => string);
+  optionLabel?: OptionLabel;
 }
+
+/**
+ * Type for the option label when displaying a list of records from a related model
+ */
+export type OptionLabel = string | ((record: Record<string, any>) => string);
```
airhorns, line 12:
```suggestion
export type OptionLabel = ReactNode | ((record: Record<string, any>) => ReactNode);
```

so folks can pass styled elements from the design system or what have you

packages/react/src/auto/polaris/inputs/relationships/PolarisListMessages.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <Text as="p" tone="subdued">
+      {props.message}
+    </Text>
+  </div>
+);
+
+export const NoRecordsMessage = () => <ListMessage message="No records found" />;
+
+export const getErrorMessage = (errorString: string) => {
+  if (errorString.includes("GGT_PERMISSION_DENIED")) return `Cannot read records from related model`;
```
airhorns, line 24: Can we use content here that indicates that it is a permissions error so that a dev knows to go fix that ?

packages/react/src/auto/hooks/useRelationshipInputController.tsx
MillanWangGadget: This approach is no longer around

packages/react/src/auto/hooks/useHasManyInputController.tsx
MillanWangGadget: Updated to use `useCallbacks` throughout

packages/react/src/auto/hooks/useHasManyInputController.tsx
MillanWangGadget: Updated the order of definitions such that variables come before callbacks

packages/react/src/auto/hooks/useRelatedModelManager.tsx
MillanWangGadget: Good call. I've replaced the `spread and forward them spread` style with a `spread locally, forward the unspread` style when possible. Definitely helps to reduce the number of forwarded things

packages/react/src/auto/hooks/useRelatedModelOptions.tsx
MillanWangGadget: Polaris list only has a `onScrolledToBottom` callback, without a `scrollToTop` equivalent. Without something like this, if you scroll past a page containing something you want, you'd have to close->reopen the list to go back to a previous page. Doing this hurts more the deeper that the pages get, especially if the thing you want is hard to search for

packages/react/src/auto/interfaces/AutoRelationshipInputProps.tsx
MillanWangGadget: Updated to be a node. Good call 👍

packages/react/src/auto/polaris/inputs/relationships/PolarisListMessages.tsx
MillanWangGadget: Updated the message to indicate that there are missing permissions

packages/react/src/auto/hooks/useRelatedModelManager.tsx
MillanWangGadget: Updated this to take in the field config as a prop instead of a field-apiID so that being within an AutoForm is not required. This calls a super general hook that takes in `model: {apiId, namespace}`

packages/react/src/auto/hooks/useRelatedModelOptions.tsx
MillanWangGadget: WIP

Currently in the midst of refactoring this such that AutoForm specific stuff usage is minimized/eliminated. A few things remain that I want to come back to with fresh eyes

danroberts:
This would be fine as a follow up PR but could you also add stories for this field?

packages/react/src/auto/polaris/PolarisAutoForm.tsx
```diff
@@ -41,6 +41,10 @@ export const PolarisAutoForm = <
       isSuccessful: isSubmitSuccessful,
       error: formError ?? metadataError,
     },
+    model: {
+      apiIdentifier: action.modelApiIdentifier,
+      namespace: action.namespace,
+    },
```
alanko0511, line 47: Why do we want these properties from the action instead of the metadata response?

packages/react/src/auto/AutoFormContext.ts
MillanWangGadget: The response metadata from useActionMetdata removes the model api id when it goes deeper into the metadata object to get the action metadata. These values from the action are also used as the GQL query variables in `useActionMetadata`

MillanWangGadget:
> After you select a gizmo in the storybook, when you start to type a search for another gizmo, the label reverts back to the gizmo id rather than the display label (it seems to be related to if the new search value doesn't match the current label)  
- I've reworked the option and search system such that the label shall be always shown 😎 

> I'm not sure why this text ended up red. Is there a bug on an error?  
- The tag text should only go red if you have a belongs to field points points to an ID that does not exist on the neighbouring value. I'm real confused as to how this screenshot happened since having the label loaded means that there must be a value on the other table. When red, it should have label text like `id: 123` only.

> I don't think there is anything we can do to fix this, but the search includes other properties on the record -- for example, search '1' in the gizmo field and you get very confusing results, I think because their ids match 1. We really need a way to control what we're searching for. In this case, maybe we actually should do a filter by the field that we're using as the display field rather than a search param.
- I kinda got around this by having 2 stages of filtering from the search
  - 1st - The search value is passed into `useFindMany`. 
    - This was the only searching done previously and  included records with labels that don't match the search
    - This is still needed to get records that match the search that are not on the first page
  - 2nd - A `option.label.contains(search.value)` type of filter is run on the options so that records that come from the search that don't have a matching label are ignored

danroberts (approved):

alanko0511 (approved):
:shipit: 👌

packages/react/src/use-action-form/utils.ts
MillanWangGadget, line 1: @pistachiobaby Hey Nick, can you look over this lil change in how the GQL requests are made for HasOne/HasMany/HasManyThrough fields to make sure that this won't break any useActionForm stuff?

Leading double underscores are illegal GQL names which makes me feel safer using them here as internal control variables

PR #451, opened 2024-07-04
Add client-side string/number validations to AutoForm

alanko0511:
This PR adds the string or number's validations to the AutoForm for simple client-side validation and prevents submitting the form when there are validation errors.

Right now, we implement the following validations:
- Regex pattern
- String size range
- Number range

## PR Checklist


packages/react/src/auto/hooks/useStringInputController.tsx
```diff
@@ -32,7 +32,6 @@ export const useStringInputController = (props: {
     fieldState: { error },
   } = useController({
     name: path,
-    rules: { required: metadata.requiredArgumentForInput },
```
alanko0511, line 35: The existing "required" validation is actually done in the `validationSchema.ts` file, so I removed places that make this custom rule to prevent confusion.

packages/react/src/validationSpecKeys.ts
```diff
@@ -0,0 +1,13 @@
...
+export const UrlValidationId = "gadget/validation/url";
+export const StringSizeValidationId = "gadget/validation/string-size";
+export const NumberRangeValidationId = "gadget/validation/number-range";
+export const RegexValidationId = "gadget/validation/regexp";
+export const ColorValidationId = "gadget/validation/color";
+export const VectorDimensionsValidationId = "gadget/validation/vector-dimensions";
+export const UniqueValidationId = "gadget/validation/unique";
+export const InvokeCodeValidationId = "gadget/validation/code";
+export const FileSizeRangeValidationId = "gadget/validation/file-size-range";
+export const ValidImageFileValidationId = "gadget/validation/valid-image-file";
```
airhorns, line 13: I think it might be a touch simpler to use the GraphQL typename of the object instead of these spec ids. GraphQL forces us to have a typename for every object, and so why have another identifier that just means the same thing here, you know? Will avoid an unnecessary increase in the bundle size as well

airhorns (approved):
LGTM after comments

packages/react/src/auto/polaris/inputs/PolarisAutoInput.tsx
```diff
@@ -20,7 +20,6 @@ export const PolarisAutoInput = (props: { field: string }) => {
     fieldState: { error },
   } = useController({
     name: path,
-    rules: { required: metadata.requiredArgumentForInput },
```
MillanWangGadget, line 23: MUI auto input still has this

packages/react/src/metadata.tsx
```diff
@@ -29,6 +29,32 @@ const FieldMetadataFragment = graphql(/* GraphQL */ `
     configuration {
       __typename
       fieldType
+      validations {
+        ... on GadgetRegexFieldValidation {
+          __typename
```
airhorns, line 34: You can stick this outside the `... on` parts before the first one to select it always, regardless of which union element the server returns

MillanWangGadget (approved):
MANUAL TESTING
- 🍏 Create actions do not send network events for validation errors that can be checked client side
- 🟡 Update actions still send network events when tehre are validation errors before making the corresponding inputs red

One minor code comment, but let's consider the yellow separately to get in the GQL changes for validations

PR #457, opened 2024-07-05
debug logging for useActionForm

infiton:
adding some debug logs for useActionForm that helped in debugging with @DevAOC 

## PR Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.15.16",
+  "version": "0.15.15",
   "files": [
     "README.md",
     "dist/**/*"
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.15.15",
+  "version": "0.15.16",
   "files": [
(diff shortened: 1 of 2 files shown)


pistachiobaby (approved):

packages/react/src/utils.ts
```diff
@@ -422,3 +422,31 @@ export const getModelManager = (
 ): AnyModelManager | undefined => {
   return get(apiClient, namespaceDataPath([apiIdentifier], namespace).join("."));
 };
+
+export class Logger {
```
pistachiobaby, line 426: @infiton You cool with this? I figured this is the most elegant way to handle enabling and disabling the debug for useActionForm

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.15.15",
+  "version": "0.15.16",
```
infiton, line 3: Maybe don’t bump the version and we can just let this go out with the next version bump

packages/react/src/utils.ts
infiton: You don't want to just check the flag each time? And use console directly? I doubt this increase bundle size much at all, but even then not sure an extra class really buys us much

PR #458, opened 2024-07-05
Added password input component

MillanWangGadget:
- **UPDATE**
  - Added support for password fields with password replacement system when given a findBy value

## PR Checklist


packages/react/spec/auto/inputs/PolarisPasswordInput.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+    operatesWithRecordIdentity: true,
+  });
+};
+
+const metadata: ActionMetadata = {
+  name: "Widget",
+  action: {
+    name: "Create",
+    apiIdentifier: "create",
+    inputFields: [
```
alanko0511, line 55: Looks like we don't use this variable in this file.

alanko0511 (approved):
One comment but other than that :shipit: 👌

PR #461, opened 2024-07-06
JSON input error states

airhorns:
Previously, the `useJSONInputController` was tracking its own error from invalid JSON inputs in its own `setState` call. This state was local to the component, so it didn't stop the form from being submitted. Instead, we need to store the error in the form state such that `react-hook-form` knows about it, and can revalidate using the validation schema. This implements that! It's still a little tricky because we need to store the raw input string the user is typing into and store the real JSON value in the form state so that it gets submitted. The controller now queues the validator that there has been a parse error by writing a special object indicating that there has been a parse error into the form state.

### DIFF
--- a/packages/react/src/auto/hooks/useJSONInputController.tsx
+++ b/packages/react/src/auto/hooks/useJSONInputController.tsx
@@ -1,48 +1,59 @@
 import { useCallback, useEffect, useState } from "react";
-import { type Control } from "react-hook-form";
+import deepEqual from "react-fast-compare";
+import type { UseControllerProps } from "react-hook-form";
+import { isFailedJSONParse, type FailedJSONParse } from "../../validationSchema.js";
 import { useStringInputController } from "./useStringInputController.js";
 
-export const useJSONInputController = (props: {
-  field: string; // The field API identifier
-  control?: Control<any>;
-}) => {
-  const { field, control } = props;
-  const stringInputController = useStringInputController({ field, control });
+export const useJSONInputController = (
+  props: {
+    field: string;
+  } & Omit<UseControllerProps, "name">
+) => {
+  const stringInputController = useStringInputController(props);
+  const jsonValue = stringInputController.value;
+  const isParseError = isFailedJSONParse(jsonValue);
 
-  const [error, setError] = useState<Error | undefined>();
-  const [jsonValue, setJSONValue] = useState<any>(stringInputController.value);
-  const [stringValue, setStringValue] = useState<string>(JSON.stringify(jsonValue, null, 2));
+  const [string, setString] = useState<string>(stringInputController.value && !isParseError ? JSON.stringify(jsonValue, null, 2) : "");
 
-  const onStringValueChange = useCallback(
-    (newString: string) => {
-      setStringValue(newString);
+  // propagate imperative changes from the form state into the string input, but only if the parsed representations aren't equal
+  useEffect(() => {
+    if (typeof jsonValue != "undefined" && jsonValue != null && !isParseError) {
       try {
-        const newJSON = JSON.parse(newString);
-        setJSONValue(newJSON);
-        setError(undefined);
-        stringInputController.onChange(newJSON);
-      } catch (error: any) {
-        if (newString !== "") {
-          setError(error);
-        } else {
-          setError(undefined);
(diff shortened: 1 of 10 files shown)


packages/react/src/auto/mui/inputs/MUIAutoJSONInput.tsx
```diff
@@ -11,24 +11,23 @@ export const MUIAutoJSONInput = (
...
-      {...originalController}
-      value={stringValue}
-      onChange={(event) => onStringValueChange(event.target.value)}
       error={inErrorState}
-      helperText={inErrorState && `Invalid JSON: ${error.message}`}
+      helperText={inErrorState && `Invalid JSON: ${errorMessage}`}
+      {...controller}
       {...focusProps}
       {...restOfProps}
+      onChange={(event) => controller.onChange(event.target.value)}
```
MillanWangGadget, line 30: An `onChange` passed into this component will not run because of this. Perhaps this should be 
```
    onChange={(event) => {
        controller.onChange(event.target.value);
        props.onChange?.(event);
      }}
```

airhorns: Is this something we're trying to keep support for across the board?

MillanWangGadget: The `onChange` would just need to go up a few lines to allow for users to override it. I don't see any onChange or onClick examples where we do the onChange from the controller as well as the passed in onChange prop. Although I realize now that some people may want to completely avoid the default behaviour and we should allow for that

airhorns: As discussed in slack we're just not gonna expose `onChange` for now since its not quite clear if it should be an override or a participant or if a user-supplied one can influence ours 👍

alanko0511 (approved):
:shipit: 👌

packages/react/src/auto/hooks/useJSONInputController.tsx
```diff
@@ -1,48 +1,54 @@
...
-  const [jsonValue, setJSONValue] = useState<any>(stringInputController.value);
-  const [stringValue, setStringValue] = useState<string>(JSON.stringify(jsonValue, null, 2));
+  const [string, setString] = useState<string>(stringInputController.value && !isParseError ? JSON.stringify(jsonValue, null, 2) : "");
 
-  const onStringValueChange = useCallback(
+  // propagate imperative changes from the form state into the string input, but only if the parsed representations aren't equal
+  useEffect(() => {
+    if (typeof jsonValue != "undefined" && !isParseError) {
+      try {
+        const stringJSONValue = JSON.parse(string);
```
MillanWangGadget, line 22: `JSON.parse("")` throws an error if the string is empty. This makes it impossible to clear out an existing record's JSON field value

packages/react/src/auto/mui/inputs/MUIAutoJSONInput.tsx
```diff
@@ -11,24 +11,23 @@ export const MUIAutoJSONInput = (
     control?: Control<any>;
   } & Partial<TextFieldProps>
```
MillanWangGadget, line 12: `Partial<Omit<TextFieldProps, "onChange">>` since passing in `onChange` will now do nothing

packages/react/src/auto/polaris/inputs/PolarisAutoJSONInput.tsx
```diff
@@ -12,21 +12,17 @@ export const PolarisAutoJSONInput = (
     control?: Control<any>;
   } & Partial<TextFieldProps>
```
MillanWangGadget, line 13: Perhaps `Partial<Omit<TextFieldProps, "onChange">>` for the same reason as the MUI variant

packages/react/src/auto/hooks/useJSONInputController.tsx
airhorns: Nice, fixed!

MillanWangGadget (approved):
- 🍏 Empty string in the JSON field makes the record null, thus allowing nullifications of non-null values

Good to merge 👍

PR #462, opened 2024-07-06
Create autoform default values

airhorns:
When building an autoform for a create action, the user should be able to provide base values for the inputs. These might be different than the backend defaults as they might be based on client side state like who's logged in, the user's preferences, etc.

`useActionForm` supports default values just fine, but previously AutoForm did not. This adds support by just passing it down to the `useActionForm` call, but there's an important ramification: if we want to actually submit the default values to the server, then we need to no longer do the dirty check introduced in https://github.com/gadget-inc/js-clients/pull/441.  Default values should be submitted, but they aren't necessarily dirty because the user didn't necessarily touch them. So, this PR undoes that change and goes back to submitting everything every time.

I think that generally speaking, the only-submit-dirty-stuff thing is unexpected. Normal HTML forms, and `useActionForm`, all submit all the values present in the form every time the form is submitted. I think that's what a user would expect too -- if they see form state on screen, and then they hit submit and go to a show page for a resource, they expect to see what the form contained. So, I think we aughta submit everything anyways.

That said, I believe it was introduced because the API doesn't like it when we submit `null` for certain field types. I think we should fix that problem at that layer, instead of introducing a behaviour divergence from the way normal forms and `useActionForm` work at this layer.

@MillanWangGadget @alanko0511 do you remember what specifically broke when we sent all the field values, and even better have a failing test case or something like that?


alanko0511:
> do you remember what specifically broke when we sent all the field values, and even better have a failing test case or something like that?

It throws an internal error when the relationship field includes a null value, but we fixed that at the API level. Another issue is the one you mentioned:

> That said, I believe it was introduced because the API doesn't like it when we submit null for certain field types.

Unless we provide the default field value, I don't really know how to identify if the user is trying to pass in a `null` value (if the field is nullable) or an undefined value and tell the server to use the default value.

airhorns:
> Unless we provide the default field value from the editor as part of the metadata to generate the form, I don't really know how to identify if the user is trying to pass in a null value (if the field is nullable) or an undefined value and tell the server to use the default value.

Ah, I see. I think that the dev being able to pass default values here is important enough that we should just do that. I will add a way to get the defaults to the metaschema and then update this PR to do that.

airhorns:
Ok, this is now fetching and prepopulating with the default values for all fields from the server. If you decide to do `<Autoform defaultValues={...}/>` on your own, those take precedence.

danroberts:
> @MillanWangGadget @alanko0511 do you remember what specifically broke when we sent all the field values, and even better have a failing test case or something like that?

Relationship fields throw errors when you submit null values to them. It looks like it was fixed at the api layer here: 
https://github.com/gadget-inc/gadget/pull/12084

But I'm not sure that fix captures the need to be able to 'remove' a relationship i.e. unlink a model, which would most naturally be represented by submitting a null value.

MillanWangGadget:
> But I'm not sure that fix captures the need to be able to 'remove' a relationship i.e. unlink a model, which would most naturally be represented by submitting a null value.

I recently merged this which allows unlinking models by doing something like this
https://github.com/gadget-inc/gadget/pull/12177

```
{
  "parentModel": {
    "otherParentField": "abc",
    "parentHasManyChildField": {
      "update": {
        "id": "111", // childModel record id
        "otherChildModelField": "xyz",
        "childBelongsToParentField": {
           "_link": null // Nullified to unlink the record
        }
      }
    }
  },
  "id": "333" // ParentModel record id
}
```

MillanWangGadget (approved):
MANUAL TESTING 
- 🍏 The default field values properly populate the form inputs for create actions
  - 🍏 Works for all field types
  - 🟡 DateTime fields with `includeTime` enabled does not work, but I think we can consider that as a separate concern
- 🍏 For non-create actions, the form inputs are `hasGivenFindBy ? matchingRecordValue : Blank`
- 🍏 Create actions always have their fields populated based on the default values. No more `match value of existing record when given findBy value` behaviour

Code changes look good 👍

danroberts (approved):
👍 on 🟢

PR #464, opened 2024-07-08
Implement file input for auto component

alanko0511:

## PR Checklist


packages/react/src/validationSchema.tsx
```diff
@@ -126,14 +124,47 @@ const validatorForField = (field: FieldMetadata) => {
...
+                }
+              } else {
+                if (max) {
+                  return createError({
+                    path,
+                    message: `must be a file smaller than ${filesize(max)}, and this file's size is ${filesize(value.$invalidFileSize)}.`,
+                  });
+                }
+              }
+            }
```
MillanWangGadget, line 164: The message is the only ting that varies here and the `createError` stuff is repeated. Could you keep just the messages in conditional blocks, and use a single return statement?
```ts
if (min && max) message="...";
else if (min) message="...";
else if (max) message="...";

return createError({message, ...})
```

packages/react/src/auto/mui/inputs/MUIAutoFileInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+        },
+        body: files[0],
+      });
+      fieldProps.onChange({
+        directUploadToken: token,
+        fileName: files[0].name,
+      });
+    },
+    [api, fieldProps]
+  );
```
MillanWangGadget, line 51: `useFieldMetadata, useApi, useController, fetch->fieldProps.onChange(..)` calls are basically the same between the MUI and Polaris variants. Could you move them to a common hook, perhaps something like`useFileInputController` which returns these common functionalities

packages/react/src/auto/polaris/PolarisAutoFileInput.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { filesize } from "filesize";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+import type { Control } from "react-hook-form";
+import { useController } from "react-hook-form";
+import { useApi } from "../../GadgetProvider.js";
+import type { GadgetOnlyImageFileFieldValidation, GadgetRangeFieldValidation } from "../../internal/gql/graphql.js";
+import { isAutoFileFieldValue, type AutoFileFieldValue } from "../../validationSchema.js";
+import { useFieldMetadata } from "../hooks/useFieldMetadata.js";
+
+const imageFileTypes = ["image/jpeg", "image/png", "image/svg+xml", "image/webp", "image/gif"];
```
MillanWangGadget, line 13: `"image/gif"` should not be here since it's added if animations are allowed

packages/react/src/auto/polaris/PolarisAutoFileInput.tsx
```diff
@@ -0,0 +1,175 @@
...
+        if (rangeValidation.max) {
+          actionHintParts.push(`between ${filesize(rangeValidation.min)} and ${filesize(rangeValidation.max)}`);
+        } else {
+          actionHintParts.push(`larger than than ${filesize(rangeValidation.min)}`);
+        }
+      } else {
+        if (rangeValidation.max) {
+          actionHintParts.push(`smaller than ${filesize(rangeValidation.max)}`);
+        }
+      }
```
MillanWangGadget, line 119: Seems to overlap with similar validation in `packages/react/src/validationSchema.tsx`

packages/react/src/auto/polaris/PolarisAutoFileInput.tsx
```diff
@@ -0,0 +1,175 @@
...
+      const baseFileValue = {
+        fileName: file.name,
+        mimeType: file.type,
+      } satisfies AutoFileFieldValue;
+
+      setImageThumbnailURL(window.URL.createObjectURL(file));
+
+      if (
+        rangeValidation &&
+        ((rangeValidation.min && file.size < rangeValidation.min) || (rangeValidation.max && file.size > rangeValidation.max))
```
MillanWangGadget, line 68: Also seems to overlap with similar validation in `packages/react/src/validationSchema.tsx`

MillanWangGadget (commented):
Manual tests
- 🟡 Add a remove button for when there is no required validation 
- 🟡 File size validations only run after submitting the first time
- 🍏 Fle uploads work as expected for small files
- 🍏 Existing record files can be retrieved
- 🍏 An error message appears when trying clicking submit when a file is uploading

packages/react/cypress/component/auto/form/PolarisFileInput.cy.tsx
```diff
@@ -0,0 +1,348 @@
...
+import { PolarisAutoForm } from "../../../../src/auto/polaris/PolarisAutoForm.js";
+import { api } from "../../../support/api.js";
+import { PolarisWrapper } from "../../../support/auto.js";
+
+describe("PolarisFileInput", () => {
+  const getSubmitButton = () => {
+    return cy.get("form [type=submit][aria-hidden!=true]");
+  };
+
+  const interceptModelCreateActionMetadata = (validations: any[]) => {
```
MillanWangGadget, line 13: Can this be moved to a new file or does it not owrk when it is not done in the test file?

MillanWangGadget (approved):
TESTS
- 🍏 Non-required files can be removed from existing records
- 🍏 Required files cannot be removed, even for create actions
- 🍏 Removing a file in an update record properly clears the corresponding DB value

One small code comment but good otherwise 👍

packages/react/cypress/component/auto/form/PolarisFileInput.cy.tsx
alanko0511: No, but let me fix forward and have a separate PR to improve these Cypress util functions

PR #465, opened 2024-07-09
Introduce cypress utils for easier testing

alanko0511:
Now that we are starting to write more tests with Cypress, let's add more util functions so that we can write Cypress tests easily.

## PR Checklist


packages/react/cypress/component/auto/form/PolarisFileInput-ClearFileButton.cy.tsx
alanko0511, line 1: It turns out that moving the test to a higher order can fix the issue, so I removed this test file.

MillanWangGadget (approved):
Test updates look good. It's cool to see these Cypress helpers added to the command system

PR #466, opened 2024-07-09
Exclude `ref` from the controllers when passing into Polaris components

alanko0511:
Since the Polaris component doesn't expose the `ref` property with React functions like `forwardRef` and `react-hook-form`'s controllers include the `ref` property, we can only exclude it when passing into the components.

## PR Checklist


packages/react/src/utils.ts
```diff
@@ -418,3 +418,8 @@ export const getModelManager = (
 ): AnyModelManager | undefined => {
   return get(apiClient, namespaceDataPath([apiIdentifier], namespace).join("."));
 };
+
+export const getPropsWithoutRef = <T extends { ref: RefCallback<any> | RefObject<any> }>(props: T) => {
```
MillanWangGadget, line 422: Maybe add a comment here explaining polaris does not work with refs and they they need to as an example as to why the ref value may need to be removed

airhorns (approved):
Superb!

MillanWangGadget (approved):

PR #468, opened 2024-07-10
Add new behaviour for updating actions

pistachiobaby:
This PR addresses a handful of shitty experiences a user in the discord had.

1. When updating a `belongsTo` relationship, you may want to update the link AND also display information returned from a `findby` associated with the child.
2. Send wasn't filtering out data being pushed into reshape
3. Adds the ability to mark fields as `ReadOnly` using a select which will be excluded from reshaping

## PR Checklist

### DIFF
--- a/packages/react/spec/useActionForm.spec.tsx
+++ b/packages/react/spec/useActionForm.spec.tsx
@@ -1,6 +1,7 @@
 import { act, renderHook } from "@testing-library/react";
 import { assert, type IsExact } from "conditional-type-checks";
 import { CombinedError } from "urql";
+import { applyDataMask } from "../src/use-action-form/utils.js";
 import { useActionForm } from "../src/useActionForm.js";
 import { bulkExampleApi, fullAuthApi, kitchenSinkApi, nestedExampleApi, relatedProductsApi } from "./apis.js";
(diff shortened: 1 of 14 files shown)


packages/react/src/useActionForm.ts
```diff
@@ -24,6 +35,55 @@ import { get, getModelManager, set } from "./utils.js";
 
 export * from "react-hook-form";
 
+type AnyActionWithId<OptionsT> =
```
infiton, line 38: worth moving these into `type.ts`?

packages/react/spec/useActionForm.spec.tsx
```diff
@@ -2679,4 +2680,496 @@ describe("useActionForm", () => {
...
+          },
+        },
+      },
+      stale: true,
+      hasNext: false,
+    });
+
+    await act(async () => {
+      await submitPromise;
+    });
```
infiton, line 2781: shouldn't we make assertions about the state of the form after the submit?

infiton: oh it looks like this is tested below; are these tests duplicate?

infiton (approved):

PR #469, opened 2024-07-10
Fallback to current date if date value is undefined in PolarisDateTimePicker

alanko0511:
This PR fixes a bug about the PolarisDateTimePicker not providing the current date as a fallback when the input value is undefined.

Previously, it assumed the `fieldProps.value` must have a value. However, if the input has not been modified or has no default (e.g. when creating a record), `fieldProps.value` is actually `null`, making a date an "Invalid Date" and unable to render the calendar.

## PR Checklist

### DIFF
--- a/packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
+++ b/packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
@@ -69,10 +69,9 @@ export const PolarisDateTimePicker = (props: {
 
   const onDateChange = useCallback<Exclude<DatePickerProps["onChange"], undefined>>(
     (range) => {
-      const date = new Date(range.start);
-      fieldProps && copyTime(date, zonedTimeToUtc(new Date(fieldProps.value), localTz));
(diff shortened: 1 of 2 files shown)


packages/react/src/auto/polaris/inputs/PolarisDateTimePicker.tsx
```diff
@@ -106,8 +105,8 @@ export const PolarisDateTimePicker = (props: {
       >
         <div style={{ padding: "16px" }}>
           <DatePicker
-            month={getDateTimeObjectFromDate(zonedTimeToUtc(new Date(fieldProps.value), localTz)).month}
-            year={getDateTimeObjectFromDate(zonedTimeToUtc(new Date(fieldProps.value), localTz)).year}
```
alanko0511, line 110: Previously, it assumed the `fieldProps.value` must have a value. However, if the input has not been modified or has no default (e.g. when creating a record), `fieldProps.value` is actually `null`, making a date an "Invalid Date" and not showing the calendar.

alanko0511:
Cancel in favour of this
https://github.com/gadget-inc/js-clients/pull/470

PR #470, opened 2024-07-10
Fix date picker

gabeb03:
The Polaris date picker is completely broken:

CleanShot 2024-07-09 at 17 46 27

This PR fixes that, and patches some flaky tests.

Previously, it assumed the fieldProps.value must have a value. However, if the input has not been modified or has no default (e.g. when creating a record), fieldProps.value is actually null, making a date an "Invalid Date" and unable to render the calendar.

## PR Checklist


alanko0511 (approved):
:shipit: 👌

PR #472, opened 2024-07-10
GGT-6702 - Show selected date (PolarisDateTimePicker)

gabeb03:
The Polaris date picker referenced `props.value` as the selected value, but that prop is not updated when a user picks a new date, so it was not showing a selected date when the user selected a new date.

### Before:

https://github.com/gadget-inc/js-clients/assets/168455431/b61475c9-0506-4685-b1ba-2f33a4f36e56

This PR makes the date picker reference `localTime` as the selected value, since this piece of state is updated with the correct date whenever the user uses the date picker. 

### After:

https://github.com/gadget-inc/js-clients/assets/168455431/59659913-b33c-497b-bea9-777e63c888f0

💭 Should a cy test be written for this change? 

## PR Checklist

### DIFF
--- a/packages/react/cypress/component/auto/form/PolarisAutoDateTimePicker.cy.tsx
+++ b/packages/react/cypress/component/auto/form/PolarisAutoDateTimePicker.cy.tsx
@@ -1,14 +1,23 @@
 import { format } from "date-fns";
 import { utcToZonedTime } from "date-fns-tz";
-import React from "react";
+import React, { useState } from "react";
 import { PolarisAutoForm } from "../../../../src/auto/polaris/PolarisAutoForm.js";
 import { PolarisAutoDateTimePicker } from "../../../../src/auto/polaris/inputs/PolarisAutoDateTimePicker.js";
 import { api } from "../../../support/api.js";
 import { PolarisWrapper } from "../../../support/auto.js";
 
-const baseDate = new Date("2021-03-05T11:23:10.000Z");
+const baseDate = new Date("2021-03-05T11:23:00.000Z");
 const localTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
 const dateInLocalTZ = utcToZonedTime(baseDate, localTz);
+
+const TestComponentWithCustomOnChange = () => {
+  const [date, setDate] = useState(baseDate);
+  return (
+    <PolarisAutoForm action={api.widget.create}>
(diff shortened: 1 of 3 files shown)


danroberts (commented):
Can you also add a test for this?

MillanWangGadget (commented):
Change makes sense. Good after test 👍

gabeb03:
Make a bunch of changes so that this change would also work when the user populated the date picker using the component's `value` prop. 

The requested test was also written.

packages/react/src/auto/polaris/inputs/PolarisAutoInput.tsx
```diff
@@ -46,6 +46,9 @@ export const PolarisAutoInput = (props: { field: string }) => {
       return <PolarisAutoBooleanInput field={props.field} />;
     }
     case FieldType.DateTime: {
+      // TODO: If a user sets the value of the DateTimePicker, and does not provide a custom onChange function
+      // the value will never be updated.
+      // Should an error be thrown in this case? I.e. value && !onChange => throw error
```
gabeb03, line 51: NB; thoughts?
I'll remove this comment once I get feedback from reviewers

MillanWangGadget: I don't think an error is required in this case. From what I've seen, most UI libraries will allow you to pass in a value and leave the component unchangeable with that

MillanWangGadget:
I think the time picker needs a similar treatment to prioritize props.value over fieldProps.value in a variable. Currently when given a hardcoded date value, the time can be adjusted while the date cannot be, and it matches the DB value which makes me think that it is attached to the `fieldProps.value` instead of `props.value`

gabeb03:
@MillanWangGadget this has been fixed; the solution is sub-optimal; the patch introduces some hard to read ternary operators, but it should be ready to cut an alpha this afternoon.

MillanWangGadget (approved):
When an explicit value is passed in, the value is properly loaded into the time picker, but the time is editable while the date picker is not. I think this comes form the fact that the time picker has both `value?: Date;` and `localTime?: Date;` as input properties. Since the time picker should always be related to a date picker, I think that only using the single `localTime` and then having a `changeTheWholeDateTime` callback from the date picker passed into the time picker can help with ensuring that the process for changing the fieldProps value is unified

For now I think we can ship this for our internal testing and revisit these points 👍

PR #474, opened 2024-07-10
Autoforms for global actions

airhorns:
- load the metadata from the new global action metadata endpoint
 - update all the metadata thingies that derive data to use the global action input fields or the model action input fields
 - use slightly different copy in the success banners for global actions, meriting a DRY'd up controller
 - add some tests that hit both MUI and Polaris

### DIFF
--- a/packages/react/src/auto/AutoForm.ts
+++ b/packages/react/src/auto/AutoForm.ts
@@ -1,11 +1,11 @@
-import type { ActionFunction, GadgetRecord } from "@gadgetinc/api-client-core";
+import type { ActionFunction, GadgetRecord, GlobalActionFunction } from "@gadgetinc/api-client-core";
 import { yupResolver } from "@hookform/resolvers/yup";
 import type { ReactNode } from "react";
 import { useMemo } from "react";
 import type { RecordIdentifier } from "src/use-action-form/types.js";
 import type { GadgetObjectFieldConfig } from "../internal/gql/graphql.js";
-import type { ActionMetadata, FieldMetadata } from "../metadata.js";
-import { filterFieldList, useActionMetadata } from "../metadata.js";
+import type { ActionMetadata, FieldMetadata, GlobalActionMetadata } from "../metadata.js";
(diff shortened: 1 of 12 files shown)


packages/react/src/auto/AutoForm.ts
```diff
@@ -14,8 +14,7 @@ import { validationSchema } from "../validationSchema.js";
 export type AutoFormProps<
   GivenOptions extends OptionsType,
   SchemaT,
-  ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any>,
-  Options extends ActionFunc["optionsType"]
```
airhorns, line 18: This type was unused and not applicable to global actions, so I deleted it.

danroberts (approved):
👍

packages/react/src/auto/AutoForm.ts
```diff
@@ -38,24 +37,26 @@ export type AutoFormProps<
 /**
  * React hook for getting the validation schema for a list of fields
  */
-export const useValidationResolver = (metadata: ActionMetadata | undefined) => {
+export const useValidationResolver = (metadata: ActionMetadata | GlobalActionMetadata | undefined) => {
   return useMemo(() => {
     if (!metadata) return undefined;
-    return yupResolver(validationSchema(metadata.action.inputFields));
+    const action = "action" in metadata ? metadata.action : metadata;
```
MillanWangGadget, line 43: Perhaps helpers like 
```
isGlobalAction(metadata)
getActionFromMetadata(metadata)
```
could be added so that it is a bit more obvious if the metadata is related to a global/model action without needing to look through the action metadata types

MillanWangGadget (approved):
MANUAL TESTS
- 🍏 Namespaced GActions load a form based on the exported parameters in the same way that model actions do
- 🍏 non-namespaced GActions work in the same way
- 🍏 Include/exclude works for custom params

One super small optional code comment but good to go 👍

packages/react/src/auto/AutoForm.ts
airhorns: Good call!

PR #475, opened 2024-07-10
Run real browsers for cypress in CI

airhorns:
For parity with how we test locally and what users will experience


MillanWangGadget (approved):
CI Chrome cypress to match local testing makes sense
🇨🇾🇨🇾🇨🇾

PR #476, opened 2024-07-10
Add a rich text input for autocomponents and apply nice polaris styling

airhorns:
- uses https://mdxeditor.dev/editor/docs/getting-started which seems like a nice, medium weight markdown editor with lightweight customizability
 - lazyloads this component using `React.lazy` for smaller default bundle size for autocomponents, as richtext is not often used that much
 - wires up mdxeditor's base toolbar to show nice formatting buttons
 - styles the rendered markdown using polaris css variables so that it matches kinda nicely in polaris

[no-changelog-required]

### DIFF
--- a/packages/react/src/auto/shared/AutoRichTextInput.tsx
+++ b/packages/react/src/auto/shared/AutoRichTextInput.tsx
@@ -21,10 +21,11 @@ import {
 } from "@mdxeditor/editor";
 import "@mdxeditor/editor/style.css";
 import type { ForwardedRef } from "react";
-import React from "react";
+import React, { useEffect, useRef } from "react";
 import { useFormContext, type Control } from "react-hook-form";
 import { get } from "../../utils.js";
 import { useStringInputController } from "../hooks/useStringInputController.js";
+import { multiref } from "../hooks/utils.js";
 
 export default function AutoRichTextInput(
   props: {
@@ -36,6 +37,13 @@ export default function AutoRichTextInput(
   const { formState } = useFormContext();
   const { field, control, editorRef, ...rest } = props;
   const controller = useStringInputController({ field, control });
+  const innerRef = useRef<MDXEditorMethods>(null);
+
+  useEffect(() => {
+    if (innerRef.current) {
+      innerRef.current.setMarkdown(controller.value?.markdown ?? "");
+    }
+  }, [controller.value]);
 
   return (
     <MDXEditor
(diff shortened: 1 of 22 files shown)


danroberts (approved):
👍 on 🟢

packages/react/src/auto/shared/AutoRichTextInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+  props: {
+    field: string;
+    control?: Control<any>;
+    editorRef?: ForwardedRef<MDXEditorMethods> | null;
+  } & Omit<MDXEditorProps, "markdown" | "onChange">
+) {
+  const { formState } = useFormContext();
+  const { field, control, editorRef, ...rest } = props;
+  const controller = useStringInputController({ field, control });
+
```
MillanWangGadget, line 39: - `useFormContext` is not needed when you have the controller
  - `controller.value?.markdown === get(formState.defaultValues, field)?.markdown`
- Add this before returning MDXEditor
  - `if (!controller.value?.markdown) {   return null;  }`
  - Fixes the `loading in without the existing record value` issue

packages/react/src/auto/shared/AutoRichTextInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <MDXEditor
+      plugins={[
+        headingsPlugin(),
+        listsPlugin(),
+        quotePlugin(),
+        thematicBreakPlugin(),
+        markdownShortcutPlugin(),
+        linkDialogPlugin(),
+        diffSourcePlugin({
+          diffMarkdown: get(formState.defaultValues, field)?.markdown ?? "",
```
MillanWangGadget, line 50: Can be replaced with `controller.value?.markdown`

packages/react/src/auto/shared/AutoRichTextInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+              <BoldItalicUnderlineToggles />
+              <ListsToggle />
+              <CodeToggle />
+              <CreateLink />
+            </DiffSourceToggleWrapper>
+          ),
+        }),
+      ]}
+      contentEditableClassName="autoform-prose"
+      {...rest}
```
MillanWangGadget, line 69: Perhaps this should be moved to the absolute bottom in the event that the user wants to override `markdown, onChange, ref`

MillanWangGadget (commented):
- 🍏 In the dependencies deep end, I see.... Things look reasonable 👍 
- 🟡 Loading existing record values is race-y. Sometimes the input is not populated with the record value
  - CleanShot 2024-07-16 at 16 20 03
  - See code comment for easy fix 😎 

Some code comments but good otherwise 👍

MillanWangGadget:
Loading existing record values is reliable now 👍

PR #477, opened 2024-07-10
Implement enum input for auto component

alanko0511:
This PR introduces `AutoEnumInput` for the auto component.
When the input allows adding additional values, it provides an "Add ..." button, and when the input allows multiple selections, it becomes a multi-select dropdown.

The Polaris component version is based on the example from Polaris with modifications for our needs:
https://polaris.shopify.com/components/selection-and-input/combobox

## PR Checklist

### DIFF
--- a/packages/react/cypress/component/auto/form/PolarisAutoEnumInput.cy.tsx
+++ b/packages/react/cypress/component/auto/form/PolarisAutoEnumInput.cy.tsx
@@ -1,0 +1,345 @@
+import React from "react";
+import { getStadiumRecord } from "../../../../spec/auto/support/stadiumModel.js";
+import { PolarisAutoForm } from "../../../../src/auto/polaris/PolarisAutoForm.js";
+import { api } from "../../../support/api.js";
+import { PolarisWrapper } from "../../../support/auto.js";
+
+describe("PolarisEnumInput", () => {
+  const blurComboboxes = () => {
+    cy.get("#type-combobox-textfield").blur({ force: true });
+    cy.get("#tags-combobox-textfield").blur({ force: true });
+  };
+
+  beforeEach(() => {
+    cy.viewport("macbook-13");
+
+    cy.intercept(
+      {
+        method: "POST",
+        url: `${api.connection.endpoint}?operation=createStadium`,
+      },
+      {
+        data: {
+          game: {
+            createStadium: {
+              success: true,
+              errors: null,
+              stadium: {}, // The response content doesn't matter for the tests
+              __typename: "CreateGameStadiumResult",
+            },
+            __typename: "GameMutations",
+          },
+        },
+      }
+    ).as("createStadium");
+
+    cy.intercept(
+      {
+        method: "POST",
+        url: `${api.connection.endpoint}?operation=updateStadium`,
+      },
+      {
+        data: {
+          game: {
+            createStadium: {
+              success: true,
+              errors: null,
+              stadium: {}, // The response content doesn't matter for the tests
+              __typename: "CreateGameStadiumResult",
+            },
+            __typename: "GameMutations",
+          },
+        },
+      }
+    ).as("updateStadium");
+  });
+
+  it("should include the enum options in the dropdown", () => {
+    cy.mockModelActionMetadata(api, baseModelActionMetadata);
+
+    cy.mountWithWrapper(<PolarisAutoForm action={api.game.stadium.create} />, PolarisWrapper);
+    cy.get("#type-combobox-textfield").click();
+    cy.contains("football").should("exist");
+    cy.contains("basketball").should("exist");
+    cy.contains("baseball").should("exist");
+    // Close the "type" dropdown
+    blurComboboxes();
+
+    cy.get("#tags-combobox-textfield").click();
+    cy.contains("hello").should("exist");
+    cy.contains("world").should("exist");
+  });
+
+  describe("searching for an option", () => {
+    it("should show the selected value in the dropdown for single select enum", () => {
+      cy.mockModelActionMetadata(api, baseModelActionMetadata);
+
+      cy.mountWithWrapper(<PolarisAutoForm action={api.game.stadium.create} />, PolarisWrapper);
+      cy.get("#type-combobox-textfield").click();
+      cy.contains("football").click();
+      cy.get(".Polaris-InlineStack").should("contain", "football");
+
+      cy.get("#tags-combobox-textfield").click();
+      cy.contains("hello").parent().parent().click();
+      cy.get(".Polaris-InlineStack").should("contain", "hello");
+      cy.get(".Polaris-InlineStack").should("not.contain", "world");
+
+      blurComboboxes();
+
+      cy.getSubmitButton().click();
+      cy.get("@createStadium")
+        .its("request.body.variables.stadium")
+        .should("deep.equal", {
+          type: "football",
+          tags: ["hello"],
+        });
+    });
+
+    it("should allow searching for options in the dropdown", () => {
+      cy.mockModelActionMetadata(api, baseModelActionMetadata);
+
+      cy.mountWithWrapper(<PolarisAutoForm action={api.game.stadium.create} />, PolarisWrapper);
+      // For single select enums
+      cy.get("#type-combobox-textfield").type("foot");
+
+      // The dropdown items should be filtered
+      cy.contains("football").should("exist");
+      cy.contains("basketball").should("not.exist");
+      cy.contains("baseball").should("not.exist");
+
+      cy.contains("football").click();
+      cy.get(".Polaris-InlineStack").should("contain", "football");
+
+      // For multi-select enums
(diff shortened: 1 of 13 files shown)


packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </Listbox.Option>
+      );
+    });
+  }, [allOptions.length, filteredOptions, searchValue, selectedOptions]);
+
+  const addExtraOptionElement = useMemo(() => {
+    if (!allowOther) return null;
+
+    const noResults = searchValue && !allOptions.includes(searchValue);
+    return noResults ? <Listbox.Action value={searchValue}>{`Add "${searchValue}"`}</Listbox.Action> : null;
```
MillanWangGadget, line 76: If `searchValue` is make up of only spaces, it will get compressed into a single space. Maybe replacing them all with a unicode-looking space that won't get lost will avoid this

MillanWangGadget (commented):
MANUAL TESTS
- ❔ Selected options always go to the top
  - CleanShot 2024-07-10 at 17 05 30
  - If you accidentally select something and want to deselect it, you will need to move your mouse to do so. This gets worse the bigger that the list of enums gets
  - Let's get design input on this
  - Custom created options should probably be kept together. Perhaps with a category
- 🍏 AllowMultiple enum fields properly creates and updates record values as arrays
- 🍏 pickOnlyOne enum fields properly create single string enum value
- 🍏 AllowAnyString enum fields allow for option creation

packages/react/cypress/component/auto/form/PolarisAutoEnumInput.cy.tsx
```diff
@@ -0,0 +1,295 @@
...
+      cy.getSubmitButton().click();
+      cy.get("@updateStadium").its("request.body.variables.stadium").should("deep.equal", {
+        type: null,
+        tags: [],
+      });
+    });
+  });
+});
+
+const interceptModelActionMetadata = (options?: {
```
danroberts, line 226: I think we should try to avoid helpers like this which take a lot of different options. I think this ends up being pretty complicated and hard to reason about what the test is actually trying to do. 

Another pattern is to have a set of mock objects which extend each other. For example:

```
StadiumBaseMock = {
  ...base stuff
}

StadiumCreateMock = {...StadiumBaseMock, {action: {apiIdentifier: 'create', operatesWithRecordIdentity: false}}}

```

And then your test can import the mock that it needs.

packages/react/src/auto/hooks/useEnumInputController.tsx
```diff
@@ -0,0 +1,90 @@
+import { useCallback, useMemo, useState } from "react";
+import { useController, type Control } from "react-hook-form";
+import { useFieldMetadata } from "./useFieldMetadata.js";
+
+export const useEnumInputController = (props: {
```
danroberts, line 5: I like that we're using a hook here, that's great.

packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+          ))}
+        </InlineStack>
+      ) : null,
+    [allowMultiple, onSelectionChange, selectedOptions]
+  );
+
+  const optionItemElement = useMemo(() => {
+    if (allOptions.length === 0) return null;
+
+    const formatOptionText = (option: string) => {
```
danroberts, line 42: there is probably a small performance optimization here to put this in a useCallback() separately from the useMemo

packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const noResults = searchValue && !allOptions.includes(searchValue);
+    return noResults ? <Listbox.Action value={searchValue}>{`Add "${searchValue}"`}</Listbox.Action> : null;
+  }, [allOptions, allowOther, searchValue]);
+
+  const emptyStateElement = useMemo(() => {
+    if (allowOther || (optionItemElement && optionItemElement.length > 0) || !searchValue) return null;
+    return <EmptySearchResult title="" description={`No  found matching "${searchValue}"`} />;
+  }, [allowOther, optionItemElement, searchValue]);
+
+  useEffect(() => {
```
danroberts, line 84: This seems like a bit of an anti pattern -- a useEffect that only sets state values. If I understand this, it is basically saying "Clear/reset the search bar" -- another option would be to make this a function/useCallback and then invoke it where you're setting allowMultiple and selectedOptions

packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const listBoxElement = useMemo(
+    () =>
+      optionItemElement || addExtraOptionElement || emptyStateElement ? (
+        <Listbox
+          autoSelection={AutoSelection.None}
+          onSelect={(selected) => {
+            onSelectionChange(selected);
+            console.log("selectedOptions:", selectedOptions);
+            if (allowMultiple) {
```
danroberts, line 106: As per previous comment, if you extracted a 'clearSearch' function you could call it here

packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    if (selectedOptions.length > 0) {
+      // Set the search value to show the selected option in the input field
+      setSearchValue(selectedOptions[0]);
+    } else {
+      setSearchValue("");
+    }
+  }, [allowMultiple, selectedOptions, setSearchValue]);
+
+  const listBoxElement = useMemo(
```
danroberts, line 98: In general, I'm not sure that these helpers that create an element need to be memoized. They don't seem to do much complicated/expensive computation, and I think there is a risk of creating subtle bugs if we forget to add something to the cache invalidation array.

I think these can just be regular helpers that generate markup, like: 

```
let listBoxElement = null;
if(optionItemElement || addExtraOpetionElement || emptyStateElement) {

 listBoxElement = ( 
  <Listbox ..
  )
  ```

File: packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx [hunk 8]
```diff
@@ -0,0 +1,149 @@
...
+    filteredOptions,
+    searchQuery,
+    label,
+    metadata,
+    isError,
+    errorMessage,
+  } = useEnumInputController({ field: fieldApiIdentifier, control });
+  const { value: searchValue, setValue: setSearchValue } = searchQuery;
+
+  const selectedTagsElement = useMemo(
```
danroberts, line 25: Same as other comment, I'm not sure this needs to be memoized

packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }, [allOptions.length, filteredOptions, searchValue, selectedOptions]);
+
+  const addExtraOptionElement = useMemo(() => {
+    if (!allowOther) return null;
+
+    const noResults = searchValue && !allOptions.includes(searchValue);
+    return noResults ? <Listbox.Action value={searchValue}>{`Add "${searchValue}"`}</Listbox.Action> : null;
+  }, [allOptions, allowOther, searchValue]);
+
+  const emptyStateElement = useMemo(() => {
```
danroberts, line 79: Same comment again, would prefer no memoization here.

danroberts (approved):

packages/react/src/auto/mui/inputs/MUIAutoEnumInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+        if (typeof value === "string") {
+          onSelectionChange(value);
+        } else {
+          for (const option of value) {
+            onSelectionChange(option);
+          }
+        }
+      }}
+    />
```
MillanWangGadget, line 38: add `(...props}`

packages/react/src/auto/polaris/inputs/PolarisAutoEnumInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const inputLabel = (
+    <>
+      {label} {metadata.requiredArgumentForInput ? <span style={{ color: "var(--p-color-text-critical)" }}>*</span> : null}
+    </>
+  );
+
+  return (
+    <Combobox
+      {...comboboxProps}
```
MillanWangGadget, line 114: This should go below `allowMultiple` and `activator` or else those will be impossible to override

MillanWangGadget (approved):
All previous concerns are addressed or can be seen as separate concerns. Good to go 👍

PR #478, opened 2024-07-10
GGT-6724 - Fix month selection when month is paginated & Persist Time when Date Changes

gabeb03:
When a user presses the right or left arrow on the date picker popover, a new date is selected. 

https://github.com/gadget-inc/js-clients/assets/168455431/00b7750e-2f14-4abb-9117-30e7675029b7

This is because when the month was changed, the hook, `handleMonthChange` was modifying the variable that stored the date that the user had selected (`fieldProps`). This means that every time `handleMonthChange` was called, `fieldProps` would get redefined. 

The PR also fixes an issue where the time was reset to 12AM every time a user selected a new date.

## PR Checklist


MillanWangGadget (approved):
MANUAL TESTS
- 🍏 Changing the date no longer affects the time 
- 🍏 Switching between months no longer affects the time selection

Nice work. Merge after squashing this PR into 1 commit with a message similar to this PR title

PR #479, opened 2024-07-11
Added Gadget field type specific named exports for AutoTextInput

MillanWangGadget:
- **UPDATE**
  - Added individual Gadget field type named exports for `AutoTextInput` so that there is always an obvious match of which component to use


alanko0511 (approved):
:shipit: 👌

PR #481, opened 2024-07-11
Exclude custom object input in the form field list

alanko0511:
This PR excludes action inputs in which a type is an object, and the input name is not the model API identifier to be available for the AutoForm.

fixed ⬇️
~~Note that this PR doesn't handle excluding the array type because of a separate bug: GGT-6743~~

## PR Checklist


packages/react/src/auto/AutoForm.ts
```diff
@@ -58,7 +58,14 @@ export const useFormFields = (
   return useMemo(() => {
     if (!metadata) return [];
     const action = isActionMetadata(metadata) ? metadata.action : metadata;
-    const objectFields = action.inputFields.filter((field) => field.configuration.__typename === "GadgetObjectFieldConfig");
+    const objectFields =
+      metadata.__typename === "GadgetGlobalAction"
```
MillanWangGadget, line 62: It seems that the intention with this to check if the metadata for this is modelAction/GlobalAction. That overlaps a bit with the above `isActionMetadata(metadata)` call so you can add this check there and then store that result as a const

packages/react/src/auto/AutoForm.ts
```diff
@@ -58,7 +58,14 @@ export const useFormFields = (
...
+        ? []
+        : action.inputFields.filter(
+            (field) =>
+              field.configuration.__typename === "GadgetObjectFieldConfig" &&
+              (metadata.__typename === "GadgetModel" ? field.apiIdentifier === metadata.apiIdentifier : true)
```
MillanWangGadget, line 67: NIT: `metadata.__typename === "GadgetModel"` could be a const outside of the filter. Also seems to have similar intention to the above `metadata.__typename === "GadgetGlobalAction"` and `isActionMetadata(metadata)`

MillanWangGadget (approved):
MANUAL TESTS
- 🍏 Model actions properly exclude object custom params
- 🍏 Global actions properly exclude object custom params

Some minor code comments but good to go otherwise

danroberts (approved):

PR #485, opened 2024-07-11
Include array type in field validator

alanko0511:
Although we don't provide an array model field type, it's possible that people can define an array input type like this:
```js
export const params = {
  p_arr: {type: "array", items: {type: "string"}},
}
```
Right now, If people have an action input with an array type, it throws an error when rendering the form. This PR prevents throwing an error and instead provides a generic array type.

## PR Checklist


MillanWangGadget (approved):
Array OK 👍

PR #486, opened 2024-07-11
Actually turn on jest ESM mode

airhorns:
We adopted ESM for the packages in js-clients a long time ago and are now correctly building CJS and ESM artifacts for all the packages. At the time, I changed the jest config around and got it working, and assumed that that meant jest was now running our code as ESM. Turns out that was wrong because we weren't yet passing `--experimental-vm-modules`. It was still transpiling all our code to CJS and running it with `require` statements, which means we can't `import` any ESM-only dependencies. Regardless of if we want to do that, I think we should be running jest against the actual ESM version of the code, which is the version that will execute almost all the time, as vite serves the ESM version to the browser.

So, this turns on ESM mode! To do that, I had to fix a bunch of stuff.

First, ESM in Jest doesn't support module mocking anymore. Really annoying, but its just straight up not possible with jest to have top level imports mocked. You can make dynamic imports mocked, but not top level. Instead of contorting all our code to do dynamic imports of all the stuff we want to mock, I just added a one more override option to the GadgetConnection object that lets us inject the one thing we need to, and deleted one low value test relying on a mock. Seems ok to me.

Second, jest no longer injects the `jest` object as a global, you now need to import it in ESM mode. Importing it changed the types of a lot of stuff somehow such that all the mocks got a lot pickier, so a lot of the changes are to appease that.

And then there was a bunch of other small changes like polyfilling in `__dirname` and some other business.

[no-changelog-required]


package.json
```diff
@@ -16,7 +16,7 @@
     "watch": "run-p --print-label watch:*",
     "watch:client": "pnpm --filter=@gadgetinc/api-client-core watch",
     "watch:react": "pnpm --filter=@gadgetinc/react watch",
-    "test": "NODE_OPTIONS=\"--max-old-space-size=4096 --stack-trace-limit=1000\" jest",
+    "test": "NODE_OPTIONS=\"--max-old-space-size=4096 --stack-trace-limit=1000 --experimental-vm-modules\" TZ=UTC jest",
```
airhorns, line 19: this actually turns it on

package.json
```diff
@@ -46,7 +46,7 @@
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
     "jest-junit": "^16.0.0",
-    "lodash": "^4.17.21",
+    "lodash-es": "^4.17.21",
```
airhorns, line 49: we only use this in tests and this one works in naked ESM fine without more build plugins

packages/api-client-core/spec/GadgetConnection.browser.spec.ts
```diff
@@ -1,9 +1,6 @@
 /**
  * @jest-environment jsdom
  */
-jest.mock("graphql-ws", () => {
-  return { createClient: jest.fn(), CloseCode: { ConnectionAcknowledgementTimeout: 1, ConnectionInitialisationTimeout: 2 } };
-});
```
airhorns, line 6: replaced by dependency injection of the mock

packages/react-shopify-app-bridge/spec/Provider.spec.tsx
```diff
@@ -49,20 +48,17 @@ describe("GadgetProvider", () => {
         }),
     };
 
-    useAppBridgeMock = jest.spyOn(AppBridgeReact, "useAppBridge").mockImplementation(() => window.shopify);
```
airhorns, line 52: Wasnt necessary, thank the gods

danroberts (approved):

PR #487, opened 2024-07-11
Spread AutoForm cypress tests across different files to be easier to run

MillanWangGadget:
The cypress test was getting quite large and slow to run. This PR simply breaks up that big test file into a series of smaller ones


packages/react/cypress/component/auto/form/AutoFormDefaultValues.cy.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import { api } from "../../../support/api.js";
+import { describeForEachAutoAdapter } from "../../../support/auto.js";
+
+describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }) => {
```
alanko0511, line 5: Could you change the title to `AutoForm - default values` or something like that to indicate that this test runs stuff for default values (same for the global action test)

alanko0511 (approved):
:shipit: 👌

PR #490, opened 2024-07-12
test(autoform): GGT-6652 - tests asserting null value behavior

danroberts:
This adds a test that asserts that null values are preserved through an update load and submit lifecycle, but are turned into boolean values if the user interacts with a boolean checkbox, for example, when the form is first loaded, the value of isChecked is null, but if the user selects and then unselects the checkbox, it should submit a 'false' value. 

## PR Checklist


alanko0511 (approved):
:shipit: 👌

MillanWangGadget (approved):
Added tests make sense.
Checkboxes showing null and false in the same way is probably ok since I think that people probably do truthy checks on the record values instead of `boolRecordValue === false`

PR #493, opened 2024-07-12
Swap the file size validation message

alanko0511:
This PR fixes an issue when rendering the file input with size validation. It shows the "Accepts larger than xyz" when the validation is set to "min = undefined and max = xyz", which is incorrect.

## PR Checklist


MillanWangGadget (approved):
Nice message change 👍

PR #494, opened 2024-07-12
Update "required" validation error message to be more human readable

alanko0511:
Right now, when you have a "required" validation error on the client-side, it will show this error like this:
```
model.field is a required field
```

This PR updates the error message to stop showing the raw dot notation path. Instead, the message will be like this:
CleanShot 2024-07-12 at 15 30 19@2x

## PR Checklist


MillanWangGadget (approved):
Field name makes sense 👍

danroberts (approved):

PR #495, opened 2024-07-12
Fix field values not being sent when having a "include" prop in the AutoForm

alanko0511:
Right now, when you set fields to include using the `include` property in the AutoForm, none of the field values are sent. This PR fixes this issue. The details are in the comment below.

## PR Checklist

### DIFF
--- a/packages/react/src/auto/AutoForm.ts
+++ b/packages/react/src/auto/AutoForm.ts
@@ -138,15 +138,15 @@ export const useAutoForm = <
     resolver: useValidationResolver(metadata),
     send: () => {
       const fieldsToSend = fields
-        .map(({ path }) => path)
-        .filter((item) => {
+        .filter(({ metadata }) => {
           if (props.include) {
-            return props.include?.includes(item);
+            return props.include?.includes(metadata.apiIdentifier);
           } else if (props.exclude) {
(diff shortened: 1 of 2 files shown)


packages/react/src/auto/AutoForm.ts
```diff
@@ -131,15 +131,15 @@ export const useAutoForm = <
...
+        .filter(({ metadata }) => {
           if (props.include) {
-            return props.include?.includes(item);
+            return props.include?.includes(metadata.apiIdentifier);
           } else if (props.exclude) {
-            return !props.exclude?.includes(item);
+            return !props.exclude?.includes(metadata.apiIdentifier);
           }
           return true;
-        });
```
alanko0511, line 142: `path` is a dot-separated string with the format `model.field` (e.g., `widget.name`), but the strings inside `include` and `exclude` are expected to be field API identifiers.

So, when the `include` prop exists, the `includes(item)` is always `false` and none of the fields are included in the payload.

MillanWangGadget (approved):
MANUAL TESTS
- 🍏 Confirmed that create and update actions properly include the field values when using an `include`
- 🍏 Confirmed that create and update actions properly include the custom action param values when using an `include`

danroberts (approved):

PR #496, opened 2024-07-12
Set file value to null instead of undefined when clearing file

alanko0511:
Right now, in some rare situations, the file is not being removed when the user clicks the "delete file" button. This PR should fix this issue by setting the value to null instead of undefined.

## PR Checklist


MillanWangGadget (approved):
- 🍏 File deletion for update actions with findBy values works as expected in the env I had where it did not work

PR #497, opened 2024-07-13
Incorporated defaultDisplayField into the relationship option label creation system

MillanWangGadget:
- **UPDATE**
  - Previously, the relationship input components relied on having a string field called `name` on the related model
  - Now, we use `defaultDisplayField` from the metadata to determine the option label, which will eventually default to ID


alanko0511 (approved):
:shipit: 👌

PR #500, opened 2024-07-15
Don't run assertion check when there is an error fetching metadata

alanko0511:
Right now, when the field has an invalid default value defined in the Gadget editor or any other errors, it throws a `no model metadata found from Gadget API` assertion error and not being able to render the whole page. However, it's unclear to the user that something went wrong when fetching the metadata to render the form.

This PR fixes this issue by showing the error banner when it fails to fetch the metadata due to an error and does not render the submit button. Note that this PR only handles the case where the user doesn't put children inside the `<AutoForm/>` component since we can't really control what the user is doing.

## PR Checklist


MillanWangGadget (approved):
MANUAL TESTS
- 🍏 No fields are shown when the metadata can't load and the error message appears as expected in the banner

Looks good 👍

danroberts (approved):

PR #501, opened 2024-07-15
Fix relationship on blur pagination behavior

MillanWangGadget:
- **UPDATE**
  - Previously, the onBlur callback for the relationship selector components would remove all of the existing options to replace them with the options on the first page, and this made it impossible to get option labels for the selected IDs that were outside of the first page.
  - The original intention behind the removed callback was to save memory when the relationship dropdown pickers were blurred, but in retrospect, this would cause redundant re-fetching and the loss of option labels
  - Now, the callback is removed, and  the options will remain loaded in the relationship picker component after blurring


alanko0511 (approved):
:shipit: 👌

PR #503, opened 2024-07-15
Convert field value to Date in time picker

alanko0511:
We recently put the default field value into the form, but for the date time field type, the default value type is actually a string instead of a date, so it throws an error when there's a default value. This PR fixes it.

## PR Checklist


gabeb03 (approved):
Good catch!

PR #504, opened 2024-07-15
Introduce `onSuccess` callback in AutoForm

alanko0511:
This PR introduces an `onSuccess` callback in the `<AutoForm />` component. We already have the same callback in our `useActionForm` so we only need to expose it as part of the available property in the React component.

The default banner will not be shown when the `onSuccess` callback exists, so people can override the behaviour.

## PR Checklist

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoForm.cy.tsx
@@ -47,6 +47,27 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
     ensureFieldInputLabelsExist();
   });
 
+  it("onSuccess callback should return a record result after the form submission", () => {
+    const onSuccessSpy = cy.spy().as("onSuccessSpy");
+    cy.mountWithWrapper(<AutoForm action={api.widget.create} exclude={["gizmos"]} onSuccess={onSuccessSpy} />, wrapper);
+
+    cy.get(`input[name="widget.name"]`).type("test record");
+    cy.get(`input[name="widget.inventoryCount"]`).type("999");
+
+    cy.getSubmitButton().click();
+
+    cy.contains(`Saved Widget successfully`).should("not.exist");
+
+    // eslint-disable-next-line jest/valid-expect-in-promise
+    cy.get("@onSuccessSpy")
+      .should("have.been.called")
+      .then(() => {
+        const record = onSuccessSpy.getCalls()[0].args[0].toJSON();
+        expect(record).property("inventoryCount", 999);
+        expect(record).property("name", "test record");
+      });
+  });
+
   it("should show an error banner and not render a form when it fails to fetch metadata", () => {
     cy.intercept(
       {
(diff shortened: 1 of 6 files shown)


packages/react/src/auto/AutoForm.ts
```diff
@@ -33,6 +33,8 @@ export type AutoFormProps<
   submitLabel?: ReactNode;
   /** What to show the user once the form has been submitted successfully */
   successContent?: ReactNode;
+  /** Called when the form submission completes successfully on the backend */
+  onSuccess?: (record: UseActionFormHookStateData<ActionFunc>) => void;
```
alanko0511, line 37: This `UseActionFormHookStateData<ActionFunc>` type can make the record not become an `any` type. Example:
CleanShot 2024-07-15 at 12 45 12@2x

packages/react/src/auto/polaris/PolarisAutoForm.tsx
```diff
@@ -68,7 +68,7 @@ export const PolarisAutoForm = <
 
   const formContent = props.children ?? (
     <>
-      <PolarisSubmitResultBanner />
+      {!props.onSuccess && <PolarisSubmitResultBanner />}
```
MillanWangGadget, line 71: Perhaps this could be a prop into the banner as something like `showSuccessBanner` so that the failure banner can still be shown

MillanWangGadget (commented):
MANUAL TESTS
- 🍏 When an action is successfully complete, the callback is properly run with the record API response 
- 🟡 If the action does not run successfully, the failure banner is hidden and there is no corresponding callback for `onFailure` to know why an action may have failed.

packages/react/cypress/component/auto/form/AutoForm.cy.tsx
```diff
@@ -47,6 +47,44 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
...
+    const onSuccessSpy = cy.spy().as("onSuccessSpy");
+    cy.mountWithWrapper(<AutoForm action={api.widget.create} exclude={["gizmos"]} onSuccess={onSuccessSpy} />, wrapper);
+
+    ensureFieldInputLabelsExist();
+
+    cy.get(`input[name="widget.name"]`).type("test record");
+    cy.get(`input[name="widget.inventoryCount"]`).type("999");
+
+    cy.get(`input[id="Roles_Autocomplete_Textfield"]`).click().type("u"); // To search for "unauthenticated"
+    cy.get(`li[data-listbox-option-value="unauthenticated"]`).click();
```
danroberts, line 60: This could be a bit simpler if you limit the number of things that need to be selected in it -- i.e. just required fields, or name, or something similar

danroberts: I don't think we should really care much about roles here.

packages/react/src/auto/polaris/PolarisAutoForm.tsx
danroberts: It might be nice to split the ResultsBanner component into two different components, one for success and one for failure.

alanko0511:
I made two new components for the banner (on success and error). Let me know what you folks think.

MillanWangGadget (approved):
- 🍏 The error banner is still visible if a onSuccess is passed in 

New banner system works good 👍

danroberts (approved):

PR #505, opened 2024-07-15
fix(autoform): use polaris inline error messages

danroberts:
Previously, we were using the `helpText` property for some Polaris inputs, but to get the nicely styled Polaris errors, we need to pass the error message to the `error` prop. 

## PR Checklist


alanko0511 (approved):
🎩 
:shipit: 👌

MillanWangGadget (approved):
Error text looks good 👍

PR #509, opened 2024-07-15
fix(autoform): remove search icon prefix from multiselects

danroberts:
When using vertical content + a prefix in the polaris combo box, the vertical alignment of the prefix icon can't be controlled easily to keep it vertically aligned with the placeholder text i.e. "Search", so it ends up looking bad. 

In the polaris docs examples, they either use vertical content + no prefix or they use a prefix and append tag pills below the input. 

According to @mike-briggs we should prefer the vertical content approach + no prefix. 

## PR Checklist


alanko0511 (approved):
:shipit: 👌

MillanWangGadget (approved):
Prefix icons gone 👍 
The `search` text should be sufficient to communicate the searchability

PR #511, opened 2024-07-15
fix(autoform): throw if the component is passed a findBy prop for a c…

danroberts:
For incorrect configuration errors the best feedback channel for a developer is a thrown error that can bubble up and be handled for example by the vite error harness. 

This also throws if you include or exclude a required field, as the form will not be able to submit.

## PR Checklist


alanko0511 (approved):
🎩 
:shipit: 👌

gabeb03 (approved):
MANUAL TEST

🍏 When a required field is excluded from the `include` prop, the application throws a fatal error.

PR #513, opened 2024-07-15
Prevent untouched password fields getting sent null values on update actions

MillanWangGadget:
- **UPDATE**
  - Previously, untouched password fields for update actions would result in a null value being sent to the DB
    - Most other field types retrieve their record value from the DB and then send them back to the API through the form. Password fields are unique as their record value is not returned by the API, which means they can't send the same record value back to the API if untouched. That unknown record value then becomes null in the API request
    - Sending null to the API will result in the password getting cleared from the record. Big security problem

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoPasswordInput.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoPasswordInput.cy.tsx
@@ -1,0 +1,59 @@
+import React from "react";
+import { api } from "../../../support/api.js";
+import { describeForEachAutoAdapter } from "../../../support/auto.js";
+
+describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }) => {
+  beforeEach(() => {
+    cy.viewport("macbook-13");
+  });
+
+  const expectUpdateActionSubmissionVariables = (expectedQueryValue?: any) => {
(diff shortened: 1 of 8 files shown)


packages/react/cypress/component/auto/form/AutoPasswordInput.cy.tsx
```diff
@@ -0,0 +1,51 @@
...
+    cy.viewport("macbook-13");
+  });
+
+  const expectUpdateActionSubmissionVariables = (expectedQueryValue?: any) => {
+    cy.intercept({ method: "POST", url: `${api.connection.endpoint}?operation=updateUser` }, (req) => {
+      // eslint-disable-next-line
+      expect(req.body.variables).to.deep.equal(expectedQueryValue);
+
+      // The response content doesn't matter for the tests
+      req.reply({ data: { updateUser: { success: true, errors: null, x: {} } } });
```
alanko0511, line 16: It would be easier to read if this is not a one-liner, i.e.:
```js
req.reply({
  data: {
    updateUser: {
      // ...
```

alanko0511 (approved):
:shipit: 👌

packages/react/cypress/component/auto/form/AutoPasswordInput.cy.tsx
MillanWangGadget: Updated such that it is a multi-liner

danroberts (approved):

PR #514, opened 2024-07-15
GGT-6770 - Temporarily removed the MUI autocomponents 

gabeb03:
The MUI autocomponents are not ready to be used by alpha testers, so they have been commented out in this PR. 

## PR Checklist

### DIFF
--- a/packages/react/cypress/support/auto.tsx
+++ b/packages/react/cypress/support/auto.tsx
@@ -7,7 +7,7 @@ import type { ComponentType, ReactNode } from "react";
 import React from "react";
 import { FormProvider, useForm } from "react-hook-form";
 import type { AutoAdapter } from "../../src/auto/index.js";
-import * as MUIAdapter from "../../src/auto/mui/index.js";
+import * as MUIAdapter from "../../src/auto/mui/test-index.js";
 import * as PolarisAdapter from "../../src/auto/polaris/index.js";
 
 interface AutoSuiteConfig {
--- a/packages/react/spec/imports/esm-auto-import.mjs
+++ b/packages/react/spec/imports/esm-auto-import.mjs
@@ -1,2 +1,1 @@
 import { AutoForm as PolarisAutoForm } from "@gadgetinc/react/auto/polaris";
-import { AutoForm as MUIAutoForm } from "@gadgetinc/react/auto/mui";
--- a/packages/react/src/auto/mui/index.ts
+++ b/packages/react/src/auto/mui/index.ts
@@ -1,22 +1,2 @@
-export * from "./MUIAutoForm.js";
-export { MUIAutoForm as AutoForm } from "./MUIAutoForm.js";
-export { MUIAutoBooleanInput as AutoBooleanInput } from "./inputs/MUIAutoBooleanInput.js";
-export { MUIAutoDateTimePicker as DateTimePicker } from "./inputs/MUIAutoDateTimePicker.js";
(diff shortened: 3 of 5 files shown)


packages/react/spec/imports/esm-auto-import.mjs
```diff
@@ -1,2 +1,4 @@
 import { AutoForm as PolarisAutoForm } from "@gadgetinc/react/auto/polaris";
-import { AutoForm as MUIAutoForm } from "@gadgetinc/react/auto/mui";
+// TODO: Uncomment this line when MUI autocompoents are ready for pre-release
```
gabeb03, line 2: Is this the best way to temporarily remove the MUI components?
If so, should we create an Autocomponents ticket with detailed instructions on how undo the work done in this PR?

packages/react/src/auto/mui/index.ts
```diff
@@ -1,22 +1,23 @@
...
-  MUIAutoTextInput as AutoNumberInput,
-  MUIAutoTextInput as AutoStringInput,
-  MUIAutoTextInput as AutoTextInput,
-  MUIAutoTextInput as AutoUrlInput,
-} from "./inputs/MUIAutoTextInput.js";
-export { MUIAutoBelongsToInput as AutoBelongsToInput } from "./inputs/relationships/MUIAutoBelongsToInput.js";
-export { MUIAutoHasManyInput as AutoHasManyInput } from "./inputs/relationships/MUIAutoHasManyInput.js";
-export { MUIAutoSubmit as AutoSubmit } from "./submit/MUIAutoSubmit.js";
-export { MUISubmitResultBanner as SubmitResultBanner } from "./submit/MUISubmitResultBanner.js";
+// The MUI auto components have been temporarily removed. See GGT-6770.
```
MillanWangGadget, line 1: I don't think we need to include the ticket number here. That is only accessible to Gadget people

packages/react/cypress/support/auto.tsx
```diff
@@ -31,7 +31,8 @@ export const PolarisWrapper = ({ children }: { children: ReactNode }) => (
 );
 
 const suites: AutoSuiteConfig[] = [
-  { name: "MUI", adapter: MUIAdapter as any, wrapper: MUIWrapper },
+  // TODO: Uncomment this line when MUI autocompoents are ready for pre-release
+  // { name: "MUI", adapter: MUIAdapter as any, wrapper: MUIWrapper },
```
MillanWangGadget, line 35: Is it possible to keep the MUI tests around ?

gabeb03: Great point: all cypress tests have been added back; they reference a `test-index.ts`.

packages/react/spec/imports/esm-auto-import.mjs
danroberts: I think these lines should be removed altogether rather than left in a commented state.

danroberts: We can refer back to this PR when it comes time to reenable them, so a linear ticket with that seems good.

MillanWangGadget (approved):
Looks good to me! Changing the file name to include `test` is a good touch 👍 

I think Daniel is right about the commented out stuff. Commented out code has a tendency to get forgotten and out of sync, and might confuse the next person who works in this area

Good otherwise 👍

packages/react/spec/imports/esm-auto-import.mjs
gabeb03: The comments are removed, and the ticket has been created: GGT-6795

PR #515, opened 2024-07-15
Fix polaris role list options not loading initially until a search was conducted

MillanWangGadget:
- **UPDATE**
  - Sometimes role list fields would not have loaded in options until you searched/unsearched
    - This was because the options in `PolarisFixedOptionsCombobox` are initialized to the passed in options prop and those options were not loaded, thus leading to `PolarisFixedOptionsCombobox` getting initialized with an empty list of options 
  - Now, we won't load the role picker component until options exist, thus we won't be stuck with a initially empty list use state


danroberts (commented):
Can you add a regression test for this?

alanko0511 (approved):
🎩 tested on an app with this issue :shipit: 👌

danroberts (approved):

PR #517, opened 2024-07-16
Set step attribute value in Polaris number input based on metadata

alanko0511:
Right now, there is an issue where the step is set to 1 in the Polaris number input if no step value is set, which triggers the browser's default validation when pressing enter inside the number input.

This PR uses the `decimals` value from GadgetMeta's number field config to set the `step` value in the Polaris number input. If no decimal value is set, the `step` value becomes "any" to not have any steps.

"any" is a valid value in the standard HTML input (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step), but Polaris doesn't allow string value at the type level, so I cast the type as `any` instead.

## PR Checklist


danroberts (approved):

MillanWangGadget (approved):
Nice 👍

PR #522, opened 2024-07-16
Added key to autoforms to make the contents re-render upon actions and findBy value changes

MillanWangGadget:
CleanShot 2024-07-16 at 17 28 32

- **UPDATE**
  - Previously, dynamically changing the action or findBy value did not cause a re-render. This was a problem because you may end up thinking that you are editing the wrong value
  - Now, changing the action or the findBy value will force a re-render, causing all of the fields to be reset to the defaults or the corresponding record values


alanko0511 (approved):
:shipit: is it possible to have a test to prove this? 👌

danroberts (approved):

MillanWangGadget:
I'll fix forward with a test to prove this just so we can get the alpha version out first

PR #523, opened 2024-07-16
Prevent null values for text input components

MillanWangGadget:
- **UPDATE**
  - Previously, there was a console error when a null DB record value was passed to text fields. 
  - Now, an empty string will be sent to the text inputs when there is a null DB value
    - This does not trigger the controller's onChange function which will ensure that that value is still null if untouched when sent back to the API, thus leaving that record field unchanged

### DIFF
--- a/packages/react/src/auto/hooks/useStringInputController.tsx
+++ b/packages/react/src/auto/hooks/useStringInputController.tsx
@@ -37,6 +37,8 @@ export const useStringInputController = (
     ...rest,
   });
 
+  const { value, ...restOfFieldProperties } = fieldProperties;
+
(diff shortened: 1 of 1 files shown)


airhorns:
What do you think about pushing this into the controller itself so that all design systems don't need to apply this fix? Would fix my rich text issue of the same nature I think too!

alanko0511:
+1 for moving this logic into the controller because I feel like this behaviour also applies to most of the UI libraries.

MillanWangGadget:
Updated such that `value: value ?? ""` occurs in the hook instead of the individual design system implementations 😎

danroberts (approved):

alanko0511 (approved):

PR #528, opened 2024-07-17
Introduce table cell renderers for AutoTable

alanko0511:
This PR does several things related to the AutoTable component:
- Introduces individual table cell renders for each supported field type
- Update the `getValue` function to return the raw value from a record

## PR Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -40,8 +40,8 @@
     "prerelease": "gitpkg publish",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
-    "cypress:open": "cypress open",
-    "cypress:run": "cypress run"
+    "cypress:open": "TZ=UTC cypress open",
+    "cypress:run": "TZ=UTC cypress run"
   },
   "dependencies": {
     "@0no-co/graphql.web": "^1.0.4",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@gadget-client/app-with-no-user-model": "^1.10.0",
     "@gadget-client/bulk-actions-test": "^1.113.0",
     "@gadget-client/full-auth": "^1.9.0",
-    "@gadget-client/js-clients-test": "1.498.0-development.854",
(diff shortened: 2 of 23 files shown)


packages/react/package.json
```diff
@@ -40,8 +40,8 @@
     "prerelease": "gitpkg publish",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
-    "cypress:open": "cypress open",
-    "cypress:run": "cypress run"
+    "cypress:open": "TZ=UTC cypress open",
+    "cypress:run": "TZ=UTC cypress run"
```
alanko0511, line 44: Our CI uses UTC timezone, so setting the timezone here can prevent us from having a different result in development and CI.

MillanWangGadget (approved):
MANUAL TESTS
- 🍏 Columns containing really long values are truncated with 
  - 🟡 Tooltip on these would be nice, but fix-forwardable 
- 🍏 Boolean values properly show true/false/null
- 🍏 Password & encrypted string fields are never visible 
  - 🟡 It's possible to tell if an encrypted string is defined or not. I wonder if this is a security risk, and it should always show the *** placeholder regardless of if the record actually has a value or not
- 🍏 DateTime.includeTime fields show the whole date string in the local timezone
- 🟡 DateTime noIncludeTime fields still show the time
- 🍏 Files load properly wtih thumbnails
- 🟡 Enum and role list cell renderers will spill out their tags to cover their neighbouring cells
  - CleanShot 2024-07-17 at 17 08 20@2x
- 🍏 String-like fields appear as expected in the table

No show stoppers, I think all yellows can be fixed forward so we can get a lot of this foundational stuff into the repo

alanko0511:
> It's possible to tell if an encrypted string is defined or not. I wonder if this is a security risk

Yes, but based on the design, the encrypted string also doesn't show the value, and we might have a "click to open modal for record details" feature in the future.

danroberts (approved):

PR #530, opened 2024-07-17
Added autotable search

MillanWangGadget:
- **UPDATE**
  - Added searching functionality to the auto table
    - This is done by using the existing fuzzy search functionality in the useFindMany hook and incorporating this into the existing query system in the Polaris table
  - Updated the response type of the `useTable` hook
    - Keeping it as a `position dependant` array is annoying if you want a late-in-the-array value and don't want the middle-of-the-array values. You'd need to count the ignored properties

## PR Checklist


danroberts:
`pdated the response type of the useTable hook
Keeping it as a position dependant array is annoying if you want a late-in-the-array value and don't want the middle-of-the-array values. You'd need to count the ignored properties`

I'm not sure I follow 100%, but I think its good to keep the convention of the other hooks, which is usually [{variables}, refresh]. The {variables} which are destructured are not position dependent.

danroberts (commented):
As part of this, could you also see if you can find a nice way in the UI to 'exit' the search bar? Maybe on blur?

packages/react/spec/auto/PolarisAutoTable.stories.jsx
```diff
@@ -14,7 +15,7 @@ export default {
       // 👇 Make it configurable by reading the theme value from parameters
       return (
         <Provider api={api}>
-          <AppProvider>
+          <AppProvider i18n={translations}>
```
MillanWangGadget, line 18: This is required to get the text into the cancel button. Unfortunately, Shopify does not have a default i18n language which leads to empty strings when one is not explicitly given

MillanWangGadget:
> I'm not sure I follow 100%, but I think its good to keep the convention of the other hooks, which is usually [{variables}, refresh]. The {variables} which are destructured are not position dependent.

Updated such that the first array parameter has `{table:{rows, columns, etc}, search:{value, set,clear,}}` 


> As part of this, could you also see if you can find a nice way in the UI to 'exit' the search bar? Maybe on blur?

Updated to add a cancel button which also clears the search

packages/react/src/useTable.tsx
```diff
@@ -45,20 +45,35 @@ export interface TableOptions {
...
-        metadata: ModelMetadata;
-        fetching: boolean;
-        error?: ErrorWrapper;
-      }
-    | { columns: null; rows: null; data: null; metadata: null; page: TablePagination; fetching: boolean; error?: ErrorWrapper }
-  ),
-  (opts?: Partial<OperationContext>) => void
+  {
+    table: TableData<Data> & {
+      page: TablePagination;
```
danroberts, line 65: I don't think I'm 100% on board for grouping these under a table key -- I think it just makes it a bit more annoying to destructure them when you're consuming the hook.

I do think all the search props can be grouped together tho.

MillanWangGadget: Updated such that the search stuff is grouped together and the previous variables are untouched

danroberts (commented):
I also think we need to debounce the input a bit -- making a network request on each keypress seems aggressive -- do you see any opportunity to debounce by ~500ms ish?

MillanWangGadget:
> I also think we need to debounce the input a bit -- making a network request on each keypress seems aggressive -- do you see any opportunity to debounce by ~500ms ish?

Updated to use debounce system for the search. I set the timeout to 300ms cause I felt like (500ms+ the request time) felt a little bit slow. This is easily adjusted in the call to `debounce` though 👍

danroberts (approved):

MillanWangGadget:
Merging now because test failure is in an unrelated form matter

PR #531, opened 2024-07-17
Add an `<AutoButton/>` component for invoking an action with a simple button

airhorns:
- adds a new component to each design system for calling an action with a set of fixed variables, see Cypress tests for examples
 - DRYs up the shared logic around invoking the action into a hook
 - toasts a success or error message by default but can be overridden

[no-changelog-required]


packages/react/cypress/support/component.tsx
```diff
@@ -32,3 +32,48 @@ Cypress.Commands.add("mountWithWrapper", (component, Wrapper, options = {}) => {
 
   return mount(wrapped, options);
 });
+
+before(() => {
```
danroberts, line 36: just curious why you needed this?

danroberts (approved):

gabeb03 (approved):
Looks great 👍

PR #532, opened 2024-07-17
feat(autocomponents): extract useList from useTable for query controls

danroberts:
A first pass at extracting a less 'table-y' version of a hook to co-ordinate the searching, filtering, paging, etc of a list of records. 

I will commit this and then refactor the useTable hook to use it in a separate PR rather than cause a bunch of merge conflicts with the currently inflight work.

## PR Checklist

### DIFF
--- a/packages/react/spec/useList.spec.ts
+++ b/packages/react/spec/useList.spec.ts
@@ -1,0 +1,115 @@
+import { jest } from "@jest/globals";
+import { act, renderHook } from "@testing-library/react";
+
+let data: any = [{ id: 1 }, { id: 2 }];
+let fetching = false;
+let error = undefined as any;
+jest.unstable_mockModule("../src/useFindMany", () => ({
+  useFindMany: jest.fn<typeof useFindMany>().mockImplementation(() => [{ data, fetching, error, stale: false }, jest.fn()]),
+}));
+
+const { useFindMany } = await import("../src/useFindMany.js");
+const { useList } = await import("../src/useList.js");
+
+describe("useList", () => {
+  const manager = { findMany: jest.fn() } as any;
+  let result: any;
+
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


packages/react/spec/useList.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+let data: any = [{ id: 1 }, { id: 2 }];
+let fetching = false;
+let error = undefined as any;
+jest.unstable_mockModule("../src/useFindMany", () => ({
+  useFindMany: jest.fn<typeof useFindMany>().mockImplementation(() => [{ data, fetching, error, stale: false }, jest.fn()]),
+}));
+
+const { useFindMany } = await import("../src/useFindMany.js");
+const { useList } = await import("../src/useList.js");
```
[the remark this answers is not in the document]
danroberts: Yup -- with esm module mocking both the dependency and the class under test have to be imported asynchronously into the test in order for the mock to work.

packages/react/jest.setup.ts
```diff
@@ -1,8 +1,7 @@
-import { jest } from "@jest/globals";
 import "@testing-library/jest-dom/extend-expect";
 import crossFetch from "cross-fetch";
 
-jest.setTimeout(process.env.CI == "vscode-jest-tests" ? 20 * 60 * 1000 : 5 * 1000);
+//jest.setTimeout(process.env.CI == "vscode-jest-tests" ? 20 * 60 * 1000 : 5 * 1000);
```
alanko0511, line 4: Why comment out this line? 🤔

alanko0511 (approved):

packages/react/jest.setup.ts
danroberts: whoops that was an oversight. Good catch.

PR #533, opened 2024-07-18
add support for fragments to tiny-graphql-query-compiler

infiton:
As part of the upsert api work I need selections to work with union types; this adds support for specifying fields as inline fragments. I didn't do the work to support named fragments, but lmk if you think we should do that now

## PR Checklist


danroberts (approved):

infiton:
going to close this for now as @airhorns pointed out that you can do this by just setting to the key to `"... on Type"`; let's revisit this when we have a more pressing need

PR #534, opened 2024-07-18
GGT-6735 - Added sort and filter props to PolarisAutoTable

gabeb03:
Added props to `PolarisAutoTable` that are passed to the `useFindMany` hook that populates the table. 

Added pieces of state that track how the table is being sorted, and rerenders the table so the sort direction is shown. 

## PR Checklist

### DIFF
--- a/packages/react/src/auto/AutoTable.tsx
+++ b/packages/react/src/auto/AutoTable.tsx
@@ -19,4 +19,6 @@ export type AutoTableProps<
   live?: boolean;
   columns?: TableOptions["columns"];
   onClick?: (row: TableRow) => void;
+  sort?: Options["sort"];
+  filter?: Options["filter"];
 };
--- a/packages/react/src/auto/polaris/PolarisAutoTable.tsx
+++ b/packages/react/src/auto/polaris/PolarisAutoTable.tsx
@@ -1,4 +1,4 @@
-import type { FindManyFunction } from "@gadgetinc/api-client-core";
+import type { FindManyFunction, SortOrder } from "@gadgetinc/api-client-core";
 import type { IndexTableProps } from "@shopify/polaris";
 import {
   Banner,
@@ -12,9 +12,9 @@ import {
   useSetIndexFiltersMode,
 } from "@shopify/polaris";
 import pluralize from "pluralize";
-import React, { useCallback, useMemo } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import { useTable } from "../../useTable.js";
-import type { TableRow } from "../../useTableUtils/types.js";
+import type { TableColumn, TableRow } from "../../useTableUtils/types.js";
 import type { ColumnValueType, OptionsType } from "../../utils.js";
 import type { AutoTableProps } from "../AutoTable.js";
 import { useTableBulkActions } from "../hooks/useTableBulkActions.js";
@@ -35,6 +35,22 @@ const PolarisSkeletonTable = (props: { columns: number }) => {
     />
   );
 };
+const gadgetToPolarisDirection = (direction?: SortOrder) => {
+  if (direction === "Ascending") {
+    return "ascending";
+  } else if (direction === "Descending") {
+    return "descending";
+  }
+  return undefined;
+};
+
+const polarisToGadgetDirection = (direction: "ascending" | "descending"): SortOrder => {
+  return direction === "ascending" ? "Ascending" : "Descending";
+};
+
+const getColumnIndex = (columns: TableColumn[], apiIdentifier: string | undefined) => {
+  return columns.findIndex((column) => column.apiIdentifier === apiIdentifier);
+};
 
 /**
  * Renders a table of records from the backend automatically for a given model using Polaris
@@ -49,7 +65,7 @@ export const PolarisAutoTable = <
 ) => {
   const { onClick } = props;
 
-  const [{ rows, columns, metadata, fetching, error, page, search, selection }, refresh] = useTable<
+  const [{ rows, columns, metadata, fetching, error, page, search, sort, selection }, refresh] = useTable<
     GivenOptions,
     SchemaT,
     FinderFunction,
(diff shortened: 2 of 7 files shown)


alanko0511 (commented):
The type for the `sort` property is not correct.
It has to be like this:
```ts
{
  [key: string]: "Descending" | "Ascending"
}
```
https://docs.gadget.dev/api/example-app/development/sorting-and-filtering

packages/react/src/auto/AutoTable.tsx
```diff
@@ -15,4 +15,6 @@ export type AutoTableProps<
   pageSize?: number;
   initialCursor?: string;
   initialDirection?: string;
+  sort?: { [key: string]: "Ascending" | "Descending" };
+  filter?: { [key: string]: { [key: string]: any } };
```
alanko0511, line 19: Just found that you can also do this:
```ts
  sort?: Options["sort"];
  filter?: Options["filter"];
```

The `Options` is referenced from the function, and it will have a full type like this:
CleanShot 2024-07-18 at 15 05 36@2x

alanko0511 (approved):

gabeb03:
Expanded scope to include GGT-6810

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -46,15 +59,48 @@ export const PolarisAutoTable = <
...
+    select: props.select,
+    columns: props.columns,
+    pageSize: props.pageSize,
+    live: props.live,
+  } as any);
+
+  const [sortColumnIndex, setSortColumnIndex] = useState<number | undefined>(
+    columns ? getColumnIndex(columns, props.sort ? Object.keys(props.sort)[0] : undefined) : undefined
+  );
+  const [setColumnIndexToDefault, setSetColumnIndexToDefault] = useState(true);
```
alanko0511, line 77: This variable is really confusing to the people who read code that `setColumnIndexToDefault` is actually a state value, not the function that modifies the value. Maybe `defaultColumnIndex` and `setDefaultColumnIndex`?

Also, some states seem to be for triggering the `useEffect` and updating another state. Is it possible to instead use hooks like `useMemo` or `useCallback` to reduce the number of states needed?

packages/react/src/useTable.tsx
```diff
@@ -148,6 +151,14 @@ export const useTable = <
     [fieldSelectionMap, metadata?.fields, options]
   );
 
+  const sortColumn = useCallback((colName: string, direction: "Ascending" | "Descending" | undefined) => {
```
alanko0511, line 154: You don't actually need to include the `undefined` in the type. You can just do something like this to let `direction` become an optional instead of a required parameter that passes in an `undefined`:

```ts
const sortColumn = useCallback((colName: string, direction?: "Ascending" | "Descending") // look at the question mark
```

(⬆️ same for places that include the `undefined` in the type)

Also, it seems that we're using the `"Ascending" | "Descending"` type in quite a few places. How about making it a separate type and sharing it in other areas?

alanko0511 (commented):
This time it changes quite a lot of stuff. I think we should add some tests to prove that.

packages/react/src/auto/polaris/PolarisAutoTable.tsx
gabeb03: The `useEffect` is so that when `column` loads, the filter icon updates, previously, it was not always pointing to the correct index because column does some time to load.

From what I understand, `useMemo` or `useCallback` don't update when `column` updates. 

I limited the dependancy array so useEffect is only triggered when `columns` updates.

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -46,15 +59,51 @@ export const PolarisAutoTable = <
...
+    live: props.live,
+    sort: props.sort,
+    filter: props.filter,
+  } as any);
+
+  const [sortColumnIndex, setSortColumnIndex] = useState<number | undefined>(
+    columns ? getColumnIndex(columns, props.sort ? Object.keys(props.sort)[0] : undefined) : undefined
+  );
+  const [defaultColumnIndex, setDefaultColumnIndex] = useState(true);
+  const [sortDirection, setSortDirection] = useState<"ascending" | "descending" | undefined>(
```
danroberts, line 80: Instead of doing the 'Ascending' => 'ascending' mapping here in the state, I think we could just do it right before we pass the value to polaris

like: defaultSortDirection={gadgetToPolarisSortKey(defaultSortDirection)}

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -46,15 +59,51 @@ export const PolarisAutoTable = <
...
+    getPresetSortDirection(props.sort ? Object.values(props.sort)[0] : undefined)
   );
+  const [disableSort, setDisableSort] = useState<boolean>(false);
+  const defaultSortDirection = getPresetSortDirection(props.sort ? Object.values(props.sort)[0] : "Descending");
+
+  useEffect(() => {
+    if (!sortColumnIndex && defaultColumnIndex) {
+      setSortColumnIndex(columns ? getColumnIndex(columns, props.sort ? Object.keys(props.sort)[0] : undefined) : undefined);
```
danroberts, line 88: I'm not sure I follow the need for this useEffect. when the columns change, we want to reset the sort back to the default?

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -32,6 +33,18 @@ const PolarisSkeletonTable = (props: { columns: number }) => {
     />
   );
 };
+const getPresetSortDirection = (direction?: SortOrder) => {
```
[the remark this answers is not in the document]
gabeb03: That's right

packages/react/src/auto/polaris/PolarisAutoTable.tsx
gabeb03: The `sortColumnIndex` state determines where the sort icon is: 
image
`columns` take some time to load, so when sortColumnIndex is initialized, `columns` is undefined; later on it is initialized, but sortColumnIndex was not being updated.

danroberts (approved):

alanko0511 (approved):
:shipit: It looks cleaner now 👌

PR #535, opened 2024-07-18
Introduce `columns` property to select fields

alanko0511:
This PR introduces a new `useTable` property called `columns` that allows you to select a list of fields to include in a table. Right now, this property only accepts strings (field API identifiers) as a value inside the array. In the future, we will also take custom column renderers and pick a field for a relationship.

## PR Checklist

### DIFF
--- a/packages/react/src/metadata.tsx
+++ b/packages/react/src/metadata.tsx
@@ -307,7 +307,7 @@ export const useActionMetadata = (actionFunction: ActionFunction<any, any, any, 
 /**
  * @internal
  */
-export const filterFieldList = (
+export const filterAutoFormFieldList = (
   fields: FieldMetadata[] | undefined,
   options?: { include?: string[]; exclude?: string[] }
 ): FieldMetadata[] => {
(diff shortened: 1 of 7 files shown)


MillanWangGadget:
Can you add a storybook entry showing off this functionality?

packages/react/src/metadata.tsx
```diff
@@ -385,6 +385,54 @@ const acceptedFieldTypes = new Set([
...
+  if (options?.include) {
+    // When including fields, the order will match the order of the `include` array
+    subset = [];
+    const includes = new Set(options.include);
+
+    for (const includedFieldApiId of Array.from(includes)) {
+      const metadataField = fields.find((field) => field.apiIdentifier === includedFieldApiId);
+      if (metadataField) {
+        subset.push(metadataField);
+      }
```
MillanWangGadget, line 404: CleanShot 2024-07-18 at 15 28 32@2x

When given a column name that does not exist, we get nothing rendered. Perhaps we should have a `else throw Error()` here so that the user knows why there is nothing rendered

alanko0511:
> Can you add a storybook entry showing off this functionality?

Add

MillanWangGadget (approved):
One minor comment but good otherwise 👍

packages/react/src/metadata.tsx
alanko0511: It's because it makes the GraphQL request to respond to error 400.
CleanShot 2024-07-18 at 15 32 09@2x

I think maybe instead we should show a banner similar to the form.

MillanWangGadget: Good point, a banner would be better so that the user can have a dynamically changing list of columns without stress of getting a whole exception

danroberts (approved):

PR #536, opened 2024-07-18
changes to api client core to support the upsert api

infiton:
These changes will support introducing the upsert api to the js client. Mainly adds support for the case where an actions return type is a Union type.

## PR Checklist

### DIFF
--- a/packages/api-client-core/spec/operationBuilders.spec.ts
+++ b/packages/api-client-core/spec/operationBuilders.spec.ts
@@ -667,6 +667,119 @@ describe("operation builders", () => {
       `);
     });
 
+    test("actionOperation should build a mutation query for an object hasReturnType", async () => {
+      expect(
+        actionOperation(
+          "upsertWidget",
+          { __typename: true, id: true, name: true, eventAt: true },
+          "widget",
+          "widget",
+          {},
+          null,
+          null,
+          null,
+          {
+            "... on CreatePostResult": { hasReturnType: false },
+            "... on UpdatePostResult": { hasReturnType: false },
+          }
+        )
+      ).toMatchInlineSnapshot(`
+        {
+          "query": "mutation upsertWidget {
+          upsertWidget {
+            success
+            errors {
+              message
+              code
+              ... on InvalidRecordError {
+                validationErrors {
+                  message
+                  apiIdentifier
+                }
+              }
+            }
+            ... on CreatePostResult {
+              __typename
+              widget {
+                __typename
+                id
+                name
+                eventAt
+              }
+            }
+            ... on UpdatePostResult {
+              __typename
+              widget {
+                __typename
+                id
+                name
+                eventAt
+              }
+            }
+          }
+          gadgetMeta {
+            hydrations(modelName: "widget")
+          }
+        }",
+          "variables": {},
+        }
+      `);
+    });
+
+    test("actionOperation should build a mutation query for an object hasReturnType with an inner return type", async () => {
+      expect(
+        actionOperation(
+          "upsertWidget",
+          { __typename: true, id: true, name: true, eventAt: true },
+          "widget",
+          "widget",
+          {},
+          null,
+          null,
+          null,
+          {
+            "... on CreatePostResult": { hasReturnType: false },
+            "... on UpdatePostResult": { hasReturnType: true },
+          }
+        )
+      ).toMatchInlineSnapshot(`
+        {
+          "query": "mutation upsertWidget {
+          upsertWidget {
+            success
+            errors {
+              message
+              code
+              ... on InvalidRecordError {
+                validationErrors {
+                  message
+                  apiIdentifier
+                }
+              }
+            }
+            ... on CreatePostResult {
+              __typename
+              widget {
+                __typename
+                id
+                name
+                eventAt
+              }
+            }
+            ... on UpdatePostResult {
+              __typename
+              result
+            }
+          }
+          gadgetMeta {
+            hydrations(modelName: "widget")
+          }
+        }",
+          "variables": {},
+        }
+      `);
+    });
+
     test("actionOperation should build a mutation query for an action that has a return type", () => {
       expect(
         actionOperation(
@@ -706,13 +819,55 @@ describe("operation builders", () => {
       `);
     });
 
+    test("actionOperation should build a bulk mutation query", () => {
+      expect(
+        actionOperation(
+          "bulkCreateWidgets",
+          { __typename: true, id: true, state: true },
+          "widget",
+          "widgets",
+          {},
+          { select: { id: true } },
+          undefined,
+          true,
+          false
+        )
+      ).toMatchInlineSnapshot(`
+        {
+          "query": "mutation bulkCreateWidgets {
+          bulkCreateWidgets {
+            success
+            errors {
+              message
+              code
+              ... on InvalidRecordError {
+                validationErrors {
+                  message
+                  apiIdentifier
+                }
+              }
+            }
+            widgets {
+              id
+              __typename
+            }
+          }
+          gadgetMeta {
+            hydrations(modelName: "widget")
+          }
+        }",
+          "variables": {},
+        }
+      `);
+    });
+
     test("actionOperation should build a bulk mutation query for an action that has a return type", () => {
       expect(
         actionOperation(
           "bulkCreateWidgets",
           { __typename: true, id: true, state: true },
           "widget",
-          "widget",
+          "widgets",
           {},
           { select: { id: true } },
           undefined,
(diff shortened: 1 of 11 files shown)


packages/api-client-core/src/GadgetFunctions.ts
```diff
@@ -100,12 +100,15 @@ export interface BulkActionWithInputs<OptionsT, VariablesT> {
...
 
+export type UnionResultType = Record<string, { hasReturnType: boolean }>;
+
 export interface ActionFunctionMetadata<OptionsT, VariablesT, SelectionT, SchemaT, DefaultsT, IsBulk> {
   type: "action";
   operationName: string;
   operationReturnType?: string;
   namespace: string | string[] | null;
   modelApiIdentifier: string;
+  operatesWithRecordIdentity?: boolean;
```
infiton, line 111: I think to be on the safe side we should make this optional and so that we don't need to add this property to pre version 1.1 apps; but let me know if you think we should make it required

packages/api-client-core/src/support.ts
```diff
@@ -588,6 +588,9 @@ export const disambiguateActionVariables = (action: AnyActionFunction, variables
 
   let newVariables: Record<string, any>;
 
+  // for backwards compatibilty, actions without the operatesWithRecordIdentity metadata should extract the id from the variables
+  const shouldExtractId = action.operatesWithRecordIdentity ?? true;
```
infiton, line 592: if we leave `operatesWithRecordIdentity` as optional, then this will make sure that by default we extract id as we currently do; I think before upsert this is fine

packages/api-client-core/src/GadgetFunctions.ts
airhorns: I think best to let it be optional since we can't guarantee the client has been regenerated it to offer it!

packages/api-client-core/src/support.ts
thegedge: Game for adding a new property, but could we also look at `action.variables` to see if there's an id arg at the root and avoid juggling a new prop? i.e.,

```
const shouldExtractId = "id" in action.variables;
```

infiton: I think this doesn't work for bulk actions where the top level variables are `inputs`; but we still disambiguate in bulk

packages/api-client-core/src/operationRunners.ts
```diff
@@ -307,15 +314,19 @@ const processActionResponse = <Shape extends RecordShape = any>(
...
   // Delete actions have a null selection. We do an early return for this because `hydrateRecordArray` will fail
   // if there's nothing at `mutationResult[modelSelectionField]`, but the caller isn't expecting a return (void).
   if (defaultSelection == null) return;
   if (!hasReturnType) {
     return hydrateRecord<Shape>(response, record[modelSelectionField]);
-  } else {
+  } else if (typeof hasReturnType == "boolean") {
     return record.result;
+  } else {
+    const innerReturnType = Object.entries(hasReturnType).find(([key]) => key == `... on ${record.__typename}`);
```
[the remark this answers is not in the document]
infiton: 👍 thanks much nicer

packages/api-client-core/src/operationBuilders.ts
```diff
@@ -195,7 +224,7 @@ export const backgroundActionResultOperation = <Action extends AnyActionFunction
       const selection = options?.select || action.defaultSelection;
 
       fields = {
-        [`... on ${resultType}`]: actionResultFieldSelection(action.modelApiIdentifier, selection, action.hasReturnType),
+        [`... on ${resultType}`]: actionResultFieldSelection(action.modelApiIdentifier, selection),
```
thegedge, line 227: Just to double check: this shouldn't be:

```ts
actionResultFieldSelection(action.modelApiIdentifier, selection, undefined, action.hasReturnType)
```

?

packages/api-client-core/src/operationRunners.ts
```diff
@@ -302,23 +307,54 @@ export const actionRunner: ActionRunner = async <Shape extends RecordShape = any
...
+  if (!hasReturnType) {
+    return hydrateRecordArray<Shape>(response, records[modelSelectionField]);
+  } else if (typeof hasReturnType == "boolean") {
+    return records.results;
+  } else {
+    return Object.keys(hasReturnType).flatMap((returnTypeField) => {
+      const results = records[returnTypeField];
+
+      return Array.isArray(results) ? results : [];
+    });
```
thegedge, line 327: Might need a small refresher here.

I thought the object-y `hasReturnType` would be for the upsert case, where we have something like:

```gql
... on CreateResult { thing { … } }
... on UpdateResult { thing { … } }
```

And the `hasReturnType` would be something like:

```ts
{
  "... on CreateResult": { hasReturnType: false },
  "... on UpdateResult": { hasReturnType: false },
}
```

Is that right? If so, those keys aren't actual fields on the results. I think you'd have to inspect the `__typename` to know how to process the result.

packages/api-client-core/src/operationBuilders.ts
infiton: yah good catch, updated to use the new call signature will all the metadata about the action

packages/api-client-core/src/operationRunners.ts
infiton: updated to inspect the `__typenames`; for the bulk case it's a little more annoying; if there is a return type the query looks like:

```
mutation {
  bulkUpsertPosts(inputs: [{}, {post: {id: "56"}}]) {
  	posts {
      __typename
      ... on UpsertPostReturnType {
        __typename
        result
      }
      ... on Post {
        __typename
        id
      }
    }
  }
}
```

thegedge: 🚀 yeah, you have this extra layer to work through which is annoying

packages/api-client-core/spec/operationRunners.spec.ts
```diff
@@ -751,6 +853,95 @@ describe("operationRunners", () => {
...
+          },
+        },
+        stale: false,
+        hasNext: false,
+      });
+
+      const results = await promise;
+
+      expect(results[0]).toEqual({ hello: "world" });
+      expect(results[1]).toEqual({ hello: "other world" });
```
airhorns, line 893:
```suggestion
            results: [{ hello: "world" }, "some scalar"],
          },
        },
        stale: false,
        hasNext: false,
      });

      const results = await promise;

      expect(results[0]).toEqual({ hello: "world" });
      expect(results[1]).toEqual("some scalar");
```

just to exercise the code path to make sure folks that return non-objects as their return type will be ok

airhorns (approved):

packages/api-client-core/spec/operationBuilders.spec.ts
```diff
@@ -667,6 +667,119 @@ describe("operation builders", () => {
...
+              message
+              code
+              ... on InvalidRecordError {
+                validationErrors {
+                  message
+                  apiIdentifier
+                }
+              }
+            }
+            ... on CreatePostResult {
```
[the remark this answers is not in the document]
infiton: since these are the type of the action return type now they would change when the action api identifier changes;

I'll make sure to test this on the api side PR

packages/api-client-core/spec/operationRunners.spec.ts
infiton: 👍

packages/api-client-core/src/operationRunners.ts
```diff
@@ -302,21 +307,69 @@ export const actionRunner: ActionRunner = async <Shape extends RecordShape = any
...
+  response: any,
+  records: any,
+  modelSelectionField: string,
+  hasReturnType?: HasReturnType | null
+) => {
+  if (defaultSelection == null) return [];
+  if (!hasReturnType) {
+    return hydrateRecordArray<Shape>(response, records[modelSelectionField]);
+  } else if (typeof hasReturnType == "boolean") {
+    return records.results;
```
thegedge, line 321: If `hasReturnType === false`, should we also be hydrating?

packages/api-client-core/src/operationRunners.ts
```diff
@@ -302,21 +307,69 @@ export const actionRunner: ActionRunner = async <Shape extends RecordShape = any
...
+  } else {
+    return Object.entries(hasReturnType).flatMap(([returnTypeField, innerHasReturnType]) => {
+      const results = records[returnTypeField];
```
thegedge, line 324: Similar to my [other comment](https://github.com/gadget-inc/js-clients/pull/536#discussion_r1684984053), `hasReturnType` has this shape, right?

```js
{
  "... on CreateResult": { hasReturnType: false },
  "... on UpdateResult": { hasReturnType: false },
}
```

If so, I wouldn't expect any mapping over `hasReturnType` when it's an object, but rather would expect the code to look a bit more like this:

```js
return records.map((record) => {
  // if typename of record `hasReturnType`, use record as is
  // if typename of record `!hasReturnType`, hydrate
})
```

Basically, what you have on lines 330–339, over the result set in the payload.

packages/api-client-core/src/operationRunners.ts
infiton: yup; that's covered above by `if (!hasReturnType)`

packages/api-client-core/src/operationRunners.ts
infiton: in the non bulk case it looks like:

``` 
hasReturnType: {
  '... on CreatePostResult': { hasReturnType: true },
  '... on UpdatePostResult': { hasReturnType: false }
};
```

but in this case it looks like:

```
hasReturnType: {
          posts: {
            hasReturnType: {
              '... on Post': { select: true },
              '... on UpsertPostReturnType': { hasReturnType: true }
            }
          }
        };
```

infiton:
@thegedge going to ship and then write a bunch of tests api side before it gets added to any apps; since we hardcode the api-client-core version that ships with each generated client this can't get released into the wild until we bump the version gadget side

packages/api-client-core/src/operationRunners.ts
thegedge: Oh right! 🤦

packages/api-client-core/src/operationRunners.ts
thegedge: Ahhh, I thought it was always the former (wasn't able to keep all of the context in my head 😅). Good to know!

PR #537, opened 2024-07-18
Ensure required validation on richText fields only runs when the field has a required validation on it

MillanWangGadget:
- **UPDATE**
  - Previously, not having an input in a rich text field would block the request from sending due to a `required` validation being added regardless of if the field actually had a required validation on it or not
  - Now, the having an untouched markdown field will only run a required validation if the Gadget field has a required validation on it


alanko0511 (approved):
:shipit: 👌

PR #538, opened 2024-07-18
test(cypress): disable testing in real chrome browser

danroberts:
I think we've given chrome a good chance -- the flakiness isn't worth it. Disabling until we find a solution. 

## PR Checklist


MillanWangGadget (approved):
Back to electron in cypress makes sense to me for the time being with the current flake situation

airhorns (approved):

PR #539, opened 2024-07-18
Show banner inside the table when there is a GraphQL error

alanko0511:
This PR shows a banner when there's a GraphQL error. It will look like this:
CleanShot 2024-07-18 at 17 24 31@2x

## PR Checklist

### DIFF
--- a/packages/react/src/auto/polaris/PolarisAutoTable.tsx
+++ b/packages/react/src/auto/polaris/PolarisAutoTable.tsx
@@ -1,7 +1,9 @@
 import type { FindManyFunction } from "@gadgetinc/api-client-core";
 import type { IndexTableProps } from "@shopify/polaris";
 import {
+  Banner,
   BlockStack,
+  Box,
   DataTable,
   EmptySearchResult,
   IndexFilters,
@@ -64,7 +66,7 @@ export const PolarisAutoTable = <
     return { headings, sortable };
   }, [columns]);
 
-  if ((fetching && !rows) || !columns) {
+  if (!error && ((fetching && !rows) || !columns)) {
     return <PolarisSkeletonTable columns={polarisTableProps.headings.length} />;
   }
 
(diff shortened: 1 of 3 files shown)


packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -93,13 +95,24 @@ export const PolarisAutoTable = <
         onQueryChange={search.set}
         onQueryClear={search.clear}
```
MillanWangGadget, line 96: Can you add `disabled={!!error}` to the `<IndexFilters>` so that it's impossible to search with an active error? Otherwise searching would still send API requests that error

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -93,13 +95,24 @@ export const PolarisAutoTable = <
...
         resourceName={resourceName}
         emptyState={<EmptySearchResult title={`No ${resourceName.plural} yet`} description={""} withIllustration />}
         loading={fetching}
         hasMoreItems={page.hasNextPage}
-        itemCount={rows?.length ?? 0}
+        itemCount={
+          error
+            ? 1 // Don't show the empty state if there's an error
+            : rows?.length ?? 0
+        }
```
MillanWangGadget, line 115: This really just adds the disabled pagination buttons when we are errored. What do you think about something like 
```
error ? <ErrorBanner/> : <IndexTable/>
```
so that error behavior doesn't need to be added to the index table?

MillanWangGadget (approved):
Some minor code comments but good otherwise 👍

packages/react/src/auto/polaris/PolarisAutoTable.tsx
alanko0511: It's not entirely the same because with your code, it will look like this:
CleanShot 2024-07-19 at 09 47 23@2x

It doesn't feel like it's inside the table.

danroberts (approved):

PR #540, opened 2024-07-18
Introduce `onClick` callback in AutoTable

alanko0511:
This PR introduces an `onClick` callback in the AutoTable component for users to customise the operation when clicking on a table row. The callback will provide the row record in the parameter.

## PR Checklist

### DIFF
--- a/packages/react/src/auto/AutoTable.tsx
+++ b/packages/react/src/auto/AutoTable.tsx
@@ -1,6 +1,6 @@
 import type { FindManyFunction } from "@gadgetinc/api-client-core";
 import type { TableOptions } from "../useTable.js";
-import type { OptionsType } from "../utils.js";
+import type { ColumnValueType, OptionsType } from "../utils.js";
 
 /**
  * The props to pass to an AutoTable. Includes both the Gadget-land props as well as the adapter specific props.
@@ -17,4 +17,5 @@ export type AutoTableProps<
   initialCursor?: string;
   initialDirection?: string;
   columns?: TableOptions["columns"];
+  onClick?: (row: Record<string, ColumnValueType>) => void;
 };
--- a/packages/react/src/auto/polaris/PolarisAutoTable.tsx
+++ b/packages/react/src/auto/polaris/PolarisAutoTable.tsx
@@ -12,9 +12,9 @@ import {
   useSetIndexFiltersMode,
 } from "@shopify/polaris";
 import pluralize from "pluralize";
-import React, { useMemo } from "react";
+import React, { useCallback, useMemo } from "react";
 import { useTable } from "../../useTable.js";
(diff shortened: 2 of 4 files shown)


packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -110,7 +110,18 @@ export const PolarisAutoTable = <
...
           rows.map((row, index) => (
-            <IndexTable.Row key={row.id as string} id={row.id as string} position={index}>
+            <IndexTable.Row
+              key={row.id as string}
+              id={row.id as string}
+              position={index}
+              onClick={
+                props.onClickRow
+                  ? () => {
+                      props.onClickRow?.(row);
```
MillanWangGadget, line 120: this can just be `() => props.onClickRow?.(row)`

MillanWangGadget (approved):
One minor code comment but good otherwise 👍 
Nice storybook entry 👍

packages/react/src/auto/polaris/PolarisAutoTable.tsx
alanko0511: This will override the default Polaris behaviour even if the `onClickRow` is undefined:
CleanShot 2024-07-19 at 10 03 31@2x
https://polaris.shopify.com/components/tables/index-table#props

MillanWangGadget: Makes sense. Didn't realize 👍

packages/react/src/auto/AutoTable.tsx
```diff
@@ -17,4 +17,5 @@ export type AutoTableProps<
   initialCursor?: string;
   initialDirection?: string;
   columns?: TableOptions["columns"];
+  onClickRow?: (row: Record<string, ColumnValueType>) => void;
```
danroberts, line 20: I don't love the 'onClickRow' name -- Let's go with onClick for now.

alanko0511: Renamed.

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -124,7 +124,18 @@ export const PolarisAutoTable = <
...
           columns &&
           rows.map((row, index) => (
-            <IndexTable.Row key={row.id as string} id={row.id as string} position={index}>
+            <IndexTable.Row
+              key={row.id as string}
+              id={row.id as string}
+              position={index}
+              onClick={
+                props.onClick
+                  ? () => {
```
danroberts, line 133: This has the potential to create unnecessary re-renders since it will allocate a new anonymous function each time. You could pull it up to a useCallback() variable maybe?

danroberts (approved):

PR #543, opened 2024-07-19
Add ID field in AutoTable

alanko0511:
This PR adds the missing ID field to the AutoTable:
CleanShot 2024-07-19 at 15 55 51@2x

## PR Checklist

### DIFF
--- a/packages/react/spec/auto/hooks/useTable.spec.tsx
+++ b/packages/react/spec/auto/hooks/useTable.spec.tsx
@@ -49,6 +49,12 @@ describe("useTable hook", () => {
       })
     ).toEqual([
       {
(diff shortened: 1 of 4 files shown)


MillanWangGadget (commented):
MANUAL TESTING 
- 🍏 IDs appear as expected in the table
- 🟡 If `columns=['fieldNames"]` is used, the id is not in the record at all. 
  - Not having ID here will make it hard to know which records we should call bulk actions on if we select them
  - Could you make the ID always get retrieved and then only render it if it's given in the `columns`

MillanWangGadget (approved):
- 🍏 IDs are now in the record even if they are not included in `columns`

PR #545, opened 2024-07-19
Truncate tags when overflow in AutoTable

alanko0511:
As the title suggests, this PR truncates the tags when they overflow in the UI or there are more than five tags.
CleanShot 2024-07-22 at 10 16 08@2x

## PR Checklist

### DIFF
--- a/packages/react/src/auto/polaris/tableCells/PolarisAutoTableTagCell.tsx
+++ b/packages/react/src/auto/polaris/tableCells/PolarisAutoTableTagCell.tsx
@@ -1,32 +1,59 @@
-import { InlineStack, Tag } from "@shopify/polaris";
-import React, { useMemo } from "react";
+import { InlineStack, Tag, Text } from "@shopify/polaris";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import type { RoleAssignmentsValueType } from "../../../utils.js";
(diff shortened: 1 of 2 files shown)


packages/react/src/auto/polaris/tableCells/PolarisAutoTableTagCell.tsx
```diff
@@ -1,32 +1,61 @@
...
+      <div
+        style={{
+          width: showMore ? "calc(100% - 20px)" : "auto",
+          overflowX: "hidden",
+        }}
+        ref={ref}
+      >
+        <InlineStack gap="100" wrap={false}>
+          {tags.map((tag) => (
+            <Tag key={tag} accessibilityLabel={`tag-${tag}`}>
```
MillanWangGadget, line 51: `accessibilityLabel` is the tooltip and when it is undefined, it will be the complete untruncated tag content. Could you remove it from here?

MillanWangGadget (commented):
MANUAL TESTS
- 🍏 Really long tags are truncated inside the tag
- 🍏 Having many tags will truncate outside of the tags for the tags that can't be shown 
- 🍏 Having a lot of really longs tags will properly truncate both

One small code comment but good otherwise 👍

MillanWangGadget (approved):

danroberts (approved):

PR #548, opened 2024-07-21
use consistent call signature for upsert between public and internal api

infiton:
In writing api tests for the internal upsert I realized I put an inconsistent call signature for the upsert between internal and public api. The fewer surprises the better so this moves the `on` param into the record payload of the internal call and pulls it out if its passed

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -490,14 +490,15 @@ export class InternalModelManager<Shape extends RecordShape = RecordData> {
    * // upserts post with slug "new-post" in the database
    * // if a post with slug "new-post" exists, it will be updated
    * // if a post with slug "new-post" does not exist, it will be created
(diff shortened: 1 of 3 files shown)


packages/api-client-core/src/InternalModelManager.ts
```diff
@@ -484,20 +484,21 @@ export class InternalModelManager<Shape extends RecordShape = RecordData> {
    * // upserts post with id 10 in the database
    * // if a post with id 10 exists, it will be updated
    * // if a post with id 10 does not exist, it will be created
-   * const post = await api.internal.post.upsert({ id: "10", title: "A new post title" });
+   * const post = await api.internal.post.upsert({ post: {id: "10", title: "A new post title"} });
```
airhorns, line 487: I think it might make sense to document the simple form cause that form will work, but then show this one below?

airhorns (approved):
LGTM!

packages/api-client-core/src/InternalModelManager.ts
infiton: 👍

PR #550, opened 2024-07-22
Added bulk action support to auto table

MillanWangGadget:
- **UPDATE**
  - Enabled record selection functionality on the Polaris table
  - Added system to parse through the actions of the current model manager and allow the execution of bulk delete actions only
    - Currently, only bulk actions that accept only ID lists are compatible
  - Added confirmation modal for bulk actions.


packages/react/src/auto/polaris/PolarisAutoBulkActionModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  return (
+    <>
+      <Modal onClose={() => setShowModal(false)} title={modalTitle} open={showModal}>
+        {
+          isBulkAction ? (
+            <BulkActionModalContent model={model} modelActionDetails={modelActionDetails} ids={ids} close={() => setShowModal(false)} />
+          ) : null
+          // TODO - Decide on support for non-bulk actions
+          // <PolarisAutoForm action={model[actionName]} />
```
alanko0511, line 38: I don't think we should include stuff like this in the codes. We should probably add tickets to them instead.

packages/react/src/auto/polaris/PolarisAutoBulkActionModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const actionResultBanner = useMemo(
+    () => (hasError ? <Banner title={errorMessage} tone="critical" /> : <Banner title={ActionSuccessMessage} tone="success" />),
+    [hasError, errorMessage]
+  );
+
+  const runAction = useCallback(() => {
+    void runBulkAction(ids);
+    setHasRun(true);
+  }, [runBulkAction, setHasRun]);
```
alanko0511, line 77: `[ids, runBulkAction]` instead?

packages/react/src/auto/hooks/useModelActionsForTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+  isBulk: boolean;
+  acceptsModelInput: boolean;
+  variables: any;
+};
+
+/**
+ * Returns a list of ModelActionDetail objects for given model
+ */
+const useModelActionsForTable = (props: { model: any }) => {
+  const modelActions: any[] = [];
```
alanko0511, line 18: Nit:
```ts
const modelActions: ModelActionDetails[] = [];
```

Btw forgot to remove console.log in line 20?

alanko0511 (approved):
🎩 looks good with some comments :shipit: 👌

packages/react/src/auto/polaris/PolarisAutoBulkActionModal.tsx
MillanWangGadget: Removed. This is debatably not needed with the custom cell renderer and the new introduction of the Action button 😎

packages/react/src/auto/polaris/PolarisAutoBulkActionModal.tsx
MillanWangGadget: Good call. Updated

packages/react/src/auto/hooks/useModelActionsForTable.tsx
MillanWangGadget: Updated to be `ModelActionDetails` and removed the leftover log

packages/react/src/auto/hooks/useModelActionsForTable.tsx
```diff
@@ -0,0 +1,67 @@
...
+  isDeleter: boolean;
+  isBulk: boolean;
+  acceptsModelInput: boolean;
+  variables: any;
+};
+
+/**
+ * Returns a list of ModelActionDetail objects for given model
+ */
+const useModelActionsForTable = (props: { model: any }) => {
```
danroberts, line 17: These feel more like helpers than hooks -- I think these could be just utils functions

packages/react/src/auto/hooks/useModelActionsForTable.tsx
```diff
@@ -0,0 +1,67 @@
...
+    isDeleter: func.isDeleter,
+    isBulk: func.isBulk,
+    acceptsModelInput: func.acceptsModelInput,
+    variables: func.variables,
+  });
+
+  return modelActions.map(funcMapper);
+};
+
+export const useTableBulkActions = (props: { model: any }) => {
```
danroberts, line 40: I'm not sure I fully understand the desired usage here -- Why does this keep the selected model action details in state?

MillanWangGadget: This hook was made to provide bulk action options and track which one has been selected so that the modal can be opened with the corresponding action info.

packages/react/src/auto/hooks/useModelActionsForTable.tsx
MillanWangGadget: Updated this helper such that it no longer has a hook-like name

danroberts (approved):

PR #554, opened 2024-07-22
Make skeleton table header wider in AutoTable

alanko0511:
We already have a skeleton table, but because the header is too small, it doesn't show the skeleton properly.

This PR fixes that by setting a width to the headers.

## PR Checklist


MillanWangGadget (approved):
Looks reasonable 👍 

CleanShot 2024-07-22 at 13 39 58@2x

PR #555, opened 2024-07-22
Added error message for passing bulk actions into the autoforms

MillanWangGadget:
- **UPDATE**
  - Previously, passing in a bulk action to the Autoform would result in a `can't find metadata error`
  - Now, we get `bulk actions are not supported` as an error message so the user can more easily identify where to fix the problem


alanko0511 (approved):
:shipit: 👌

PR #556, opened 2024-07-22
Added humanized action names to forms

MillanWangGadget:
- **UPDATE**
  - Added humanized action names to forms 
    - They are based on the GQL operation name, which is based on the API ID and the model
    - Global actions get titles as well
    - Action operation names containing numbers will keep the numbers beside each other 
      - Ex: `Do99Things` -> `Do 99 Things`
    - This only appears in one-liner AutoForms

### DIFF
--- a/packages/react/src/auto/polaris/PolarisAutoForm.tsx
+++ b/packages/react/src/auto/polaris/PolarisAutoForm.tsx
@@ -1,9 +1,9 @@
 import type { ActionFunction } from "@gadgetinc/api-client-core";
 import type { FormProps } from "@shopify/polaris";
-import { BlockStack, Form, FormLayout, SkeletonBodyText, SkeletonDisplayText } from "@shopify/polaris";
-import React from "react";
+import { BlockStack, Form, FormLayout, SkeletonBodyText, SkeletonDisplayText, Text } from "@shopify/polaris";
+import React, { useMemo } from "react";
 import { FormProvider } from "react-hook-form";
-import type { OptionsType } from "../../utils.js";
+import { humanizeCamelCase, type OptionsType } from "../../utils.js";
 import type { AutoFormProps } from "../AutoForm.js";
 import { useAutoForm } from "../AutoForm.js";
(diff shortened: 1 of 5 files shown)


packages/react/src/utils.ts
```diff
@@ -539,3 +539,24 @@ export type RelatedFieldColumn = {
...
   return typeof value === "object" && value !== null && "field" in value && "relatedField" in value;
 };
+
+/**
+ * Humanizes a camelCase string by adding spaces between words and capitalizing the first letter
+ * Examples
+ *    humanizeCamelCase("createNewWidget") => "Create New Widget"
+ *    humanizeCamelCase("do99Things") => "Do 99 Things"
+ */
+export const humanizeCamelCase = (camelCaseString: string): string => {
```
alanko0511, line 549: Given the number of regex operations in this function, it's quite difficult to understand the expected results.

Maybe add a test to prove that the results are expected?

packages/react/src/auto/polaris/PolarisAutoForm.tsx
```diff
@@ -90,6 +90,9 @@ const PolarisAutoFormComponent = <
 
   const formContent = props.children ?? (
     <>
+      <Text variant="headingLg" as="h5">
+        {humanizeCamelCase(action.operationName)}
```
alanko0511, line 94: Do we need to memoise the function result with `useMemo`?

alanko0511 (approved):
🎩 with some small comments :shipit: 👌

packages/react/src/utils.ts
MillanWangGadget: Added a jest test to prove these examples 👍

packages/react/src/auto/polaris/PolarisAutoForm.tsx
MillanWangGadget: Good call. Updated to memoize these humanized names

danroberts (approved):

PR #557, opened 2024-07-22
Added system to clear form fields upon successful action submission for create and update actions

MillanWangGadget:
- **UPDATE**
  - Added a field value resetting system that runs upon successful completion of create and global actions
    - Some state based inputs had to get updates such that their input values would stay in sync with the overall hook context
  - Inputs shall be untouched on errors
  - Passing in an `onSuccess` property will avoid this behaviour

### DIFF
--- a/packages/react/src/auto/AutoForm.ts
+++ b/packages/react/src/auto/AutoForm.ts
@@ -122,6 +122,8 @@ export const useAutoForm = <
   // filter down the fields to render only what we want to render for this form
   const fields = useFormFields(metadata, props);
   const operatesWithRecordId = !!(metadata && isActionMetadata(metadata) && metadata.action.operatesWithRecordIdentity);
+  const isDeleteAction = metadata && isActionMetadata(metadata) && metadata.action.isDeleteAction;
+  const isGlobalAction = action.type === "globalAction";
   const modelApiIdentifier = action.type == "action" ? action.modelApiIdentifier : undefined;
   const defaultValues: Record<string, unknown> = useMemo(
     () =>
@@ -139,6 +141,7 @@ export const useAutoForm = <
   const {
     submit,
     error: formError,
+    reset,
     formState: { isSubmitSuccessful, isLoading, isReady, isSubmitting, touchedFields },
     originalFormMethods,
   } = useActionForm(action, {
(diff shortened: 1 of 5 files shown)


packages/react/src/auto/hooks/useHasManyInputController.tsx
```diff
@@ -24,6 +26,13 @@ export const useHasManyInputController = (props: AutoRelationshipInputProps) =>
 
   const retrievedSelectedRecordIds = selected.records?.map((record: { id: string }) => record.id) ?? [];
   const unlinkedRecordIds = fields.filter((field: any) => field.__unlinkedInverseField).map((field: any) => field.__id);
+  const formContextValue = getValues(path);
+
+  useEffect(() => {
+    if (!formContextValue) {
+      void clearAllFields(); // This is called asynchronously to avoid an infinite loop
+    }
+  }, [!formContextValue || formContextValue.length === 0]);
```
alanko0511, line 35: Why do we need to use `getValues` to see if it has value instead of the `fields` property from the `useFieldArray` hook?

Also I think the dependency array can be like this?
```
[clearAllFields, formContextValue]
```

packages/react/src/auto/AutoForm.ts
```diff
@@ -120,6 +120,8 @@ export const useAutoForm = <
   // filter down the fields to render only what we want to render for this form
   const fields = useFormFields(metadata, props);
   const operatesWithRecordId = !!(metadata && isActionMetadata(metadata) && metadata.action.operatesWithRecordIdentity);
+  const isDeleteAction = metadata && isActionMetadata(metadata) && metadata.action.isDeleteAction;
+  const isGlobalAction = metadata && !isActionMetadata(metadata);
```
alanko0511, line 124: How about `action.type === "globalAction"` like on line 129?

packages/react/src/auto/hooks/useHasManyInputController.tsx
MillanWangGadget: I use `getValues()` here because calling `useActionForm(...) .reset()` does not reset the values used in useFieldArray. They are only reset in `getValues`, thus leading to this synchronization process

And I made the dependency array like that so the useEffect would only run when emptied and unemptied. Having the whole `formContextValue` in there would cause it to run upon every change in the selection, even if the change would not be relevant to that process

packages/react/src/auto/AutoForm.ts
MillanWangGadget: Updated such that the check is `action.type === "globalAction"`

alanko0511 (approved):
🎩 
:shipit: 👌

danroberts (approved):

PR #559, opened 2024-07-23
Introduce custom cell renderer in AutoTable

alanko0511:
This PR adds support for having custom cell renderers in the `columns` property. By passing in an object like this:
```jsx
{
  name: "some custom stuff",
  render: (record) => <div>{record.name}</div>
}
```
it will render the column using the `render` function included in the object instead of using the default cell renderers.

## PR Checklist

### DIFF
--- a/packages/react/src/useTable.tsx
+++ b/packages/react/src/useTable.tsx
@@ -6,78 +6,12 @@ import {
   type LimitToKnownKeys,
   type Select,
 } from "@gadgetinc/api-client-core";
-import type { OperationContext } from "@urql/core";
-import type { ReactNode } from "react";
 import { useMemo } from "react";
-import { GadgetFieldType } from "./internal/gql/graphql.js";
-import type { ModelMetadata } from "./metadata.js";
 import { filterAutoTableFieldList, useModelMetadata } from "./metadata.js";
-import type { SearchResult } from "./useDebouncedSearch.js";
-import type { PaginationResult } from "./useList.js";
 import { useList } from "./useList.js";
-import type { CustomCellColumn, RelatedFieldColumn } from "./utils.js";
-import {
-  isCustomCellColumn,
-  isRelatedFieldColumn,
-  type ColumnValueType,
-  type ErrorWrapper,
-  type OptionsType,
-  type ReadOperationOptions,
-} from "./utils.js";
-
-type BaseTableColumn = {
-  name: string;
-  apiIdentifier: string;
-};
-
-type RecordTableColumnValue = BaseTableColumn & {
-  fieldType: GadgetFieldType;
-  sortable: boolean;
-  relatedField?: RecordTableColumnValue;
-  getValue: (record: GadgetRecord<any>) => ColumnValueType;
-  isCustomCell: false;
-};
-
-type CustomTableColumnValue = BaseTableColumn & {
-  getValue: (record: GadgetRecord<any>) => ReactNode;
-  isCustomCell: true;
-  sortable: false;
-};
-
-export type TableColumn = RecordTableColumnValue | CustomTableColumnValue;
-
-export type TableRow = Record<string, ColumnValueType | ReactNode>;
-
-export interface TableOptions {
-  pageSize?: number;
-  initialCursor?: string;
-  initialDirection?: "forward" | "backward";
-  columns?: (string | RelatedFieldColumn | CustomCellColumn)[];
-}
-
-export type TableData<Data> =
-  | {
-      columns: TableColumn[];
-      rows: TableRow[];
-      data: Data;
-      metadata: ModelMetadata;
-    }
-  | {
-      columns: null;
-      rows: null;
-      data: null;
-      metadata: null;
-    };
-
-export type TableResult<Data> = [
-  TableData<Data> & {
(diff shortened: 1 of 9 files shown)


packages/react/src/useTable.tsx
```diff
@@ -205,8 +228,25 @@ export const useTable = <
...
         sortable: "sortable" in field && field.sortable,
         relatedField,
-      } as TableColumn;
+        isCustomCell: false,
+      } satisfies RecordTableColumnValue;
     });
+
+    if (options?.columns) {
+      // Add custom cell columns
+      for (const custom of options.columns) {
```
danroberts, line 237: I think this should be for (const column of options.columns) for more accurate naming

packages/react/src/useTable.tsx
```diff
@@ -205,8 +228,25 @@ export const useTable = <
         },
         sortable: "sortable" in field && field.sortable,
         relatedField,
-      } as TableColumn;
+        isCustomCell: false,
+      } satisfies RecordTableColumnValue;
     });
+
+    if (options?.columns) {
```
danroberts, line 235: Isn't this always going to add the custom columns at the end of the column list? I think they should be in the order they're passed.

alanko0511: Updated to sort the column result based on the columns option

packages/react/src/useTable.tsx
```diff
@@ -205,8 +228,43 @@ export const useTable = <
...
```
MillanWangGadget, line 235: The types in here are pretty wild, but do you think it would be feasilble to move this section to a helper function? This hook is starting to get quite big

MillanWangGadget (approved):
MANUAL TESTS
- 🍏 custom cell renderers appear in the same order that they're defined in the `columns` prop, respecting string-only columns
- 🍏 Having an onClick for the row as well as a onClick in the custom cell renderer works
  - Cell renderer onClick runs first before row onClick
  - Can prevent the row onClick propagation with `e.preventDefault(); e.stopPropagation();`
- 🍏 Multiple custom renderers with the same name can appear at the same time
- 🍏 You can put a table within a table 
  - CleanShot 2024-07-23 at 15 42 41@2x


One optional nit, but good to go otherwise 👍

packages/react/src/useTable.tsx
alanko0511: Moved the types into a separate file.

packages/react/src/useTable.tsx
```diff
@@ -248,10 +149,105 @@ const getRows = (props: { data: GadgetRecord<any>[]; columns: TableColumn[] }) =
...
-    const row: Record<string, ColumnValueType> = { id: record.id };
+    const row: TableRow = { id: record.id };
     for (const { apiIdentifier, getValue } of columns) {
       row[apiIdentifier] = getValue(record);
     }
     return row;
   });
 };
+
+const getTableColumnArray = (fields: FieldMetadataFragment[], columnsFromOptions: TableOptions["columns"]) => {
```
danroberts, line 160: this should be in a useMemo or useCallback or it will be regenerated on every re-render

danroberts: I also notice this is true of `getTableData` and `getRows`.

packages/react/src/useTable.tsx
```diff
@@ -248,10 +149,105 @@ const getRows = (props: { data: GadgetRecord<any>[]; columns: TableColumn[] }) =
...
+  const columnsMap =
```
danroberts, line 161: I know this was pre-existing, but I think this name is a bit unclear, maybe it could be 'relatedColumnFieldMap or something like that?

packages/react/src/useTable.tsx
```diff
@@ -248,10 +149,105 @@ const getRows = (props: { data: GadgetRecord<any>[]; columns: TableColumn[] }) =
...
+          return [column, undefined];
+        }
+      })
+    );
+
+  const result: TableColumn[] = fields.map((field) => {
+    let relatedField: RecordTableColumnValue | undefined;
+    const relatedFieldColumn = columnsMap?.get(field.apiIdentifier);
+
+    if (relatedFieldColumn) {
```
danroberts, line 177: This entire piece of logic for getting the related field feels like it could be pulled into a helper function

packages/react/src/useTable.tsx
danroberts, line 160: I'm not sure how clear `columnsFromOptions` is as a variable name -- does it matter why they come from the options? Couldn't this just be 'columns' or something similar?

alanko0511: `getTableColumnArray` is memoised with `useMemo` when calling the function.

danroberts: That doesn't really matter, it's the definition of the function that needs to be memoized or it will be re-allocated every render cycle.

danroberts: or the function can be moved outside of the hook scope if it doesn't use any of the state other than it's params, which I would say it probably shouldn't

danroberts: disregard this, misread the closure scope, it's already outside the hook.

PR #561, opened 2024-07-23
Removed Lodash from react package

MillanWangGadget:
- **UPDATE**
  - This dependency disrupts the importation of the react package. Replaced the package import with locally defined functions


alanko0511 (approved):
:shipit: 👌

PR #562, opened 2024-07-23
Introduce `onFailure` callback in AutoForm

alanko0511:
This PR introduces an `onFailure` callback in the `<AutoForm />` component. We already have a similar callback called `onError` in our `useActionForm` so we only need to expose it as part of the available property in the React component.

The default banner will not be shown when the `onFailure` callback exists, so people can override the behaviour.

## PR Checklist

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoForm.cy.tsx
@@ -64,6 +64,26 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
       });
   });
 
+  it("onFailure callback should return an error if the form submission fails", () => {
+    const onFailureSpy = cy.spy().as("onFailureSpy");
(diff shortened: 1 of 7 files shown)


packages/react/cypress/component/auto/form/AutoForm.cy.tsx
```diff
@@ -64,6 +64,26 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
       });
   });
 
+  it("onFailure callback should return an error if the form submission fails", () => {
+    const onFailureSpy = cy.spy().as("onFailureSpy");
+    cy.mountWithWrapper(<AutoForm action={api.widget.alwaysThrowError} exclude={["gizmos"]} onFailure={onFailureSpy} />, wrapper);
```
alanko0511, line 69: `alwaysThrowError` is a special model action that throws an error no matter what.

MillanWangGadget (approved):
MANUAL TESTS
- 🍏 Confirmed that the onFailure runs on serverside errors
- 🍏 Confirmed that the onFailure runs on client side errors (ex: Missing required param)
- 🍏 Error banner only appears when there is no `onFailure` passed into the form 

Code changes look reasonable

danroberts (approved):

PR #563, opened 2024-07-24
Added error for custom params colliding with model fields

MillanWangGadget:
- **UPDATE**
  - Added an error for custom params with api IDs that collide with model fields
  - Why make an error? 
    - `<AutoInput field={"fieldApiId} />` does a contextual api Id based lookup to through the action metadata to find out which one it should correspond to. 
      - To discriminate between a model field and an action custom param, we would need...
        - A: Another component property to ALL auto inputs that can collide with modelFIelds
        - B: Force the user to use path like `modelA.field`. 
      - More complexity for EVERYONE when this is a really rare use case
    - The automatic labeling is based on API id and would result in duplicately named inputs. Confusing
    - `include`/`exclude` would need some changes in order to discriminate between modelFields vs customParams
    - The way that useActionForm combines the existing record data with the input values to make a GQL request will need to be updated since pathing the custom param to the root and the modelField to `modelApiId.fieldApiId` conflicts with this system
  - Overall, having those collisions will be very rare, and preventing feels more correct than all of the required technical complexity to support it

### DIFF
--- a/packages/react/src/auto/AutoForm.ts
+++ b/packages/react/src/auto/AutoForm.ts
@@ -69,6 +69,7 @@ export const useFormFields = (
   return useMemo(() => {
     if (!metadata) return [];
     const action = isActionMetadata(metadata) ? metadata.action : metadata;
+
     const isModelMetadata = metadata.__typename === "GadgetModel";
 
     const objectFields = isModelMetadata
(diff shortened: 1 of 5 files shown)


packages/react/src/auto/AutoForm.ts
```diff
@@ -96,10 +97,24 @@ export const useFormFields = (
...
+    return allFormFields;
   }, [metadata, options]);
 };
 
+const validateFormFieldApiIdentifierUniqueness = (apiIdentifiers: string[]) => {
+  const seen = new Set<string>();
+
+  for (const apiId of apiIdentifiers) {
+    if (seen.has(apiId)) {
+      throw new Error(`Input "${apiId}" appears multiple times in the action metadata. Input names must be unique.`);
```
alanko0511, line 112: "action metadata" is more like internal stuff, and this error can be thrown when the user sets up their model action with duplicated input parameters, not when our side has some unexpected behaviours, so I think this error message is quite confusing for users.

How about a message like "Action iput parameter ${apiId} must be unique." or something similar?

packages/react/src/auto/AutoForm.ts
```diff
@@ -66,6 +66,7 @@ export const useFormFields = (
   metadata: ActionMetadata | GlobalActionMetadata | undefined | null,
   options: { include?: string[]; exclude?: string[] }
 ): readonly { path: string; metadata: FieldMetadata }[] => {
+  console.log("metadata :", metadata);
```
alanko0511, line 69: Forgot to remove the console.log?

alanko0511 (approved):
🎩 
:shipit: 👌

packages/react/src/auto/AutoForm.ts
MillanWangGadget: Good catch. Removed

packages/react/src/auto/AutoForm.ts
MillanWangGadget: Updated to be 
```
`Input "${apiId}" is not unique for action "${actionApiIdentifier}"`
```

PR #564, opened 2024-07-24
fix(autocomponents): fix up package dependencies + add vite int. test

danroberts:
We allowed a few packaging issues to slip through, which caused errors when importing from @gadgetinc/react/auto.

This fixes a few of those depednency issues and introduces an integration test which pacakges and installs @gadgetinc/react in a new vite project and imports AutoForm.

## PR Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -33,7 +33,10 @@
   "sideEffects": false,
   "scripts": {
     "typecheck": "tsc --noEmit",
-    "build": "rm -rf dist && mkdir -p dist/cjs dist/esm auto/polaris auto/mui && echo '{\"type\": \"commonjs\"}' > dist/cjs/package.json && echo '{\"main\": \"../../dist/cjs/auto/polaris/index.js\"}' > auto/polaris/package.json && echo '{\"main\": \"../../dist/cjs/auto/mui/index.js\"}' > auto/mui/package.json && echo '{\"type\": \"module\"}' > dist/esm/package.json && tsc -b tsconfig.cjs.json tsconfig.esm.json ",
+    "clean": "rimraf dist/ auto/",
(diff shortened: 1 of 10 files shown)


danroberts:
Superseded by https://github.com/gadget-inc/js-clients/pull/564

packages/react/package.json
```diff
@@ -48,6 +51,8 @@
     "@gadgetinc/api-client-core": "^0.15.24",
     "@graphql-typed-document-node/core": "^3.2.0",
     "@hookform/resolvers": "^3.3.1",
+    "@types/tmp": "^0.2.6",
+    "copyfiles": "^2.4.1",
```
MillanWangGadget, line 55: Out of curiousity, what goes into the decision of making these `dependencies` vs `devDependencies`. It that they are not referenced by prodution facing files

MillanWangGadget (approved):
- 🍏 No more shopify app bridge in PolarisAutoButton
- 🍏 The errors for missing CSS files from within this repo are gone upon gitpkg installation 
- 🍏 New vite tests look reasonable

MillanWangGadget:
For the mdxEditor css file, would it be reasonable to make a copy in the repo and import it in the same way that these css flies are included? That way, the richText can truly become an optional dep

packages/react/package.json
danroberts: They should be devDependencies, good catch.

danroberts:
> For the mdxEditor css file, would it be reasonable to make a copy in the repo and import it in the same way that these css flies are included? That way, the richText can truly become an optional dep

I'm not sure how this would work, but I think we should do a followup here to make the mdx editor more of a standalone optional dep -- right now, installing the @mdxeditor package as a peer dependency is still required.

airhorns (approved):
LGTM

alanko0511 (approved):

PR #566, opened 2024-07-24
Revert "Added error for custom params colliding with model fields"

MillanWangGadget:
Reverts gadget-inc/js-clients#563


MillanWangGadget:
No longer needed

PR #567, opened 2024-07-25
Cypress test to ensure explicit sort is handled correctly

gabeb03:
A test is missing from #534. It is added in this PR.

It ensures that when a sort prop is included in AutoTable, the correct column is sorted. 

## PR Checklist


MillanWangGadget (commented):
Nice test 👍 

Passing in a `sort={{ someFIeld: "Ascending" }}` to an autotable makes it such that the table will be initialized with that sort, but it will still be possible to manually sort afterwards. I wonder if a prop name like `defaultSort` or `initialSort` would be more intuitive since the user can adjust the sort afterwards, and dynamically changing the input sort prop would not change anything.

MillanWangGadget (approved):

danroberts:
I'd have a slight preference for `initialSort` I think it's a tiny bit more accurate

PR #572, opened 2024-07-25
fix(autoform): dynamically import mdx module to prevent vite bundling…

danroberts:
In order to keep @mdxeditor/editor a true optional dependency, we need to dynamically import the module inside the react component. 

I don't think Vite was seeing the React.lazy load as a dynamic import, and was trying to bundle the dependency at build time. 

I also had to vendor the style.css from @mdxeditor/editor as I couldn't figure out a way to dynamically import the .css file

## PR Checklist

### DIFF
--- a/packages/react/src/auto/shared/AutoRichTextInput.tsx
+++ b/packages/react/src/auto/shared/AutoRichTextInput.tsx
@@ -1,49 +1,73 @@
-import {
-  BlockTypeSelect,
-  BoldItalicUnderlineToggles,
-  CodeToggle,
-  CreateLink,
-  DiffSourceToggleWrapper,
-  ListsToggle,
-  MDXEditor,
-  Separator,
-  UndoRedo,
-  diffSourcePlugin,
-  headingsPlugin,
-  linkDialogPlugin,
-  listsPlugin,
-  markdownShortcutPlugin,
-  quotePlugin,
-  thematicBreakPlugin,
-  toolbarPlugin,
-  type MDXEditorMethods,
-  type MDXEditorProps,
-} from "@mdxeditor/editor";
-import "@mdxeditor/editor/style.css";
 import type { ForwardedRef } from "react";
-import React, { useEffect, useRef } from "react";
-import { useFormContext, type Control } from "react-hook-form";
+import React, { useEffect, useRef, useState } from "react";
+import type { Control } from "react-hook-form";
+import { useFormContext } from "react-hook-form";
 import { get } from "../../utils.js";
 import { useStringInputController } from "../hooks/useStringInputController.js";
 import { multiref } from "../hooks/utils.js";
 
-export default function AutoRichTextInput(
-  props: {
-    field: string;
-    control?: Control<any>;
-    editorRef?: ForwardedRef<MDXEditorMethods> | null;
-  } & Omit<MDXEditorProps, "markdown" | "onChange">
-) {
+interface MDXEditorMethods {
+  setMarkdown: (markdown: string) => void;
+}
+
+interface AutoRichTextInputProps {
+  field: string;
+  control?: Control<any>;
+  editorRef?: ForwardedRef<MDXEditorMethods> | null;
+}
+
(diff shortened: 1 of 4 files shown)


airhorns (commented):
This approach makes sense to me, but I think its worth testing out how things will work with MUI exports in the package when you dont have MUI installed, because if we can't make that work, then I think we need to switch all of this approach around and it will change this bit.

MillanWangGadget (commented):
MANUAL TESTING
- 🍏 We can load autoforms without the mdx peer dependency installed, and we have a loader for where the richText should be used 
  -  CleanShot 2024-07-26 at 11 18 25@2x
  - Temporarily brought back the MUI index file to test this 
- 🟡 Both MUI and polaris don't look right when the mdxEditor is in the package json
  - CleanShot 2024-07-26 at 11 25 17@2x
  - The css file is in the sandbox at `node_modules/@gadgetinc/react/dist/esm/auto/shared/styles/mdxeditor`
  - Even after adding the peer that appear as warnings in after running `yarn install` in the editor
```
"@codemirror/language": "^6.0.0",
"@lezer/highlight": "^1.0.0",
"yjs": "13.5.22"
```

danroberts:
@MillanWangGadget good catch. The way we're importing the css isn't going to work, and I don't see any way to do it dynamically. We will have to require the user to import the css in their app with instructions in the documentation the same way we do for polaris. 

You will have to add 

import '@mdxeditor/editor/style.css'

Somewhere in your project, and then it should work.

packages/react/src/auto/shared/AutoRichTextInput.tsx
```diff
@@ -1,49 +1,72 @@
...
+
+  useEffect(() => {
+    if (innerRef.current && isEditorLoaded) {
       innerRef.current.setMarkdown(controller.value?.markdown ?? "");
     }
-  }, [controller.value]);
+  }, [controller.value, isEditorLoaded]);
+
+  if (!isEditorLoaded || !mdxModule) {
+    return <div>Loading editor...</div>;
```
airhorns, line 48: Did you want to try to use suspense to allow users to customize this loading state?

airhorns:
> We will have to require the user to import the css in their app with instructions in the documentation the same way we do for polaris.

I think if you can dynamically import the module you can also dynamically import the CSS, why not?

danroberts:
> I think if you can dynamically import the module you can also dynamically import the CSS, why not? 

I think the reason is that the css import is a vite plugin that transforms the css import into a compiled stylesheet, and I don't see any way to to teach the plugin about dynamic imports.

> I think its super undesirable to have the base state be unstyled until you go find this tiny callout in the docs that says you have to add an import

I agree, but that's already how it works for all of the polaris components.

danroberts:
> > I think its super undesirable to have the base state be unstyled until you go find this tiny callout in the docs that says you have to add an import
> 
> I agree, but that's already how it works for all of the polaris components.

Also how it would work if you were consuming the base mdxeditor component.

airhorns (approved):

MillanWangGadget:
This works when the MDX dependency is installed, but when it is not installed, importing AutoForm fails, even as an unused import

CleanShot 2024-07-29 at 18 58 45@2x

airhorns:
Hm, I think that means the tests are not testing what we think. Do we need to also do a cypress / playwright test against the built vite and test that it works maybe?

danroberts:
> Hm, I think that means the tests are not testing what we think. Do we need to also do a cypress / playwright test against the built vite and test that it works maybe?

The test ended up using vite 5, which seems to be able to build this with the missing css import, whereas we're using vite 4, which can't.

danroberts:
@MillanWangGadget I think this should work now.

MillanWangGadget (approved):
MANUAL TESTING
- 🍏 Same behaviour in Web template and Shopify template
- 🍏 Without MDX import, using AutoForm works without richtext fields
- 🍏 Without MDX import, richtext fields cause missing dependency errors
- 🍏 With MDX import, richtext fields render properly

Good to go 👍

PR #574, opened 2024-07-26
fix(autotable): sorting after using pagination results in error due t…

danroberts:
If you paginate to a new page and then sort the column by a header, the query would throw an error because the cursor location wasn't necessarily valid anymore. 

I decided to fix this in useList, to make the logic: "if the sort option changes, reset the cursor tracking" so that we don't have to fix this bug again if we leverage useList in other places. 

## PR Checklist

### DIFF
--- a/packages/react/src/useList.ts
+++ b/packages/react/src/useList.ts
@@ -1,10 +1,11 @@
 import type { DefaultSelection, FindManyFunction, GadgetRecord, LimitToKnownKeys, Select } from "@gadgetinc/api-client-core";
 import type { OperationContext } from "@urql/core";
-import { useCallback, useState } from "react";
+import { useCallback, useMemo, useState } from "react";
 import type { SearchResult } from "./useDebouncedSearch.js";
 import { useDebouncedSearch } from "./useDebouncedSearch.js";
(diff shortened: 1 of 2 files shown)


packages/react/src/useList.ts
```diff
@@ -56,6 +57,12 @@ export const useList = <
   const [direction, setDirection] = useState<"forward" | "backward">("forward");
   const clearCursor = useCallback(() => setCursor(undefined), []);
 
+  useMemo(() => {
```
MillanWangGadget, line 60: My first instinct to do something like this would be to use a `useEffect`. What goes into the decision of `useEffect` vs `useMemo`

MillanWangGadget (approved):
MANUAL TESTING
- 🍏 Everytime the column sorting is changed, the cursor is reset and we go back to page 1

packages/react/src/useList.ts
danroberts: just a 🧠 fart -- thought i was typing useEffect

gabeb03 (approved):
MANUAL TESTING

🍏 Sorted column on page 2; pagination is reset & sort works
🍏 Sorting persists when you load the next page

PR #575, opened 2024-07-26
Refactor useTable row & columns generation to support relationship

alanko0511:
This PR refactors how we generate a list of rows and columns to support relationship fields better, fixes bugs when selecting file/rich text field, and introduces a new `excludeColumns` property.

Related tickets:
GGT-6844, GGT-6863, GGT-6847, GGT-6846

## PR Checklist


packages/react/src/useTableUtils/types.ts
```diff
@@ -36,6 +36,7 @@ export interface TableOptions {
   initialCursor?: string;
   initialDirection?: "forward" | "backward";
   columns?: (string | RelatedFieldColumn | CustomCellColumn)[];
+  exclude?: string[];
```
MillanWangGadget, line 39: Could you rename this to `excludeColumns`

I'm adding a system that does a similar `include` thing with actions, and I think that it would be nice to have `actions` & `excludeActions` with names that align with the columns

packages/react/src/useTableUtils/UseTableDataGenerator.ts
MillanWangGadget, line 1: Having this file start with `use` can kinda give off the impression that this is a react hook when it is actually a class. This is used in quite a hook-like way though so what do you think about converting this into a hook or renaming it to avoid starting with `use`

packages/react/spec/auto/table/PolarisAutoTable.spec.tsx
```diff
@@ -331,6 +332,7 @@ describe("PolarisAutoTable", () => {
 
     const firstRow = rows[0];
     const firstRowCells = firstRow.getElementsByClassName(POLARIS_TABLE_CLASSES.CELL);
+    console.log(firstRowCells[2].innerHTML);
```
MillanWangGadget, line 335: Leftover log

packages/react/spec/auto/hooks/useTable.spec.tsx
```diff
@@ -151,7 +151,7 @@ describe("useTable hook", () => {
         isCustomCell: false,
         name: "Secret key",
         relatedField: undefined,
-        sortable: false,
+        sortable: true,
```
MillanWangGadget, line 154: How come EncryptedString and Role fields became sortable? They shouldn't be

packages/react/src/useTableUtils/UseTableDataGenerator.ts
```diff
@@ -0,0 +1,205 @@
...
+      const excludeSet = new Set(excludeColumns);
+      this.targetColumns = filterAutoTableFieldList(fieldMetadataArray)
+        .filter((field) => !excludeSet.has(field.apiIdentifier))
+        .map((field) => field.apiIdentifier);
+    } else {
+      this.targetColumns = filterAutoTableFieldList(fieldMetadataArray).map((field) => field.apiIdentifier);
+    }
+  }
+
+  getTableSelectionMap() {
```
MillanWangGadget, line 81: In here, there should probably be some error for passing `columns={["passwordField"]}` since this will result in this GQL error, which does not obviously indicate that we can't select password fields as columns

CleanShot 2024-07-26 at 17 45 50@2x

MillanWangGadget (commented):
MANUAL TESTING
- Including field api ids in the `columns` prop
  - 🍏 Rich text fields work
  - 🍏 File fields work
  - 🍏 HasMany fields work
  - 🍏 HasOne fields work
  - 🍏 BelongsTo fields work
- Custom cell renderers
  - 🟡 Computed fields only appear in the record if they're also included by apiId only
    - CleanShot 2024-07-26 at 16 37 49@2x
  - 🟡 Relationship fields are not available in the custom cell renderer
    - Debatable if this is desired. Perhaps trying to do this should carry an error
- 🍏 Cannot use `columns` and `excludeColumns` at the same time
- 🟡 You can exclude columns that don't exist. 
  - Maybe this should be an error so that renaming a gadget field won't automatically expose the previously hidden column


Some code comments but functionally sound 👍

packages/react/src/useTableUtils/UseTableDataGenerator.ts
MillanWangGadget: Same thing with hasManyThrough fields. THose are in the metadata but currently can't be queried and don't have a cell renderer

danroberts:
I think the custom renderer should get access to whatever is in the default selection. That doesn't include computed fields or relationship fields, and if you want access to those, you need to pass a custom selection and manage it yourself.

packages/react/src/useTable.tsx
```diff
@@ -53,78 +53,42 @@ export const useTable = <
...
-          selectionMap[column.field] = {
-            edges: {
-              node: {
-                [column.relatedField]: true,
-              },
-            },
-          };
-        }
-      }
+  const useTableDataGenerator = useMemo(() => {
```
danroberts, line 56: We might have to dig into the details here, but I'm not sure that a class is the best way to extract these helpers -- any reason not to extract the individual helpers as functions? 

The reason I think a class might be a challenge down the road is that it introduces additional state management that might be a bit tricky to get right in terms of memoization, etc.

alanko0511: I used a class instead of individual helper functions because I personally find it annoying to pass in the same parameters, such as `metadata` and `columns,` to all these functions all over the place.

MillanWangGadget (approved):
- 🍏 Functionification of helpers makes sense
- 🍏 Illegal field types are properly excluded
- 🍏 OK to exclude certain field types from being directly accessible in the cell renderer record value if they are computationally expensive

danroberts (approved):

PR #576, opened 2024-07-26
fix(autotable): enable tooltip on hover for string cells with truncat…

danroberts:
Due to a few unfortunate limitations in polaris, it's quite hard to figure out if a <Text> element with `truncate` is displaying with an ellipsis or not. 

This method works but requires a ref on the element to detect if it has overflowed, so I essentially vendored the polaris classes into the text cell markup. 

I think the only other solution would be to take over truncation in js land, which raises some performance concerns for me. 

## PR Checklist

### DIFF
--- a/packages/react/src/auto/polaris/tableCells/PolarisAutoTableTextCell.tsx
+++ b/packages/react/src/auto/polaris/tableCells/PolarisAutoTableTextCell.tsx
@@ -1,13 +1,34 @@
-import { Text } from "@shopify/polaris";
-import React from "react";
+import { Tooltip } from "@shopify/polaris";
+import React, { useEffect, useRef, useState } from "react";
 
 export const PolarisAutoTableTextCell = (props: { value: any }) => {
   const { value } = props;
   const stringifiedValue = typeof value === "object" ? ("markdown" in value ? value.markdown : JSON.stringify(value)) : value;
 
-  return (
-    <Text as="span" truncate>
+  const [isOverflowed, setIsOverflow] = useState(false);
+  const textElementRef = useRef<HTMLSpanElement>(null);
+
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


airhorns (approved):
Works for me!

packages/react/src/auto/polaris/tableCells/PolarisAutoTableTextCell.tsx
```diff
@@ -1,13 +1,30 @@
...
-    </Text>
+    </span>
+  );
+
+  if (!isOverflowed) {
+    return polarisTextWithRef;
+  }
+
+  return (
+    <Tooltip content={stringifiedValue} hoverDelay={500}>
```
MillanWangGadget, line 26: CleanShot 2024-07-26 at 15 45 07@2x

If the length gets really extreme, the tooltip can spill out of the window. This `content` prop can take in react components though which might make it possible to put in a scrollable. 

Although, it might be easier to do that with a popover since it seems that Tooltips seem to flash in/out of existence upon mouse scrolling. Using `active={true}` in the tooltip wont force it to stay open either

airhorns: I spoke with Briggs about this -- I think the tooltip should be allowed to be wider than the column, and have its own vertical stroll within it

danroberts: @MillanWangGadget updated to widen the tooltip and scroll within the tooltip container

MillanWangGadget:
Nice and scrollable 👍 
CleanShot 2024-07-29 at 16 11 32

MillanWangGadget (approved):

PR #577, opened 2024-07-27
Added custom bulk action callback and render system

MillanWangGadget:
- **UPDATE**
  - The bulk action system has been updated to allow for custom callbacks and renderers
    - Custom renderers
      - Similar to the cell renderer, the user can now pass in a React component returning function that takes in record IDs and the full records.
      - POTENTIAL USE CASES
        - Using IDs to power a useBulkAction or other Gadget hooks 
        - Taking the records and doing an operation on their values. Ex: Showing the sum of all records on a particular field
    - Custom callbacks 
      - Similar to the custom renderer, runs the given callback with the selectedIds and selectedRecords and then shows the result of the callback in the modal 
      - This could be useful so that users don't need to make a whole react component, but debatable if this is redundant with the custom renderer. 
  - An `excludeActions` property has been added to the AutoTable
    - Error if excluding an action that does not exist on the model so that renaming a model won't accidentally expose it
  - `bulk` is now removed from the labels for bulk actions since it was contextually obvious

### DIFF
--- a/packages/react/spec/auto/PolarisAutoTable.stories.jsx
+++ b/packages/react/spec/auto/PolarisAutoTable.stories.jsx
@@ -73,8 +73,7 @@ export const onClickCallback = {
   args: {
     model: api.autoTableTest,
     onClick: (row) => {
-      // eslint-disable-next-line no-undef
-      window.alert(`You clicked on row: ${JSON.stringify(row, null, 2)}`);
+      windowAlert(`You clicked on row: ${JSON.stringify(row, null, 2)}`);
     },
   },
 };
@@ -200,8 +199,7 @@ const CustomDeleteButtonCellRenderer = (props) => {
 
   useEffect(() => {
     if (error) {
-      // eslint-disable-next-line no-undef
-      window.alert(`Error deleting record: ${error.message}`);
+      windowAlert(`Error deleting record: ${error.message}`);
     }
   }, [error]);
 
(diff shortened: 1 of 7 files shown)


packages/react/spec/auto/PolarisAutoTable.stories.jsx
```diff
@@ -182,3 +180,64 @@ export const CustomCellWithDeleteButton = {
...
+  args: {
+    model: api.autoTableTest,
+    excludeActions: ["delete"],
+    actions: [
+      "customAction",
+      {
+        label: "Sum ids",
+        callback: (ids) => {
+          const idSum = ids.reduce((i, total) => parseInt(i) + parseInt(total));
+          return idSum;
```
danroberts, line 194: What does it mean functionally for the callback to return a value like this?

MillanWangGadget: A defined callback return value is shown in the confirmation modal after the action is run. if there is no defined renderer

danroberts: I think the behavior for custom action handlers is it should override the modal behavior, it should just be a callback with the ids/records.

danroberts: you would have to implement your own confirmation/success messages as you desired.

MillanWangGadget: Do you think there should be a confirmation modal at all if you have a callback with no renderer? Otherwise it could be hard to tell if you clicked on the action or closed the menu without that feedback

MillanWangGadget: Updated `actions` such that
- Passing in a string will look for a corresponding Gadget bulk action on the model
  - Has a confirmation modal before it is run
- Passing in an object with `callback` will call upon that function with `(ids, selectedRows)` immediately upon selecting the action from the dropdown
- Passing in an object with `render` will load a modal containing the result of the `render` function. 
- `render` and `callback` are mutually exclusive. Having both or none in an object will throw an error

danroberts (approved):

alanko0511 (approved):
🎩 
:shipit: 👌

PR #578, opened 2024-07-29
Exclude relationship fields in AutoForms when related to shopifyShop model

MillanWangGadget:
- **UPDATE**
  - Some model relationships should not be touched by forms
    - Ex: Don't reassign a shopifyOrder to a different shopifyShop
  - Added a system that prevents relationship fields that are related to models with specific keys from being shown in the form
- **STORYBOOK TESTING**
  - Add `"DataModel-T9yqdTpcpBBH"` to `specialModelKeys` to remove the gizmo related field input from widget


packages/react/spec/auto/hooks/useFormFields.spec.tsx
```diff
@@ -63,6 +63,11 @@ describe("useFormFields hook", () => {
     expect(fields.map((field) => field.metadata.apiIdentifier)).toEqual(["stringField1", "params_str", "params_num", "params_bool"]);
   });
 
+  test("Excludes relationship fields that have shopifyShop as the related model even if explicitly included", () => {
```
danroberts, line 66: I don't think it should exclude it if it is explicitly included, just not included by default.

MillanWangGadget: - Updated such that explicitly including a relationship field will dodge the check to exclude relationship fields to special models. 
- This requires that `include` is in the autoForm context though because we previously only used `options.include` to determine what fields to show, and not in the field metadata lookup process in AutoInput 
  - This meant that checking the includes to allow the special model relationships from within AutoInput would always fail because we didn't have access to the includes at that point

danroberts (approved):

alanko0511 (approved):

PR #579, opened 2024-07-29
Added error for missing findBy value for actions that require IDs

MillanWangGadget:
- **UPDATE**
  - It was previously possible to have an update/delete/custom action in an auto form without a findBy value. In this state, the `submit` button would do nothing and there would be no corresponding error
  - Now, similar to the error for having a findBy on a create action, there is now an error for NOT having a findBy value on actions that require an ID input

### DIFF
--- a/packages/react/spec/auto/PolarisAutoForm.spec.tsx
+++ b/packages/react/spec/auto/PolarisAutoForm.spec.tsx
@@ -36,7 +36,22 @@ describe("PolarisAutoForm", () => {
           // @ts-expect-error: mixing a create and a findBy should throw a type error too
           render(<PolarisAutoForm action={api.widget.create} findBy="1234" />, { wrapper: PolarisMockedProviders });
           loadMockWidgetCreateMetadata();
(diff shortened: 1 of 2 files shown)


packages/react/spec/auto/PolarisAutoForm.spec.tsx
```diff
@@ -701,22 +716,24 @@ function loadMockWidgetCreateMetadata() {
...
 
-  mockUrqlClient.executeQuery.pushResponse("widget", {
-    stale: false,
-    hasNext: false,
-    data: getWidgetRecord({
-      name: "Test Widget",
-      inventoryCount: 42,
-    }),
-  });
+  if (mockFindByResult) {
```
danroberts, line 727: I think this should be a separate helper function rather than a boolean flag

MillanWangGadget: Updated such that this is now a separate helper function

danroberts (approved):

alanko0511 (approved):

PR #580, opened 2024-07-29
Add a pre-render with mocked GQL responses to test to try to prevent the test stuck-in-suspense test flake

MillanWangGadget:
- **UPDATE**
  - The test flake in `packages/react/spec/auto/PolarisAutoForm.spec.tsx` was is caused by getting stuck in a suspense's fallback
    - This only happens on the first test in the whole test suite that mocks a GQL response.
    - A simple render test has been added to absorb the the stuck-in-suspense situation, allowing us to proceed normally though all other tests

Otherwise...
CleanShot 2024-07-29 at 17 38 56@2x


alanko0511 (commented):
Have you tried several times with CI to ensure it really fixes the flake? 🤔

PR #581, opened 2024-07-30
GGT-6874 - Added select prop to AutoTable

gabeb03:
Currently, the rows of the AutoTable are always selectable. 

This PR gives AutoTable users access to the Index Table's `selectable` prop, if `selectable={false}`, the rows are no longer selectable: 

image

## PR Checklist


MillanWangGadget (approved):
- 🍏 Selectable defaults to true
- 🍏 Cannot select with selectable=false
- 🍏 onClick callback still works with selectable=false

packages/react/src/auto/AutoTable.tsx
```diff
@@ -24,4 +24,5 @@ export type AutoTableProps<
   filter?: Options["filter"];
   actions?: TableOptions["actions"];
   excludeActions?: TableOptions["excludeActions"];
+  selectable?: boolean;
```
danroberts, line 27: This is making me realize that we're not exposing or passing along the Polaris table props -- this should have already been exposed via that channel I think.

gabeb03: After our online discussion, I will update this PR to pass all the relevant Polaris index table props, namely

```
emptyState?
lastColumnSticky?
selectable?
hasZebraStriping?
resourceName?
{ singular: string; plural: string; }
condensed?
```

danroberts (approved):

PR #582, opened 2024-07-30
Fixed AutoForm number input spinner button usage

MillanWangGadget:
- **Previously**
  - The up/down spinner buttons in number field inputs did not work when number fields did not have a defined `decimals` value
  - This was because we defaulted their step to "any". We did this to avoid the default `step={1}` which would sometimes erroneously cause a `must be an integer` error when the field would accept decimals 
- **Now**
  - The step is set dynamically to match the current value in the input if there is no decimal count from the field configuration 
- **Things to test**
  - UpDown spinner buttons on number inputs work
  - Using decimals and the UpDown spinner buttons will change the highest precision decimal 
  - UpDown arrow keys will be similar to clicking the UpDown spinner buttons, but they will not add trailing zeros
    - This basically incrementally increases the steps as you hold the key, but that feels ok. Basically, an accelerating number changer

### DIFF
--- a/packages/react/spec/auto/inputs/PolarisAutoTextInput.spec.tsx
+++ b/packages/react/spec/auto/inputs/PolarisAutoTextInput.spec.tsx
@@ -189,7 +189,7 @@ describe("PolarisAutoTextInput", () => {
     test("it renders number inputs with any step value when no decimals are set", async () => {
       result = render(<PolarisAutoNumberInput field="inventoryCount" />, getCreateWrapper());
       const field = result.container.querySelector(`input[name="widget.inventoryCount"]`);
-      expect(field).toHaveAttribute("step", "any");
+      expect(field).toHaveAttribute("step", "1");
     });
 
     test("it renders number inputs with a step value when decimals are set", async () => {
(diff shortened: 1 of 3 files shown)


packages/react/spec/auto/inputs/PolarisAutoTextInput.spec.tsx
```diff
@@ -189,7 +189,7 @@ describe("PolarisAutoTextInput", () => {
     test("it renders number inputs with any step value when no decimals are set", async () => {
       result = render(<PolarisAutoNumberInput field="inventoryCount" />, getCreateWrapper());
       const field = result.container.querySelector(`input[name="widget.inventoryCount"]`);
-      expect(field).toHaveAttribute("step", "any");
+      expect(field).toHaveAttribute("step", "1");
```
danroberts, line 192: can you add a test here for a field that has a step of a decimal value?

danroberts (approved):
👍

packages/react/spec/auto/inputs/PolarisAutoTextInput.spec.tsx
MillanWangGadget: Immediately below this test there's a test with a field config decimal value of 2

alanko0511 (approved):

PR #583, opened 2024-07-30
Refactor columns property to support dot notation relationship path

alanko0511:
This PR refactor how we handle the `columns` property to support the dot notation relationship path. For example, when there is a "widget field has many gizmos" relationship, you can type `widget.edges.node.name` to grab the gizmo name. This PR also introduces a new kind of object that has a `header` that allows you to set a custom column header.

## PR Checklist


packages/react/src/utils.ts
```diff
@@ -543,22 +543,59 @@ export const isRoleAssignmentsArray = (value: ColumnValueType): value is RoleAss
...
 };
 
-export type RelatedFieldColumn = {
+export const isRelationshipColumn = (value: string) => {
+  const parts = value.split(".");
+  return parts.length === 2 || parts.length === 4;
+};
+
+export const maybeGetRelationshipFromColumnPath = (value: string) => {
+  const parts = value.split(".");
```
danroberts, line 552: I'm not sure this is the best strategy -- is it possible to not split the string but instead pass it as a path directly so that we could do `shop.owner.name` or any other valid gql out of the box?

alanko0511: But we have to build the selection map and find the related field based on the string alone. How do we get the related field without splitting the string?

danroberts: I think the selection is always the path objectified + true: 

`shop.owner.name` => `{shop: {owner: {name: true}}}`

danroberts: To me, that's the benefit of including the edges/nodes notation in the field string

PR #584, opened 2024-07-30
Allow hiding searching and/or pagination in AutoTable

alanko0511:
This PR introduces two new parameters in AutoTable, `search` and `paginate`, which allow you to hide the search bar and pagination buttons by setting them as `false`.

Fixes GGT-6881, GGT-6882

## PR Checklist

### DIFF
--- a/packages/react/src/auto/AutoTable.tsx
+++ b/packages/react/src/auto/AutoTable.tsx
@@ -24,4 +24,6 @@ export type AutoTableProps<
   filter?: Options["filter"];
   actions?: TableOptions["actions"];
   excludeActions?: TableOptions["excludeActions"];
+  searchable?: boolean;
+  paginate?: boolean;
 };
--- a/packages/react/src/auto/polaris/PolarisAutoTable.tsx
+++ b/packages/react/src/auto/polaris/PolarisAutoTable.tsx
@@ -87,6 +87,8 @@ const PolarisAutoTableComponent = <
   props: AutoTableProps<GivenOptions, SchemaT, FinderFunction, Options>
 ) => {
   const { onClick } = props;
+  const searchable = props.searchable ?? true;
+  const paginate = props.paginate ?? true;
 
   const [{ rows, columns, metadata, fetching, error, page, search, sort, selection }, refresh] = useTable<
     GivenOptions,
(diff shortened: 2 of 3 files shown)


packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -168,22 +169,24 @@ const PolarisAutoTableComponent = <
...
-        selected={1}
-        loading={fetching}
-        cancelAction={{ onAction: () => search.clear() }}
-        disabled={!!error}
-        // Search
-        queryValue={search.value}
-        onQueryChange={search.set}
-        onQueryClear={search.clear}
-      />
+      {!hideSearching && (
```
alanko0511, line 172: Right now we don't have any other filter options other than the full-text search, so we can just hide the whole thing.

packages/react/src/auto/AutoTable.tsx
```diff
@@ -24,4 +24,6 @@ export type AutoTableProps<
   filter?: Options["filter"];
   actions?: TableOptions["actions"];
   excludeActions?: TableOptions["excludeActions"];
+  hideSearching?: boolean;
```
danroberts, line 27: I think this should just be `search: boolean` and `paginate: boolean`

alanko0511: I'm not sure if people will think the `search` property can control the search query, not toggle the search bar. 🤔

danroberts: `searchable` would also work, but I think `search: boolean` is clear

packages/react/src/auto/AutoTable.tsx
```diff
@@ -24,4 +24,6 @@ export type AutoTableProps<
   filter?: Options["filter"];
   actions?: TableOptions["actions"];
   excludeActions?: TableOptions["excludeActions"];
+  search?: boolean;
```
danroberts, line 27: Discussed offline, we decided to go with `searchable`.

MillanWangGadget (approved):
MANUAL TESTING 
- 🍏 The 2 new props default to true and can properly hide the search and pagination components
- 🍏 Running bulk actions is still possible when you cannot paginate or search

Code changes look reasonable 👍

danroberts (approved):

PR #585, opened 2024-07-30
Update AutoTable actions api signature and add promoted actions system

MillanWangGadget:
- **UPDATES TO ACTIONS PROP**
  - Added optional `promoted` property so that actions can be accessed without clicking on the `more actions` button
    - Custom actions are by default promoted to false where actions will be put into the dropdown menu
    - Gadget model actions are promoted by default
  - renamed `callback` to `action`
  - Added `action: function | string` so that it is now possible to relabel Gadget bulk actions

### DIFF
--- a/packages/react/src/auto/hooks/useTableBulkActions.tsx
+++ b/packages/react/src/auto/hooks/useTableBulkActions.tsx
@@ -1,6 +1,6 @@
 import React, { useMemo } from "react";
 import deepEqual from "react-fast-compare";
-import { ActionCallback, TableOptions } from "../../useTableUtils/types.js";
+import { ActionCallback, promptComponent, TableOptions } from "../../useTableUtils/types.js";
 import { humanizeCamelCase } from "../../utils.js";
 
 export type ModelActionDetails =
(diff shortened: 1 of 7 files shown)


packages/react/src/auto/hooks/useTableBulkActions.tsx
```diff
@@ -15,14 +15,15 @@ export type ModelActionDetails =
   | {
       isGadgetAction: false;
       apiIdentifier: string;
-      render?: (recordIds: string[], records: any[]) => React.ReactNode;
+      onPromptComponent?: onPromptComponent;
```
danroberts, line 18: This feels like a mouthful. How about `promptContent`?

MillanWangGadget: Updated to be `promptContent` throughout

alanko0511 (approved):
🎩 
:shipit: 👌

MillanWangGadget:
Updated such that there is no longer the ability to pass in custom content to the default bulk action modal

danroberts (approved):

PR #587, opened 2024-07-31
refactor(autotable): move sort state into useTable hook

danroberts:
In order to pass around all the sort callbacks to the TableContext, it would be easier if the sort state was in the useTable hook. 

## PR Checklist

### DIFF
--- a/packages/react/src/useTableUtils/types.ts
+++ b/packages/react/src/useTableUtils/types.ts
@@ -35,6 +35,7 @@ export interface TableOptions {
   pageSize?: number;
   initialCursor?: string;
   initialDirection?: "forward" | "backward";
+  initialSort?: { [column: string]: SortOrder };
   columns?: (string | RelatedFieldColumn | CustomCellColumn)[];
   excludeColumns?: string[];
   actions?: (string | ActionCallback)[];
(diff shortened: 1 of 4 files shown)


packages/react/src/useTableUtils/types.ts
```diff
@@ -35,6 +35,7 @@ export interface TableOptions {
   pageSize?: number;
   initialCursor?: string;
   initialDirection?: "forward" | "backward";
+  initialSort?: { [column: string]: SortOrder };
```
gabeb03, line 38: There is already a `sort` prop in `useTable`. We should have either `sort` or `initialSort`.

gabeb03 (changes requested):
MANUAL TESTS

🍏 Included `initialSort` prop in `useTable`. The records were sorted as expected.
🍏 Included `initialSort` prop in `<PolarisAutoTable />`. The records were sorted as expected.

packages/react/src/useTableUtils/types.ts
danroberts: Updated to omit the `sort` param, although I find the resulting type less helpful than it could be.

alanko0511 (approved):
:shipit: 👌

PR #588, opened 2024-07-31
fix(autoform): fix testing flake waiting for lazy load suspense

danroberts:
The AutoForm component under test here is doing a React.lazy suspense,
and for some reason, it seems that possibly using the `screen` findBy
api results in a more reliable test result.
CleanShot 2024-07-31 at 12 49 11@2x

## PR Checklist


MillanWangGadget (approved):
- 🍏 6/6 on CI runs looks great. I had 1/5 and 1/6 PRs yesterday
- 🍏  Doing a lookup based on `screen` instead of the render result makes sense

I'm ok with this change only applying to the single flake test for now

alanko0511 (approved):
:shipit: 👌

PR #589, opened 2024-07-31
Added error for using actions without the API trigger in autoforms

MillanWangGadget:
- **UPDATE**
  - Added error for when an action without the API trigger is passed into an autoform
    - Most of the time, when the API client is settled, modelActions and GActions without the API trigger will have `action.type == "stubbedAction"`
    - When the trigger is removed from an action and the old hasTheTrigger action is still in the api client, the metadata will be checked for the presence of the trigger, and an error will be thrown if it is absent.


alanko0511:
Why do we still need to check whether the action has the API trigger, given the action function already tells you that the action doesn't exist?

MillanWangGadget:
> Why do we still need to check whether the action has the API trigger in the action metadata, given the action function already tells you that the action doesn't exist?

Without the metadata check, we will still be able to press submit on the form, but the GQL error will be `this action does not exist`, which is misleading

packages/react/src/metadata.tsx
```diff
@@ -315,6 +321,8 @@ export const useActionMetadata = (actionFunction: ActionFunction<any, any, any,
       action: actionName,
       includeRelatedFields: false,
     };
+  } else {
```
danroberts, line 324: This doesn't seem to me to be a 100% safe change -- are you sure that we want to throw an error here? Could there be other cases that were being silently handled by the permissive else that will now start throwing when they shouldnt?w

danroberts (approved):

alanko0511 (approved):

PR #590, opened 2024-07-31
Prevent showing time for dateTime fields with `includeTime:false`

MillanWangGadget:
- **UPDATE**
  - DateTime fields with `includeTime:false` will now only show the date
  - Showing the time requires that `includeTime:true` || `includeTime: undefined`

### DIFF
--- a/packages/react/src/useTableUtils/helpers.ts
+++ b/packages/react/src/useTableUtils/helpers.ts
@@ -173,6 +173,10 @@ export const getTableColumns = (spec: Pick<TableSpec, "fieldMetadataTree" | "tar
       column.relationshipType = firstField.fieldType as RelationshipType;
     }
 
+    if (targetField.configuration?.__typename === "GadgetDateTimeConfig") {
+      column.includeTime = targetField.configuration.includeTime;
+    }
+
     columns.push(column);
   }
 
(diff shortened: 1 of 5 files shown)


packages/react/src/useTableUtils/helpers.ts
```diff
@@ -171,6 +195,10 @@ export const getTableColumns = (spec: TableSpec) => {
         }
       }
 
+      if (field.fieldType === GadgetFieldType.DateTime) {
+        column.includeTime = (field.configuration as GadgetDateTimeConfig).includeTime;
+      }
```
alanko0511, line 200: Another way you could do is with this:
```ts
      if (field.configuration.__typename === 'GadgetDateTimeConfig') {
        column.includeTime = field.configuration.includeTime;
      }
```

That way you don't need to cast the config type.

MillanWangGadget: Good call. Updated

packages/react/src/useTableUtils/helpers.ts
```diff
@@ -135,6 +135,30 @@ export const getTableRows = (spec: TableSpec, records: GadgetRecord<any>[]) => {
   return records.map((record) => _recordToRow(spec, record));
 };
 
+const getColumnRelatedField = (targetColumn: string | RelatedFieldColumn | CustomCellColumn, field: FieldMetadataFragment) => {
```
danroberts, line 138: I'm confused why this ended up in here?

MillanWangGadget: Good catch. This was a leftover from getting through a merge conflict

danroberts (approved):

alanko0511 (approved):

PR #591, opened 2024-07-31
Add customizable title control to AutoForm

MillanWangGadget:
- **UPDATE**
  - Previously there was no way to control the title of the autoForm
  - This control has been added. If it is not given then the humanized operation name will be used as the default
  - Passing in an empty string will make it such that there is no title element rendered at all because the rendering it with an empty string will create some extra spacing

### DIFF
--- a/packages/react/src/auto/mui/MUIAutoForm.tsx
+++ b/packages/react/src/auto/mui/MUIAutoForm.tsx
@@ -1,7 +1,7 @@
 import type { ActionFunction } from "@gadgetinc/api-client-core";
 import type { GridProps } from "@mui/material";
 import { Grid, Skeleton, Typography } from "@mui/material";
-import React, { useMemo } from "react";
+import React from "react";
 import { FormProvider } from "react-hook-form";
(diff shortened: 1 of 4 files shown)


packages/react/src/auto/mui/MUIAutoForm.tsx
```diff
@@ -69,15 +69,18 @@ export const MUIAutoFormComponent = <
     },
   };
 
-  const humanizedOperationName = useMemo(() => humanizeCamelCase(action.operationName), [action.operationName]);
+  const formTitle = useMemo(
```
danroberts, line 72: I don't think this needs to be a memo -- I think we can just derive the title directly

MillanWangGadget: Updated to be used directly with no memo

danroberts (approved):

gabeb03 (approved):
MANUAL TESTS

🍏 Explicitly passing a title string renders the title correctly 
🍏 Excluding the `title` prop renders an auto-generated title
🍏 Passing `false` to the `title` prop hides the title

PR #593, opened 2024-07-31
Refactor rows and columns

danroberts:
In advance of switching over to a path notation for relationship fields, I think it makes sense to make some changes to the way the useTable rows and columns works. Right now, rows is mostly a representation of the record response, and the columns data structure hold a lot of information about i.e. the relationship and related model fields. 

Whatever is consuming the useTable has to implement some logic to i.e. extract the correct data from the edges/node type relationship gql response. 

After this change, that job is moved into the useTable hook, and `rows` now includes a more direct representation of what you would end up seeing in the final table (i.e. if a hasMany fields, the row entry would contain 'fieldName': ['first', 'second', 'third'] rather than a gql style edges/node data structure. 

## PR Checklist


MillanWangGadget (approved):
One leftover log statement but good otherwise. I didn't find any major behavioural changes 👍

PR #595, opened 2024-08-01
Refactor columns property to support dot notation column path

alanko0511:

## PR Checklist


packages/react/spec/auto/hooks/useTable.spec.tsx
```diff
@@ -759,12 +812,18 @@ const loadMockWidgetModelMetadataForRelationship = () => {
                       apiIdentifier: "id",
                       fieldType: "ID",
                       __typename: "GadgetModelField",
+                      configuration: {
```
danroberts, line 815: Why do we care about the field config?

alanko0511: We only care about the relationship config, but I used the `__typename` from the field config to determine whether the field is a relationship. Maybe I can just look at the field type instead.

danroberts: I think we already rely on the field type so it might be better to just couple ourselves to one thing -- I think the configs are a bit more likely to change.

packages/react/src/useTableUtils/helpers.ts
```diff
@@ -145,82 +130,202 @@ export const getTableColumns = (spec: TableSpec) => {
...
+  return `${columnPath}.${newSegment}`;
+};
+
+const isHasOneOrBelongsToField = (field: FieldMetadataFragment) => {
+  return field.fieldType === GadgetFieldType.HasOne || field.fieldType === GadgetFieldType.BelongsTo;
+};
+
+const isHasManyField = (field: FieldMetadataFragment) => {
+  return field.fieldType === GadgetFieldType.HasMany;
+};
```
MillanWangGadget, line 240: There are a couple of places throughout this file where you're doing these checks again instead of using these helpers

alanko0511: Ahh right, updated to use these functions instead.

MillanWangGadget:
CleanShot 2024-08-01 at 16 09 04@2x
Given `modelB.hasMany->modelA` and `<AutoTable model={api.modelB} columns={["modelAs.edges.node.belongToModelB"]} />`, this can appear where we just get `object` in the pill. Perhaps there should be an error for ending the dot.separated.path on a relationship

MillanWangGadget:
When you have hasMany paths like this, it'd be ideal if the errors could give suggestions on what it should look like
```js
"hasManyField.realFieldName" // Error: Field 'hasManyField.realFieldName ' does not exist in the model

// Below are all "Cannot read properties of undefined (reading 'toString')"
"hasManyField."
"hasManyField.edges" 
"hasManyField.edges." 
"hasManyField.edges.node"

```

danroberts (approved):
👍

MillanWangGadget:
Concerns with hasMany field pathing error messages have been resolved. Good to fix forward the `ending on relationship field` stuff

MillanWangGadget (approved):

PR #596, opened 2024-08-01
feat(autotable): Add a useTableContext hook for custom cell rendering

danroberts:
In order to enable maximum flexibility for custom cell renderers, there is now a AutoTableContext and useAutoTableContext hook which basically returns everything that useTable() returns. 

You can access this in your component to trigger search actions, get pagination information, or access the raw data across records/rows. 

## PR Checklist

### DIFF
--- a/packages/react/src/auto/AutoTableContext.ts
+++ b/packages/react/src/auto/AutoTableContext.ts
@@ -1,0 +1,30 @@
+import { GadgetRecord } from "@gadgetinc/api-client-core";
+import React, { useContext } from "react";
+import type { TableResult } from "src/useTableUtils/types.js";
+
+export type AutoTableContext = TableResult<GadgetRecord<any>>;
+
+/**
+ * React context that stores an instance of the metadata loaded for a particular form action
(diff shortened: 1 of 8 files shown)


packages/react/src/auto/AutoTableContext.ts
```diff
@@ -0,0 +1,30 @@
...
+ * React context that stores an instance of the metadata loaded for a particular form action
+ */
+
+export const AutoTableContext = React.createContext<AutoTableContext | undefined>(undefined);
+/**
+ * Get the current `metadata` object and `submit` function from React context
+ * Must be called within a component wrapped by the `<AutoForm>` component.
+ **/
+
+export const useAutoTableContext = () => {
```
alanko0511, line 17: Super small nit: it looks nicer not to have an empty line between the comment block and the variable.

alanko0511 (approved):
Could you add a storybook so that we can easily play around with it?
other than that :shipit: 👌

gabeb03 (approved):

PR #598, opened 2024-08-02
Added Upsert support in AutoForm

MillanWangGadget:
- **UPDATE**
  - Added upsert support to AutoForm
    - When findBy is passed in 
      - Is effectively an update action, but will create a record if given the ID that does not exist
      - Does not clear inputs upon submission
    - Without findBy
      - An ID field input will be rendered allowing users to set the ID
      - Clears all inputs upon submit
- **OTHER CONSIDERATIONS**
  - Consider if/how the `on` field in upsert actions should work in fields. Currently unsupported

## PR Checklist

### DIFF
--- a/packages/react/.changeset/little-bikes-dress.md
+++ b/packages/react/.changeset/little-bikes-dress.md
@@ -1,0 +1,27 @@
+---
+"@gadgetinc/react": minor
+---
+
+# Added support for `upsert` actions in AutoForm.
+
+- Models that have `create` and `update` actions get the `upsert` meta-action
+  - Upsert actions update records with given ID, and create a new record if the record does not exist
+  - The ID field is optional and omitting it will run the create action
+- Upsert actions can be used in `AutoForm` components in the same way that other actions are used
+- Usage
+  - With a given `findBy` value, the `upsert` action will run on the given specified ID value
+    - ```jsx
+      <AutoForm action={api.someModel.upsert} findBy="1" />
+      ```
+  - Without a given `findBy` value, an ID input field will appear in the form
+    - ```jsx
+      <AutoForm action={api.someModel.upsert} />
+      ```
+  - Just like other actions, the content of the form can be customized
+    - ```jsx
+      <AutoForm action={api.someModel.upsert}>
+        <AutoInput field="id" />
+        <AutoInput field="name" />
+        <AutoSubmit />
+      </AutoForm>
+      ```
--- a/packages/react/src/auto/AutoForm.ts
+++ b/packages/react/src/auto/AutoForm.ts
@@ -91,7 +91,10 @@ export const useFormFields = (
     );
 
     const includedObjectFields = objectFields.flatMap((objectField) =>
-      filterAutoFormFieldList((objectField.configuration as unknown as GadgetObjectFieldConfig).fields as any, options as any).map(
+      filterAutoFormFieldList((objectField.configuration as unknown as GadgetObjectFieldConfig).fields as any, {
+        ...(options as any),
+        isUpsertAction: true, // For upsert meta-actions, we allow IDs, and they are object fields instead of root level
+      }).map(
         (innerField) =>
           ({
             path: `${objectField.apiIdentifier}.${innerField.apiIdentifier}`,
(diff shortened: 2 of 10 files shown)


packages/react/src/auto/AutoForm.ts
```diff
@@ -146,14 +149,19 @@ export const useAutoForm = <
...
       props.defaultValues ??
       (action.type === "globalAction"
         ? {}
         : {
-            [modelApiIdentifier!]: record ?? (!operatesWithRecordId && metadata && isActionMetadata(metadata) && metadata?.defaultRecord),
-            id: "0", // The ID value will be replaced when sending the form to use the record found by `findBy`
+            [modelApiIdentifier!]:
+              record ??
+              (!(operatesWithRecordId || isUpsertWithFindBy) && metadata && isActionMetadata(metadata) && metadata?.defaultRecord),
+            id: findBy,
```
MillanWangGadget, line 164: In the GQL variable maker, this will stomp out `modelApiId.id`. We used to send `id:"0" ` for create actions which was a lil strange. Like this, they are sent as undefined

packages/react/src/metadata.tsx
```diff
@@ -345,7 +345,7 @@ export const useActionMetadata = (actionFunction: ActionFunction<any, any, any,
  */
 export const filterAutoFormFieldList = (
   fields: FieldMetadata[] | undefined,
-  options?: { include?: string[]; exclude?: string[] }
+  options?: { include?: string[]; exclude?: string[]; allowIdFieldType?: boolean }
```
danroberts, line 348: I think I'd prefer this option to refer to the fact that it's related to upserting -- we've fallen into a bad happen of passing these behavior control flags around which can be difficult to understand the implications of changing down the road. 

Maybe this would be `options?: { include?: string[]; exclude?: string[]; upsertAction?: boolean }`

packages/react/src/metadata.tsx
```diff
@@ -414,7 +414,11 @@ const isNotRelatedToSpecialModelFilter = (field: FieldMetadata) => {
 
 const specialModelKeys = new Set(["DataModel-Shopify-Shop"]);
 
-const isAcceptedFieldTypeFilter = (field: FieldMetadata) => acceptedAutoFormFieldTypes.has(field.fieldType);
+const isAcceptedFieldTypeFilter = (allowIdFieldType?: boolean) => {
```
danroberts, line 417: Again, I would prefer to avoid behavioral flag parameters. You could achieve the same effect by introducing a new function like

```
const isAcceptedUpsertFieldTypeFilter = (field: FieldMetadata) => isAcceptedFieldTypeFilter(field.fieldType) || field.fieldType === FieldType.Id);

File: packages/react/src/metadata.tsx (hunk 2 above)
Comment (MillanWangGadget) [reply packages/react/src/metadata.tsx]: Updated such that the option name refers to the upsert operation

File: packages/react/src/metadata.tsx (hunk 3 above)
Comment (MillanWangGadget) [reply packages/react/src/metadata.tsx]: Updated such that the filters for upsert/nonUpsert are different and selected via ternary

**Comment by danroberts** [comment]
Looks awesome. When [this](https://github.com/gadget-inc/js-clients/pull/607) merges could you add [a changeset entry?](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

**Review by alanko0511 — APPROVED** [review]

**Comment by MillanWangGadget** [comment]
> Looks awesome. When [this](https://github.com/gadget-inc/js-clients/pull/607) merges could you add [a changeset entry?](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

Updated to have a changeset entry

File: packages/react/.changeset/little-bikes-dress.md [hunk 4]
```diff
@@ -0,0 +1,5 @@
+---
+"@gadgetinc/react": minor
+---
+
+Added support for Upsert actions in AutoForm
```
danroberts, line 5: Can you add an example of how to use it so we know what to add to the docs?

MillanWangGadget: Updated with examples showing usage

packages/react/.changeset/little-bikes-dress.md
```diff
@@ -0,0 +1,27 @@
+---
+"@gadgetinc/react": minor
+---
+
+# Added support for `upsert` actions in AutoForm.
+
+- Models that have `create` and `update` actions get the `upsert` meta-action
```
danroberts, line 7: I'm not 100% sure this usage makes sense to me. I have to admit I can still get fuzzy on upsert sometimes, but I think what we want is something along the lines of: 

"given some attribute that I will always submit with my upsert action -- lets say, "email", or "shopifyProductId", either create the record in the gadget db, or update the existing record"

danroberts (approved):

packages/react/.changeset/little-bikes-dress.md
MillanWangGadget: That `on` parameter based stuff is currently not supported in AutoForm. Since that parameter is an array and array params are an open question right now, I'd prefer to revisit this note if we add support for array custom action params

PR #599, opened 2024-08-02
Switch how we call `render` function to a JSX style

alanko0511:
Previously, passing the JSX element into the custom cell renderer's `render` function was impossible when a hook was inside. This PR fixes that by treating the `render` property as a JSX element and invoking it in a JSX style. 

Fixes GGT-6904

## PR Checklist

### DIFF
--- a/packages/react/spec/auto/hooks/useTable.spec.tsx
+++ b/packages/react/spec/auto/hooks/useTable.spec.tsx
@@ -1,9 +1,11 @@
-import { renderHook } from "@testing-library/react";
+import { render, renderHook } from "@testing-library/react";
 import React from "react";
+import { useAction } from "../../../src/useAction.js";
 import { useTable } from "../../../src/useTable.js";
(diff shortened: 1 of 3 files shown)


packages/react/spec/auto/hooks/useTable.spec.tsx
```diff
@@ -527,6 +529,39 @@ describe("useTable hook", () => {
   });
 
   describe("custom cell renderer", () => {
+    // This test makes sure that the custom cell renderer is a valid React component by trying to render it.
+    // If it is not a valid React component, it will throw an error.
+    it("should be able to pass a JSX element with a hook inside", () => {
```
alanko0511, line 534: Without the fix introduced in the `helper.tsx` file, this test throws a lot of React-related errors.

danroberts (approved):

gabeb03 (approved):
:shipit:

PR #600, opened 2024-08-02
Use default selection when `columns` property does not exist

alanko0511:
Some model fields are unavailable to query but are still included in the GadgetMeta response. This PR uses the `defaultSelection` instead to build the `targetColumns` array.

Fixes GGT-6903

## PR Checklist

### DIFF
--- a/packages/react/cypress/component/auto/table/AutoTableSort.cy.tsx
+++ b/packages/react/cypress/component/auto/table/AutoTableSort.cy.tsx
@@ -72,31 +72,31 @@ describe("AutoTable - Sort", () => {
     cy.wait("@getWidgets").its("request.body.variables").should("deep.equal", { first: 50 });
 
     mockGetWidgetsWithDescendingSort();
-    cy.contains("Sortable Id").click({ force: true });
+    cy.contains("ID").click({ force: true });
     cy.wait("@getWidgetsWithDescendingSort");
     cy.get("@getWidgetsWithDescendingSort")
       .its("request.body.variables")
       .should("deep.equal", {
         first: 50,
         sort: {
-          sortableId: "Descending",
+          id: "Descending",
         },
       });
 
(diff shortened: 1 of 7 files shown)


packages/react/cypress/component/auto/table/AutoTableSort.cy.tsx
```diff
@@ -72,31 +72,31 @@ describe("AutoTable - Sort", () => {
     cy.wait("@getWidgets").its("request.body.variables").should("deep.equal", { first: 50 });
 
     mockGetWidgetsWithDescendingSort();
-    cy.contains("Sortable Id").click({ force: true });
+    cy.contains("ID").click({ force: true });
```
alanko0511, line 75: Now that we use the default selection map from the real API client, the fields have to be available in the real model. So I have to rename these two fields.

packages/react/spec/apis.ts
```diff
@@ -18,4 +18,13 @@ export const simpleExampleApi = new SimpleClient({ environment: "Development" })
...
+export const testApi = new TestClient({
+  environment: "Development",
+  authenticationMode:
+    "VITE_JS_CLIENTS_TEST_API_KEY" in (import.meta as any)
+      ? ({
+          apiKey: (import.meta as any).env.VITE_JS_CLIENTS_TEST_API_KEY,
+          dangerouslyAllowBrowserApiKey: true,
+        } as any)
+      : undefined,
+});
```
alanko0511, line 30: Shopify models aren't available with the `unauthenticated` role, so I updated this API client for us to test Shopify models in Storybook.

To test it, you need a `packages/react/.env` file inside this path. The content is:
```env
VITE_JS_CLIENTS_TEST_API_KEY="the-key"
```

danroberts (approved):

gabeb03 (approved):

PR #601, opened 2024-08-02
fix(auto-table): remove accidental polaris import from useTable hooks

danroberts:
We accidentally relied on a Polaris type import when building one of the base hooks, which results in an error when consuming from a project that doesn't have Polaris installed. 

Also added a new test to confirm that importing the base @gadgetinc/react components works without Polaris installed. 
## PR Checklist


alanko0511 (approved):
:shipit: 👌

gabeb03 (approved):

PR #605, opened 2024-08-12
Expose button props on Polaris auto submit

MillanWangGadget:
- **UPDATE**
  - Previously, we did not expose the Polaris button props on the PolarisAutoSubmitButton component. Now we can

### DIFF
--- a/packages/react/.changeset/chilled-drinks-mate.md
+++ b/packages/react/.changeset/chilled-drinks-mate.md
@@ -1,0 +1,19 @@
+---
+"@gadgetinc/react": patch
+---
+
+Updated PolarisAutoSubmit to expose Polaris button props
+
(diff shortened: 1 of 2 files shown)


alanko0511 (approved):
:shipit: 👌

packages/react/.changeset/chilled-drinks-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gadgetinc/react": patch
+---
+
+Updated PolarisAutoSubmit to expose Polaris button props
```
danroberts, line 5: Can you add an example here so we know what to update the docs with?

MillanWangGadget: Updated to show an example of Polaris button props being used in the button

danroberts (approved):

PR #606, opened 2024-08-13
Accept location object when initialising GadgetProvider for SSR support

alanko0511:
Currently, we use the `window.location` object to determine if the connection and set-up are ready. However, in some SSR frameworks like Remix.js, the `window` object is undefined, and instead, the `useLocation` hook is needed to access the location object.

This PR accepts passing the location object to our Shopify app provider and other fixes related to using the `window` object to enhance SSR support.

Example use case for a Remix.js app:
```jsx
// In root.jsx file
export default function App() {
  const location = useLocation();  // <-- create a location object

  return (
    <html lang="en">
      <body>
        <AppProvider i18n={enTranslations}>
          <GadgetProvider
            type={AppType.Embedded}
            shopifyApiKey={shopifyApiKey}
            api={new Client({ environment: "development" })}
            location={location}  // <-- then pass that to the provider
          >
            <Outlet />
          </GadgetProvider>
        </AppProvider>
      </body>
    </html>
  );
}
```

## PR Checklist

### DIFF
--- a/packages/react-shopify-app-bridge/src/Provider.tsx
+++ b/packages/react-shopify-app-bridge/src/Provider.tsx
@@ -11,6 +11,8 @@ export enum AppType {
   Embedded,
 }
 
+type ShopifyInstallState = NonNullable<NonNullable<Window["gadgetConfig"]>["shopifyInstallState"]>;
+
 /** Internal props used to create the right structure of providers */
 type GadgetProviderProps = {
   children: ReactNode;
@@ -21,6 +23,7 @@ type GadgetProviderProps = {
   host: string | undefined;
   isReady: boolean;
   isRootFrameRequest: boolean;
+  shopifyInstallState?: ShopifyInstallState;
 };
 
 // mutation which passes the session token back to Gadget to see if this shop is already installed and if not, install it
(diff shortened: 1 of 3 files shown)


airhorns (commented):
These changes make sense to me but can you add some tests that prove it renders in a windowless environment? I think you can do that by setting up a jest suite that uses the `jest-environment-node` env instead of `jest-environment-jsdom`. maybe we want to run the same suite or some subset of it?

alanko0511:
I just added some tests to prove that the Provider works in windowless environment by using the `ReactDOMServer` API:
https://react.dev/reference/react-dom/server

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -266,17 +280,31 @@ type ProviderLocation = {
...
-  const location = useMemo<ProviderLocation>(
-    () => ({
-      asPath: `${window.location.pathname}${window.location.search}`,
-      query: new URLSearchParams(window.location.search),
-    }),
-    []
-  );
+  const location = useMemo<ProviderLocation>(() => {
+    const pathname = _location?.pathname ?? window.location.pathname;
+    const search = _location?.search ?? window.location.search;
```
infiton, line 302: should we be throwing a better error here if window isn't defined and `_location` isn't passed?

alanko0511: Right I think it is nicer.

packages/react-shopify-app-bridge/src/Provider.tsx
```diff
@@ -266,17 +280,35 @@ type ProviderLocation = {
   asPath?: string;
 };
 
-export const Provider = ({ type, children, api }: { type?: AppType; children: ReactNode; shopifyApiKey: string; api: AnyClient }) => {
+export const Provider = ({
+  type,
+  children,
+  api,
+  location: _location,
```
airhorns, line 287: Are there any doc comments that need updating anywhere

airhorns (approved):
LGTM!

packages/react-shopify-app-bridge/src/Provider.tsx
alanko0511: I added some comments to the two new optional parameters. Let me see if there are docs on the gadget side.

infiton (approved):

PR #607, opened 2024-08-13
chore(react): introduce changeset tool for change tracking

danroberts:
A tool for managing a change log to the library

## PR Checklist


packages/react/package.json
MillanWangGadget, line 1: Is there are reason why these new Deps are added to the react package and not to the root? I was expecting to do `pnpm changeset` at the root, but am unable to since all of the changeset stuff is in the react package. 

The CLI UI then asks about all of the packages in the repo

MillanWangGadget (commented):
Is there a way to make the created filenames based on timestamps? They're currently random word sequences that end up sorted in non-chronological order

CleanShot 2024-08-13 at 15 38 15@2x


Would it also be possible to automatically include the version number in the file contents?
CleanShot 2024-08-13 at 15 39 33@2x

scott-rc (approved):
LGTM 👍 

I've had a good experience using changesets in ggt

MillanWangGadget (approved):

PR #612, opened 2024-08-19
Add mock response wait statements to autoform cypress test metadata requests

MillanWangGadget:
- UPDATE
  - It seems that there are some cases where the AutoForm metadata response is not awaited which leads to tests failing their assertions without failing CI
    - packages/react/cypress/component/auto/form/PolarisAutoDateTimePicker.cy.tsx was an example of this
  - Now those metadata requests are awaited so that failed assertions are properly noted in CI


MillanWangGadget:
No longer needed

PR #613, opened 2024-08-19
AGMD-24 Removed date-fns and date-fns-tz from all autocomponents

gabeb03:
The date-fns library is bloating the `@gadget/react` package; this PR replaces the date-fns imports with functions in `src/utils.ts`. The functions were copy-pasted from the [date-fns open source repo](https://github.com/date-fns/date-fns) and the [date-fns-tz repo](https://github.com/marnusw/date-fns-tz). 

@infiton, I left imports to date-fns in the test suite; my thinking is this will ensure that the ported functions behave like the original functions from date-fns. This means that date-fns is still installed in js-client. Does the package remain bloated because of this? 

## PR Checklist


packages/react/src/utils.ts
```diff
@@ -482,6 +482,312 @@ export const sortByProperty = <T>(arr: T[], property: keyof T, order: SortOrder
...
+  return date
+    .toLocaleString("en-US", {
+      year: "numeric",
+      month: "short",
+      day: "numeric",
+      hour: "numeric",
+      minute: "2-digit",
+      hour12: true,
+    })
+    .replace(/,(?=[^,]*$)/, "");
```
infiton, line 513: what's this for?

packages/react/src/utils.ts
```diff
@@ -482,6 +482,312 @@ export const sortByProperty = <T>(arr: T[], property: keyof T, order: SortOrder
...
+  try {
+    new Intl.DateTimeFormat(undefined, { timeZone: timeZoneString });
+    validIANATimezoneCache[timeZoneString] = true;
+    return true;
+  } catch (error) {
+    return false;
+  }
+};
+
+export const tzParseTimezone = (timezoneString: string | undefined, date: Date | number | undefined, isUtcDate?: boolean): number => {
```
infiton, line 557: is this from `date-fns` library? if so we should credit that in the comments

infiton:
> I left imports to date-fns in the test suite; my thinking is this will ensure that the ported functions behave like the original functions from date-fns

can you move `date-fns` to be a `devDependency` then?

packages/react/src/utils.ts
gabeb03: That regex replaces the last comma in the returned date string with nothing.

The table used `date-fns`'s `format` to formatted the string like so:
`June 19, 2024 12:00 AM`
Without the `.replace`, the built-in formatter formats the string like so:
`June 19, 2024, 12:00 AM`

This was causing tests to fail, so I removed the last comma.

packages/react/src/utils.ts
gabeb03: Good point: will update the PR

infiton (approved):
nice! thanks for doing that

MillanWangGadget (approved):
Good to go 👍

packages/react/src/utils.ts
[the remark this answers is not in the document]
gabeb03: Good point, I moved the code to `src/dateTimeUtils.ts`.

PR #615, opened 2024-08-21
fix(react): handle when hasMany relationship has null value

danroberts:
When checking to see if something was a RoleAssignment list when rendering tags, we didn't handle the case where not only could the value be `null`, but it could also be an array with `nulls` in the array, leading to a runtime error when that happened. 

## PR Checklist

### DIFF
--- a/packages/react/src/utils.ts
+++ b/packages/react/src/utils.ts
@@ -523,22 +523,25 @@ export type HasManyValueType = {
   }[];
 };
 
+export type Nullable<T> = T | null;
+
 export type ColumnValueType =
-  | string
-  | number
-  | boolean
-  | Date
+  | Nullable<string>
+  | Nullable<number>
+  | Nullable<boolean>
+  | Nullable<Date>
   | null
-  | string[]
+  | Nullable<string>[]
+  | Array<null>
   | RoleAssignmentsValueType[]
-  | FileValueType
-  | RichTextValueType
-  | ValueWithTypename
-  | HasManyValueType;
+  | Nullable<FileValueType>
+  | Nullable<RichTextValueType>
+  | Nullable<ValueWithTypename>
+  | Nullable<HasManyValueType>;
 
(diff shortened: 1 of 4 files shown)


packages/react/src/utils.ts
```diff
@@ -538,7 +539,8 @@ export type ColumnValueType =
 
 export const isRoleAssignmentsArray = (value: ColumnValueType): value is RoleAssignmentsValueType[] => {
   if (!Array.isArray(value) || value.length === 0) return false;
-  if (!value.every((item) => typeof item === "object" && "__typename" in item && item.__typename === "Role")) return false;
+  if (!value.every((item) => !(item === null) && typeof item === "object" && "__typename" in item && item.__typename === "Role"))
```
alanko0511, line 542: `!(item === null)` -> `item !== null` ?

alanko0511 (approved):

packages/react/src/utils.ts
```diff
@@ -538,7 +539,7 @@ export type ColumnValueType =
 
 export const isRoleAssignmentsArray = (value: ColumnValueType): value is RoleAssignmentsValueType[] => {
   if (!Array.isArray(value) || value.length === 0) return false;
-  if (!value.every((item) => typeof item === "object" && "__typename" in item && item.__typename === "Role")) return false;
+  if (!value.every((item) => item !== null && typeof item === "object" && "__typename" in item && item.__typename === "Role")) return false;
```
MillanWangGadget, line 542: If there's a null in the `value` array and `isRoleAssignmentsArray` returns false, we will go to an else block where we do 
```js
 formattedTags = value.map((tag) => tag.toString());
```

Over there, I think we need some null safety

danroberts: Added + a test for insurance. 👍 good catch.

MillanWangGadget (approved):
- 🍏 Having a mix of `null,undefined,definedValues` in the `values` array will only show the defined values as tags in the table

One code comment but good to ship otherwise 👍

PR #624, opened 2024-09-01
adds new internal authentication method that can be used for session …

infiton:
We want to add the ability for internal api clients in Gadget's app sandbox to be able to act either as the sys admin (the current state), or to act with the same identity as the session that is making the request. To do this I am changing the auth configuration for `internal` to take options to allow this functionality

This will be used in https://github.com/gadget-inc/gadget/pull/12813

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/ClientOptions.ts
+++ b/packages/api-client-core/src/ClientOptions.ts
@@ -92,9 +92,16 @@ export interface AuthenticationModeOptions {
   // Use no authentication at all, and get access only to the data that the Unauthenticated backend role has access to.
   anonymous?: true;
 
+  // @deprecated Use internal instead
+  internalAuthToken?: string;
+
   // @private Use an internal platform auth token for authentication
   // This is used to communicate within Gadget itself and shouldn't be used to connect to Gadget from other systems
-  internalAuthToken?: string;
+  internal?: {
+    authToken: string;
+    actAsSession?: boolean;
+    getSessionId?: () => Promise<string | undefined>;
+  };
 
   // @private Use a passed custom function for managing authentication. For some fancy integrations that the API client supports, like embedded Shopify apps, we use platform native features to authenticate with the Gadget backend.
   custom?: {
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -69,6 +69,7 @@ export interface GadgetConnectionOptions {
 export enum AuthenticationMode {
   BrowserSession = "browser-session",
   APIKey = "api-key",
+  Internal = "internal",
   InternalAuthToken = "internal-auth-token",
   Anonymous = "anonymous",
   Custom = "custom",
@@ -165,6 +166,8 @@ export class GadgetConnection {
     if (options) {
       if (options.browserSession) {
         this.enableSessionMode(options.browserSession);
+      } else if (options.internal) {
+        this.authenticationMode = AuthenticationMode.Internal;
       } else if (options.internalAuthToken) {
         this.authenticationMode = AuthenticationMode.InternalAuthToken;
       } else if (options.apiKey) {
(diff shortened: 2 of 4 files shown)


packages/api-client-core/src/ClientOptions.ts
```diff
@@ -92,9 +92,16 @@ export interface AuthenticationModeOptions {
...
+  internalAuthToken?: string;
+
   // @private Use an internal platform auth token for authentication
   // This is used to communicate within Gadget itself and shouldn't be used to connect to Gadget from other systems
-  internalAuthToken?: string;
+  internal?: {
+    authToken: string;
+    useSession?: boolean;
+    getSessionId?: () => Promise<string | undefined>;
+  };
```
airhorns, line 104: 👍

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -491,11 +506,25 @@ export class GadgetConnection {
...
+        this.authenticationMode == AuthenticationMode.Internal
+          ? this.options.authenticationMode!.internal!.authToken
+          : this.options.authenticationMode!.internalAuthToken!;
+
+      headers.authorization = "Basic " + base64("gadget-internal" + ":" + authToken);
+
+      if (this.authenticationMode == AuthenticationMode.Internal && this.options.authenticationMode!.internal!.useSession) {
+        headers["x-gadget-use-internal-session"] = "true";
+
+        const sessionId = await this.options.authenticationMode!.internal!.getSessionId?.();
```
airhorns, line 523: Tiny nit: what do you think about calling this `masquerade` or `impersonate` or something like that? I feel like it might be a little clearer to call the header that and the function that for the next person who comes along

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -491,11 +506,25 @@ export class GadgetConnection {
...
-      headers.authorization = "Basic " + base64("gadget-internal" + ":" + this.options.authenticationMode!.internalAuthToken!);
+    if (this.authenticationMode == AuthenticationMode.Internal || this.authenticationMode == AuthenticationMode.InternalAuthToken) {
+      const authToken =
+        this.authenticationMode == AuthenticationMode.Internal
+          ? this.options.authenticationMode!.internal!.authToken
+          : this.options.authenticationMode!.internalAuthToken!;
+
+      headers.authorization = "Basic " + base64("gadget-internal" + ":" + authToken);
+
+      if (this.authenticationMode == AuthenticationMode.Internal && this.options.authenticationMode!.internal!.useSession) {
```
airhorns, line 520: Are these two options separate because sometimes you want to use a session but you don't know the ID yet or something like that? I feel like it designs into existence passing the getter but not the boolean which is a bit confusing? Can we just use the presence of the get-me-the-id function as the indicator we should be doing this?

airhorns (approved):
LGTM after comments

packages/api-client-core/src/GadgetConnection.ts
infiton: > Can we just use the presence of the get-me-the-id function as the indicator we should be doing this?

we need both since when you call `api.asSession` we create a new api client, but I want to use a `getSessionId` function that was passed to the original `api` instance

packages/api-client-core/src/GadgetConnection.ts
infiton: `x-gadget-act-as-internal-session`?

packages/api-client-core/src/GadgetConnection.ts
infiton: this in Client.ts

```
get asSession(): Client {
        assert(this.options?.authenticationMode?.internal, "asSession can only be used for API clients using internal authentication")

        return new Client({
          ...this.options,
          authenticationMode: {
            internal: {
              ...this.options.authenticationMode.internal,
              asSession: true,
            }
          }
        })
      }```

PR #626, opened 2024-09-03
Update AutoForm error messages for bad actions

MillanWangGadget:
- **UPDATE**
  - Updated the error message for passing in non-actions to be more explicit in describing that AutoForm needs valid Gadget actions with API triggers

AGMD-48

### DIFF
--- a/packages/react/src/auto/AutoFormActionValidators.ts
+++ b/packages/react/src/auto/AutoFormActionValidators.ts
@@ -8,14 +8,14 @@ export const validateNonBulkAction = (action: ActionFunction<any, any, any, any,
 };
 
 const GadgetApiTriggerSpecId = "gadget/trigger/graphql_api";
-const ApiTriggerRequiredError = `Actions must have API triggers to be used in AutoForms`;
+export const TriggerableActionRequiredErrorMessage = `"action" must be a valid Gadget action with an API trigger to be used in AutoForms`;
 const validActionTypes = ["globalAction", "action"];
 
 export const validateTriggersFromApiClient = (action: ActionFunction<any, any, any, any, any> | GlobalActionFunction<any>) => {
   if (!validActionTypes.includes(action.type)) {
     // When the API client is built with an action without the API trigger, the type will be "stubbedAction"
     // action.type === "globalAction" | "action" // Only when the action has the API trigger when the api client is built
-    throw new Error(ApiTriggerRequiredError);
+    throw new Error(TriggerableActionRequiredErrorMessage);
   }
 };
 
@@ -32,7 +32,7 @@ export const validateTriggersFromMetadata = (metadata?: ActionMetadata | GlobalA
 
     const hasApiTrigger = triggersAsArray.some((trigger) => trigger.specID === GadgetApiTriggerSpecId);
     if (!hasApiTrigger) {
-      throw new Error(ApiTriggerRequiredError);
+      throw new Error(TriggerableActionRequiredErrorMessage);
     }
   }
 };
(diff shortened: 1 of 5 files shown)


packages/react/src/auto/polaris/PolarisAutoForm.tsx
```diff
@@ -32,8 +32,12 @@ export const PolarisAutoForm = <
   const { action, findBy } = props as AutoFormProps<GivenOptions, SchemaT, ActionFunc> &
     Omit<Partial<FormProps>, "action"> & { findBy: any };
 
+  if (!action) {
+    throw new Error("A valid Gadget action is required for AutoForm components");
+  }
+
   // Component key to force re-render when the action or findBy changes
-  const componentKey = `${action.modelApiIdentifier}.${action.operationName}.${findBy}`;
+  const componentKey = `${action.modelApiIdentifier ?? ""}.${action.operationName}.${findBy}`;
```
alanko0511, line 40: Why do we need this `??` fallback?

packages/react/src/auto/mui/MUIAutoForm.tsx
```diff
@@ -35,8 +35,12 @@ export const MUIAutoForm = <
 ) => {
   const { action, findBy } = props as MUIAutoFormProps<GivenOptions, SchemaT, ActionFunc> & { findBy: any };
 
+  if (!action) {
+    throw new Error("A valid Gadget action is required for AutoForm components");
+  }
```
alanko0511, line 40: Could we just reuse the same `ApiTriggerRequiredError` message here?

packages/react/src/auto/AutoFormActionValidators.ts
```diff
@@ -8,7 +8,7 @@ export const validateNonBulkAction = (action: ActionFunction<any, any, any, any,
 };
 
 const GadgetApiTriggerSpecId = "gadget/trigger/graphql_api";
-const ApiTriggerRequiredError = `Actions must have API triggers to be used in AutoForms`;
+export const ApiTriggerRequiredError = `"action" must be a valid Gadget action with an API trigger to be used in AutoForms`;
```
alanko0511, line 11: `ApiTriggerRequiredError` sounds like an actual error class, but instead is an error message.
How about `ApiTriggerRequiredErrorMessage`?

alanko0511 (approved):

packages/react/src/auto/polaris/PolarisAutoForm.tsx
MillanWangGadget: Global actions don't have `modelApiIdentifier`, and they previously had `undefined` in the key string

packages/react/src/auto/mui/MUIAutoForm.tsx
MillanWangGadget: Updated to use the reference the same error message string

packages/react/src/auto/AutoFormActionValidators.ts
MillanWangGadget: Renamed to `TriggerableActionRequiredErrorMessage`

danroberts (commented):
Can you add a changelog and make this a minor version bump because it's kind of a breaking change ( in that we're throwing errors now)

MillanWangGadget:
> Can you add a changelog and make this a minor version bump because it's kind of a breaking change ( in that we're throwing errors now)

Changeset with minor version bump has been added 👍

danroberts (approved):

PR #627, opened 2024-09-03
Fixed a bug with AutoTable where custom cell renderers with duplicate headers would have duplicate columns content

MillanWangGadget:
- **UPDATE**
  - Columns with custom cell renderers were previously added to a record with the header as the key
    - If duplicate custom cell headers were used, the earliest custom cell renderers would be overwritten due to the duplicate header used as a record key
  - Now, the custom cell renderer index is incorporated into the record key, thus duplicate header named custom cell renderers no longer overwrite each other

### DIFF
--- a/packages/react/src/use-table/helpers.tsx
+++ b/packages/react/src/use-table/helpers.tsx
@@ -138,16 +138,23 @@ export const getTableSelectionMap = (spec: TableSpec) => {
   return selectionMap;
 };
 
-export const getTableRows = (spec: Pick<TableSpec, "fieldMetadataTree" | "targetColumns">, records: GadgetRecord<any>[]) => {
-  return records.map((record, index) => recordToRow(spec, record, index));
+export const getTableRows = (
+  spec: Pick<TableSpec, "fieldMetadataTree" | "targetColumns">,
+  columns: TableColumn[],
+  records: GadgetRecord<any>[]
+) => {
+  return records.map((record, recordIndex) => recordToRow(spec, columns, record, recordIndex));
 };
 
 export const getTableColumns = (spec: Pick<TableSpec, "fieldMetadataTree" | "targetColumns">) => {
   const columns: TableColumn[] = [];
 
-  for (const targetColumn of spec.targetColumns) {
+  for (const [i, targetColumn] of spec.targetColumns.entries()) {
     if (isCustomCellColumn(targetColumn)) {
+      const identifier = crypto.randomUUID();
       columns.push({
+        identifier,
+        render: targetColumn.render,
         header: targetColumn.header,
         field: targetColumn.header,
         type: "CustomRenderer",
@@ -161,6 +168,7 @@ export const getTableColumns = (spec: Pick<TableSpec, "fieldMetadataTree" | "tar
     const { firstField, targetField, isHasMany, isHasOneOrBelongsTo } = getFieldInformationByColumnPath(spec.fieldMetadataTree, columnPath);
 
     const column: TableColumn = {
+      identifier: columnPath,
       header: header ?? firstField.name,
       field: columnPath,
       type: targetField.fieldType,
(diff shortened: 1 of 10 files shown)


packages/react/src/use-table/helpers.tsx
```diff
@@ -192,10 +192,14 @@ const recordToRow = (spec: Pick<TableSpec, "fieldMetadataTree" | "targetColumns"
...
 
-  for (const targetColumn of spec.targetColumns) {
+  const customCellColumnHeaders = spec.targetColumns.filter(isCustomCellColumn).map((column) => column.header ?? "");
+  for (const [i, targetColumn] of spec.targetColumns.entries()) {
     if (isCustomCellColumn(targetColumn)) {
       const CellComponent = targetColumn.render;
-      row[targetColumn.header] = <CellComponent record={record} index={index} />;
+      // Index is included here to allow for multiple custom cells with the same header
+      const isDuplicate = customCellColumnHeaders.filter((header) => header === targetColumn.header).length > 1;
+      const rowKey = isDuplicate ? `${targetColumn.header}.${i}` : targetColumn.header;
```
MillanWangGadget, line 201: Here, I only append the index when the current custom renderer is a dupe because I don't like the idea of changing the `useList` response if it is not strictly needed

packages/react/src/use-table/helpers.tsx
```diff
@@ -192,10 +192,14 @@ const recordToRow = (spec: Pick<TableSpec, "fieldMetadataTree" | "targetColumns"
...
   };
 
-  for (const targetColumn of spec.targetColumns) {
+  const customCellColumnHeaders = spec.targetColumns.filter(isCustomCellColumn).map((column) => column.header ?? "");
+  for (const [i, targetColumn] of spec.targetColumns.entries()) {
     if (isCustomCellColumn(targetColumn)) {
       const CellComponent = targetColumn.render;
-      row[targetColumn.header] = <CellComponent record={record} index={index} />;
+      // Index is included here to allow for multiple custom cells with the same header
+      const isDuplicate = customCellColumnHeaders.filter((header) => header === targetColumn.header).length > 1;
```
alanko0511, line 200: This `recordToRow` function will run on each record. Is there a way to use a map or other approaches to reduce the need to do the `filter` in each for loop?

alanko0511 (approved):

packages/react/src/use-table/helpers.tsx
MillanWangGadget: Good call, updated to make a map of occurrences and avoid repeated filtering

packages/react/src/use-table/helpers.tsx
```diff
@@ -138,16 +138,24 @@ export const getTableSelectionMap = (spec: TableSpec) => {
...
 };
 
 export const getTableColumns = (spec: Pick<TableSpec, "fieldMetadataTree" | "targetColumns">) => {
   const columns: TableColumn[] = [];
 
-  for (const targetColumn of spec.targetColumns) {
+  for (const [i, targetColumn] of spec.targetColumns.entries()) {
     if (isCustomCellColumn(targetColumn)) {
+      // Fallback to index if crypto.randomUUID is not available. All browsers have crypto.randomUUID, but test envs don't
+      const identifier = crypto.randomUUID?.() ?? i;
```
danroberts, line 155: I think it's a bad idea to have the tests and production code working differently -- maybe we could shim with something like https://github.com/ai/nanoid to keep the package size down?

MillanWangGadget: I added this to shim the UUIDs because they need to be deterministic as we also need to reference them in the tests. Pretty cool stuff 


```
  let timesUsedCryptoRandomUUID = 0;
  let uuids: string[] = [];

  const shimCryptoRandomUUID = () => {
    Object.defineProperty(globalThis, "crypto", {
      value: {
        randomUUID: () => {
          timesUsedCryptoRandomUUID++;
          const mostRecentCryptoRandomUUID = `000-000-000-000-${timesUsedCryptoRandomUUID}`;
          uuids.push(mostRecentCryptoRandomUUID);
          return mostRecentCryptoRandomUUID;
        },
      },
    });
  };

  beforeEach(() => {
    shimCryptoRandomUUID();
  });
```

danroberts (approved):

PR #628, opened 2024-09-03
Updated useTable `select` property to override default selection instead of merging with it

MillanWangGadget:
- UPDATE
  - Updated useTable `select` property to override default selection instead of merging with it
  - This is a breaking change for those who are using `select` property and are relying on the default values merged into it

### DIFF
--- a/packages/react/.changeset/kind-plants-rule.md
+++ b/packages/react/.changeset/kind-plants-rule.md
@@ -1,0 +1,6 @@
+---
+"@gadgetinc/react": patch
+---
+
+(Breaking change)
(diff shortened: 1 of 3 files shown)


alanko0511 (approved):

MillanWangGadget:
CleanShot 2024-09-04 at 09 42 53@2x
This is the only place in the docs where I found a reference to the select property. I've added this sentence in a separate PR that will only ship after the next version release

https://github.com/gadget-inc/templates
Here, there are no templates that use `AutoTable` with the select prop,

danroberts (approved):

PR #629, opened 2024-09-04
refactor(api-client-core): add operation builder plans to types

danroberts:
We want to remove the dependency from @gadgetinc/react -> api-client-core, because it's really easy to end up with multiple incompatible versions of api client core installed.

A main reason for the dependency is that @gadgetinc/react needs to access the operationBuilder functions in order to create query plans to pass to urql in hooks.

This PR attaches the plan builder functions as part of the action function prototype, alongside the other metadata about the action that is already present in the prototype.

This PR is a chain, currently consisting of:

1. This change to js-clients to introduce the plan to the types
2. This change to add the plan functions to the client instance when building it: https://github.com/gadget-inc/gadget/pull/12843
3. [in progress] having the react hooks consume these plans rather than import them from the api-client-core lib

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/GadgetFunctions.ts
+++ b/packages/api-client-core/src/GadgetFunctions.ts
@@ -23,7 +23,7 @@ export interface FindOneFunction<OptionsT, SelectionT, SchemaT, DefaultsT> {
   selectionType: SelectionT;
   optionsType: OptionsT;
   schemaType: SchemaT | null;
-  plan: <Options extends OptionsT>(fieldValue: string, options?: LimitToKnownKeys<Options, OptionsT>) => GQLBuilderResult;
+  plan?: <Options extends OptionsT>(fieldValue: string, options?: LimitToKnownKeys<Options, OptionsT>) => GQLBuilderResult;
 }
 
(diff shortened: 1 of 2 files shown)


packages/api-client-core/src/GadgetFunctions.ts
```diff
@@ -162,6 +172,7 @@ export interface GetFunction<OptionsT, SelectionT, SchemaT, DefaultsT> {
   selectionType: SelectionT;
   optionsType: OptionsT;
   schemaType: SchemaT | null;
+  plan?: <Options extends OptionsT>(options?: LimitToKnownKeys<Options, OptionsT>) => GQLBuilderResult;
 }
 
 export interface GlobalActionFunction<VariablesT> {
```
MillanWangGadget, line 178: Should `GlobalActionFunction` get a `plan`?

danroberts: yup. good catch.

MillanWangGadget (approved):
Now with path property
- Global actions
- model actions
- bulk model actions
- upsert meta actions
- Find_x_ actions

I checked all of the types in `GadgetFunctions.ts` and they all seem to get the plan property directly, or through type inheritance

alanko0511 (approved):

PR #630, opened 2024-09-04
Context based loading spinner on AutoSubmit in AutoForm

MillanWangGadget:
- **UPDATE**
  - Previously
    - Having `AutoSubmit` as a child of an AutoForm would not show the spinner while the form was submitting
    - This was because that functionality was controlled by a prop in the one-liner approach
  - NOW
    - `AutoSubmit` has its loading state controlled by a hook thus it will properly show the spinner while it is submitting
    - The state can be overridden if desired

### DIFF
--- a/packages/react/src/auto/polaris/submit/PolarisAutoSubmit.tsx
+++ b/packages/react/src/auto/polaris/submit/PolarisAutoSubmit.tsx
@@ -2,6 +2,7 @@ import type { ButtonProps } from "@shopify/polaris";
 import { Button } from "@shopify/polaris";
 import type { ReactNode } from "react";
 import React from "react";
+import { useAutoFormMetadata } from "../../AutoFormContext.js";
 
 /**
  * Button for submitting the AutoForm values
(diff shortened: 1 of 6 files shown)


packages/react/src/auto/polaris/submit/PolarisAutoSubmit.tsx
```diff
@@ -15,11 +16,13 @@ import React from "react";
 export const PolarisAutoSubmit = (
   props: {
     children?: ReactNode;
-    isSubmitting?: boolean;
```
danroberts, line 18: I think we should continue to export this for backwards compatibility, and fallback to the context value if it is not provided

MillanWangGadget: Good call. Updated to keep the original prop around and fallback to the context value

danroberts (approved):

alanko0511 (approved):

PR #633, opened 2024-09-05
fix(use-action-form): relationships that have only readonly fields

danroberts:
If you use useActionForm and select some fields from a relationship field and tag them as ReadOnly, it will still try to submit an empty value to the relationship field.

If your relationship field contains only ReadOnly values, it will consider the relationship field ReadOnly as well.

## PR Checklist

### DIFF
--- a/packages/react/spec/useActionForm.spec.tsx
+++ b/packages/react/spec/useActionForm.spec.tsx
@@ -3144,6 +3144,114 @@ describe("utils", () => {
     `);
   });
 
+  test("it removes relationship fields that are entirely read-only", async () => {
+    const result = applyDataMask({
+      modelApiIdentifier: "answer",
+      data: {
+        answer: {
+          text: "Answer create",
+          question: {
+            text: "question text",
+          },
+        },
+      },
+      select: {
+        id: true,
+        text: true,
+        question: {
+          text: "ReadOnly",
+        },
+      },
+    });
+
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "answer": {
+          "text": "Answer create",
+        },
+      }
+    `);
+  });
+
+  test("it does not remove relationship fields that are not entirely read-only", async () => {
+    const result = applyDataMask({
+      modelApiIdentifier: "answer",
+      data: {
+        answer: {
+          text: "Answer create",
+          question: {
+            foo: "bar",
+            text: "question text",
+          },
+        },
+      },
+      select: {
+        id: true,
+        text: true,
+        question: {
+          foo: true,
+          text: "ReadOnly",
+        },
+      },
+    });
+
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "answer": {
+          "question": {
+            "foo": "bar",
+          },
+          "text": "Answer create",
+        },
+      }
+    `);
+  });
+
+  test("it does not remove relationship fields when only some of their fields are read-only", async () => {
+    const result = applyDataMask({
+      modelApiIdentifier: "answer",
+      data: {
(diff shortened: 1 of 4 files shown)


packages/react/spec/useActionForm.spec.tsx
```diff
@@ -3144,6 +3144,69 @@ describe("utils", () => {
...
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "answer": {
+          "text": "Answer create",
+        },
+      }
+    `);
+  });
+
+  test("it does not remove relationship fields that are not entirely read-only", async () => {
```
MillanWangGadget, line 3176: In this test case, I did some experimenting and found this

```
    const result = applyDataMask({
      modelApiIdentifier: "answer",
      data: {
        answer: {
          text: "Answer create",
          question: {
            foo: "bar",
            text: "question text",
            a: {
              b: {
                c: "NICE",
                d: "abc",
              },
              e: "xyz", // See below
            },
          },
        },
      },
      select: {
        id: true,
        text: true,
        question: {
          foo: true,
          text: "ReadOnly",
          a: {
            b: {
              c: true,
              d: "ReadOnly",
            },
            e: true, // Without this, the whole "a" object is lost. If readonly. the whole "question" object is gone
          },
        },
      },
    });
```

packages/react/src/use-action-form/utils.ts
```diff
@@ -460,6 +460,23 @@ export function applyDataMask(opts: {
...
+        readOnlyPaths.push(currentPrefix);
+      } else {
+        // Recursively check nested relationships
+        checkRelationshipFields(obj[field], currentPrefix);
+      }
+    }
+  };
+
+  checkRelationshipFields(select, modelApiIdentifier);
+
```
MillanWangGadget, line 479: `readOnlyPaths` ends up as a list of paths like `answer,question.text`. which would be the key of a flattened object. Perhaps we could get around the thing I observed in the tests with a flattened object approach

packages/react/src/use-action-form/utils.ts
```diff
@@ -460,6 +460,23 @@ export function applyDataMask(opts: {
 
   const readOnlyPaths = getReadOnlyPaths(select, modelApiIdentifier);
 
+  // Check for relationship fields that are entirely read-only, including nested relationships
+  const checkRelationshipFields = (obj: any, prefix = "") => {
```
pistachiobaby, line 464: This makes total sense to me, however I still wonder if we should also just make computed fields a special case in the reshaper?

Thoughts?

danroberts: My thought is that we need both -- regardless of computed fields, I don't think we should be trying to submit relationship data when all the selected fields of the relationship are read only.

MillanWangGadget (commented):
Within the select object

`key: false` - Still includes that field
`key: "ReadOnly"` - Omits the field


Perhaps `getReadOnlyPaths`-> `if (obj[key] === "ReadOnly") {` should also consider when it is set to false in the select

danroberts:
> Within the select object
> 
> `key: false` - Still includes that field `key: "ReadOnly"` - Omits the field
> 
> Perhaps `getReadOnlyPaths`-> `if (obj[key] === "ReadOnly") {` should also consider when it is set to false in the select

That's already the existing behavior of the useActionForm, so that could potentially be a breaking change. I'm not sure that key: false really matters, you really shouldn't do that in a selection.

MillanWangGadget (approved):
Not wanting a breaking change with `key:false` makes sense. Good to ship 👍

PR #634, opened 2024-09-05
Release @gadgetinc/react v0.17.0

MillanWangGadget:
Releasing the next version of @gadgetinc/react

Note that the version boost for the peer dependency on the other packages should not affect any of their behaviour. They don't have any references to AutoComponent stuff, thus they can't be affected by the breaking changes

### DIFF
--- a/packages/react-bigcommerce/package.json
+++ b/packages/react-bigcommerce/package.json
@@ -41,6 +41,6 @@
   "peerDependencies": {
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
-    "@gadgetinc/react": "^0.16.0"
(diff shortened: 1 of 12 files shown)


danroberts (commented):
The last time I updated the react version, I also had to update the packages that had direct dependencies on it or ran into issues when consuming it gadget app side.

MillanWangGadget:
> The last time I updated the react version, I also had to update the packages that had direct dependencies on it or ran into issues when consuming it gadget app side.

Boosted the Gadget mono repo to consume v17 here
https://github.com/gadget-inc/gadget/pull/12892

danroberts (approved):

PR #635, opened 2024-09-09
AutoForm `defaultValues` included in request if not included in include/exclude

MillanWangGadget:
- PREVIOUSLY
  - AutoForm `include/exclude` props would filter out what would be sent in the submission request. This filtering would apply to explicit passed in default values, making it impossible to set a non-editable default value
- NOW
  - The `defaultValues` property will be used in the request regardless of the `include` / `exclude` properties

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoFormDefaultValues.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoFormDefaultValues.cy.tsx
@@ -26,6 +26,48 @@ describeForEachAutoAdapter("AutoForm - Default model field values", ({ name, ada
     cy.get(`input[name="part.notes"]`).should("have.value", "");
   });
 
+  it("can set default values in props for excluded fields and send them to the server", () => {
+    cy.mountWithWrapper(
+      <AutoForm action={api.part.create} defaultValues={{ part: { name: "test record" } }} exclude={["name"]} />,
+      wrapper
+    );
+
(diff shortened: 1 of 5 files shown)


packages/react/cypress/component/auto/form/AutoFormDefaultValues.cy.tsx
```diff
@@ -26,6 +26,46 @@ describeForEachAutoAdapter("AutoForm - Default model field values", ({ name, ada
...
   });
 
+  it("can set default values in props for excluded fields and send them to the server", () => {
+    cy.mountWithWrapper(
+      <AutoForm action={api.part.create} defaultValues={{ part: { name: "test record" } }} exclude={["name"]} />,
+      wrapper
+    );
+
+    cy.get(`input[name="part.count"]`).should("have.value", "");
+    cy.get(`input[name="part.notes"]`).should("have.value", "");
```
alanko0511, line 36: Could you add one check to make sure the `name` field doesn't exist?

alanko0511 (approved):
Could you also add one in the storybook to showcase this?

danroberts (approved):

packages/react/cypress/component/auto/form/AutoFormDefaultValues.cy.tsx
MillanWangGadget: Updated to include these cases into storybook and to check that the name input is gone in the Cypress test

PR #636, opened 2024-09-10
Updated `AutoInput` to accept an optional `label?: string` input

MillanWangGadget:
- Updated `AutoInput` to accept an optional `label?: string` input
- In many cases, I needed to explicitly put it later in a React element so that the prop would not be overridden by the default. 
- Any label you pass in explicitly should always be used by the AutoInput

### DIFF
--- a/packages/react/.changeset/clean-flies-provide.md
+++ b/packages/react/.changeset/clean-flies-provide.md
@@ -1,0 +1,5 @@
+---
+"@gadgetinc/react": patch
+---
+
+Updated `AutoInput` to accept an optional `label?: string` input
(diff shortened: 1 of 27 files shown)


alanko0511 (approved):

danroberts (commented):
Looks good -- can you add a basic test for it, doesn't need to cover every input, just a few common ones.

MillanWangGadget:
> Looks good -- can you add a basic test for it, doesn't need to cover every input, just a few common ones.

Updated to add a MUI and Polaris based test to prove that the label can be overridden

danroberts (approved):

PR #638, opened 2024-09-11
Send dev harness event from react hook when it is a stubbed action

alanko0511:
This PR sends a dev harness event when the action is a stubbed action (i.e. no API trigger)

Gadget side to consume this new event type: https://github.com/gadget-inc/gadget/pull/12894

## PR Checklist

### DIFF
--- a/packages/react/spec/useAction.spec.tsx
+++ b/packages/react/spec/useAction.spec.tsx
@@ -805,4 +805,38 @@ describe("useAction", () => {
 
     expect(result.current[0]).toBe(beforeObject);
   });
+
+  test("sends a stubbedActionError event when the action is stubbed", () => {
+    let eventDispatched: CustomEvent | undefined;
+
+    globalThis.addEventListener("gadget:devharness:stubbedActionError", (event) => {
+      eventDispatched = event as CustomEvent;
+    });
+
+    const { result, rerender } = renderHook(
+      () =>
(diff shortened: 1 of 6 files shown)


packages/react/spec/useAction.spec.tsx
```diff
@@ -805,4 +805,38 @@ describe("useAction", () => {
...
+      () =>
+        useAction({
+          // @ts-expect-error intentionally passing the wrong type
+          type: "stubbedAction",
+          reason: "MissingApiTrigger",
+          dataPath: "fakePath",
+          functionName: "fakeFunction",
+          modelApiIdentifier: "fakeModel",
+          variables: {},
+        }),
```
alanko0511, line 826: These parameters are unavailable today, so I need to make a fake object to trigger an event. I'll update the tests once we have shipped the change to the Gadget side.

MillanWangGadget (commented):
I think the useBulkAction needs the same treatment since bulk actions also becom stubbs. 
See ` api.autoTableTest.bulkNoTriggerAction`
CleanShot 2024-09-12 at 13 44 09@2x

MillanWangGadget (approved):
- 🍏 Properly pops the dev harness for no-trigger modelActions, bulkModelActions, and global actions

danroberts (approved):

PR #640, opened 2024-09-12
Fixed bug with AutoForm include/exclude system which prevented submission due to validations getting applied to hidden fields

MillanWangGadget:
- PREVIOUSLY
  - If you have a required field on a model action and it was hidden from the AutoForm with `include` or `exclude`, the form could not be submitted
  - This was due to the required validation applying to the field 
  - This issue did not have a corresponding error message
- NOW
  - Input field validations are now only run on the fields that are shown when using `include` and `exclude`
  - If a required field is excluded, the validation will occur server side, which allows for the field to be set in the model action code

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoForm.cy.tsx
@@ -235,10 +235,8 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
   });
 
   it("can render a rich text editor for markdown content", async () => {
-    cy.mountWithWrapper(<AutoForm action={api.widget.create} include={["name", "inventoryCount", "description"]} />, wrapper);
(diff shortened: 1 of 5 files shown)


packages/react/cypress/component/auto/form/AutoForm.cy.tsx
```diff
@@ -235,10 +235,8 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
   });
 
   it("can render a rich text editor for markdown content", async () => {
-    cy.mountWithWrapper(<AutoForm action={api.widget.create} include={["name", "inventoryCount", "description"]} />, wrapper);
+    cy.mountWithWrapper(<AutoForm action={api.widget.create} include={["description"]} />, wrapper);
```
MillanWangGadget, line 238: Previously, `"name", "inventoryCount"` were only included in this test because they were required. This test passing without those req'd fields proves that those validations don't run on non-shown fields 😎

danroberts (approved):

alanko0511 (approved):

PR #642, opened 2024-09-13
Add functionName and actionApiIdentifier when sending stubbed action event

alanko0511:
This fixes bug introduced in https://github.com/gadget-inc/js-clients/pull/638 where a wrong action API identifier being sent when it is a bulk action.

## PR Checklist


MillanWangGadget (approved):
Works well with related PR in Gadget main repo. `functionName` and `actionApiIdentifier` will only be different for bulk actions, and checking their equality seems like a reasonable way to tell if you're looking at a global action.

danroberts (approved):

PR #643, opened 2024-09-16
Update auto component error messages

alanko0511:
This PR updates the error message when an invalid model or action is passed to AutoForm or AutoTable.

## PR Checklist

### DIFF
--- a/packages/react/src/auto/AutoFormActionValidators.ts
+++ b/packages/react/src/auto/AutoFormActionValidators.ts
@@ -1,5 +1,6 @@
 import type { ActionFunction, GlobalActionFunction } from "@gadgetinc/api-client-core";
 import type { ActionMetadata, GlobalActionMetadata } from "../metadata.js";
+import type { useAutoForm } from "./AutoForm.js";
 
 export const validateNonBulkAction = (action: ActionFunction<any, any, any, any, any> | GlobalActionFunction<any>) => {
   if (action.isBulk) {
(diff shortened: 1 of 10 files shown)


packages/react/src/auto/mui/MUIAutoForm.tsx
```diff
@@ -36,8 +36,12 @@ export const MUIAutoForm = <
 ) => {
   const { action, findBy } = props as MUIAutoFormProps<GivenOptions, SchemaT, ActionFunc> & { findBy: any };
 
+  if (!("action" in props)) {
```
danroberts, line 39: Any way we can centralize this in the hook?

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -65,6 +66,11 @@ export const PolarisAutoTable = <
   props: AutoTableProps<GivenOptions, SchemaT, FinderFunction, Options>
 ) => {
   const { model } = props;
+
+  if (!model) {
```
danroberts, line 70: Same as above, any way we can move this to the hook?

packages/react/src/auto/AutoFormActionValidators.ts
```diff
@@ -8,14 +9,28 @@ export const validateNonBulkAction = (action: ActionFunction<any, any, any, any,
 };
 
 const GadgetApiTriggerSpecId = "gadget/trigger/graphql_api";
-export const TriggerableActionRequiredErrorMessage = `"action" must be a valid Gadget action with an API trigger to be used in AutoForms`;
+
+export const MissingActionPropErrorMessage = "Specify a valid Gagdet action to use AutoForm";
+export const InvalidActionErrorMessage = `"action" is not a valid Gadget action`;
```
MillanWangGadget, line 14: NIT: maybe mention here that the action is used in AutoForm

packages/react/src/auto/AutoTableValidators.ts
```diff
@@ -0,0 +1 @@
+export const InvalidModelErrorMessage = `"model" is not a valid Gadget model`;
```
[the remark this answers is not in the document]
alanko0511: The error message is now thrown inside the `useTable` hook, so I don't think specifying the word AutoTable is really correct. 🤔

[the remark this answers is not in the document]
MillanWangGadget: Good point, and I guess mentioning `useTable` might end up being confusing for people using `AutoTable` who aren't using the hook directly

packages/react/src/auto/polaris/PolarisAutoTable.tsx
```diff
@@ -65,7 +65,10 @@ export const PolarisAutoTable = <
   props: AutoTableProps<GivenOptions, SchemaT, FinderFunction, Options>
 ) => {
   const { model } = props;
-  const componentKey = `${[model.findMany.namespace, model.findMany.modelApiIdentifier].join("_")}AutoTable`;
+
+  const componentKey = model
```
danroberts, line 69: This seems a bit weird to me -- why do we sent the key to undefined here?

alanko0511: We throw the error in the hook, which is below this area. So at that moment the `model` is undefined.

danroberts: Hmm then with apologies I think I might change my previous feedback about moving the errors into the hooks -- I think we want to validate the inputs as soon as we can so that we don't have to do things like this.

packages/react/src/auto/mui/MUIAutoForm.tsx
```diff
@@ -53,6 +49,9 @@ export const MUIAutoFormComponent = <
   props: MUIAutoFormProps<GivenOptions, SchemaT, ActionFunc>
 ) => {
   const { record: _record, action, findBy, ...rest } = props as MUIAutoFormProps<GivenOptions, SchemaT, ActionFunc> & { findBy: any };
+
+  validateAutoFormProps(props);
+
```
MillanWangGadget, line 54: This should be moved up to where you removed `if (!action)` just like polaris AutoForm
Otherwise, when `action === undefined` we will get the `can't reference operationName on undefined` before we get here

MillanWangGadget (approved):

danroberts (approved):

PR #645, opened 2024-09-17
Removed deprecated Polaris `Frame` usage from AutoTable

MillanWangGadget:
- Previously, the toast system for bulk action completion in AutoTable required a `Frame` wrapper to house the Toast component. This `Frame` unfortunately had some extra side effects that would cause the table to take on extra height, disrupting existing app user interfaces
- The frame component has been removed and the toast based bulk action feedback system has been reverted to the banner in modal approach

### DIFF
--- a/packages/react/.changeset/fair-files-repair.md
+++ b/packages/react/.changeset/fair-files-repair.md
@@ -1,0 +1,5 @@
+---
+"@gadgetinc/react": patch
+---
+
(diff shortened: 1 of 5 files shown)


packages/react/.changeset/fair-files-repair.md
MillanWangGadget, line 1: Forgot to add this in the prev PR to fix that bug so added here

PR #648, opened 2024-09-19
Allow `useGlobalAction` callback to take no arguments

MillanWangGadget:
- UPDATE
  - Some Gadget template apps have type errors because they have `useGlobalAction` and no parameter in the callback.
    - `Expected 1-2 arguments, but got 0`
  - This actually works and should be allowed by the type 

 
This will no longer have a type error with the default code snip
CleanShot 2024-09-19 at 16 26 40@2x

### DIFF
--- a/packages/react/src/utils.ts
+++ b/packages/react/src/utils.ts
@@ -127,14 +127,18 @@ export type RequiredKeysOf<BaseType> = Exclude<
  * Includes the data result object and a function for running the mutation.
  **/
 export type ActionHookResult<Data = any, Variables extends AnyVariables = AnyVariables> = RequiredKeysOf<Variables> extends never
-  ? [
-      ActionHookState<Data, Variables>,
-      (variables?: Variables, context?: Partial<OperationContext>) => Promise<ActionHookState<Data, Variables>>
-    ]
-  : [
-      ActionHookState<Data, Variables>,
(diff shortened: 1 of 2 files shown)


MillanWangGadget:
> It seems to me that we should be covering this case in react/src/utils when we do:
>     maybe there is something wrong with that type?
>  

Good call. It turns out that the `Record<string,never>` that global actions pass into `ActionHookResult`was causing the callback type to have a mandatory `variables` argument parameter. Since globalActions cannot have required params, I've updated the return type to always have an optional `variables` argument

packages/react/src/utils.ts
```diff
@@ -127,14 +127,18 @@ export type RequiredKeysOf<BaseType> = Exclude<
...
-      ActionHookState<Data, Variables>,
-      (variables?: Variables, context?: Partial<OperationContext>) => Promise<ActionHookState<Data, Variables>>
-    ]
-  : [
-      ActionHookState<Data, Variables>,
-      (variables: Variables, context?: Partial<OperationContext>) => Promise<ActionHookState<Data, Variables>>
-    ];
+  ? ActionHookResultWithOptionalCallbackVariables<Data, Variables>
+  : ActionHookResultWithRequiredCallbackVariables<Data, Variables>;
+
```
MillanWangGadget, line 132: Note that I tried doing a ternary chain here, but this causes a lot of type errors

danroberts (approved):

alanko0511 (approved):

PR #650, opened 2024-09-23
Don't re-export PolarisAutoForm as PolarisAutoForm

airhorns:
We already export the PolarisAutoForm component as AutoForm from the auto/polaris/index.ts file. We were accidentally also exporting it as PolarisAutoForm, which is a bit confusing for onlookers like the AI. Let's just export it once under a canonical name!

I also changed the skeleton to be explicitly exported as well using the existing naming convention.


alanko0511 (approved):
:shipit: 👌

danroberts (approved):

PR #651, opened 2024-09-24
Hide hasMany input on join model when using actions on models with hasManyThrough fields

MillanWangGadget:
- UPDATE
  - Updated AutoForm HasManyThrough fields so that no inputs are shown. 
  - They would previously show a hasMany field for the join model, which corresponds to the API, but was confusing in an AutoForm context
  - Adding support for `hasManyThrough` inputs will come in a follow up


packages/react/src/metadata.tsx
```diff
@@ -355,6 +390,7 @@ export const filterAutoFormFieldList = (
   fields: FieldMetadata[] | undefined,
   options?: { include?: string[]; exclude?: string[]; isUpsertAction?: boolean }
 ): FieldMetadata[] => {
+  console.log("fields :", fields);
```
danroberts, line 393: 🔥

packages/react/src/metadata.tsx
```diff
@@ -408,6 +447,12 @@ const specialModelKeys = new Set(["DataModel-Shopify-Shop"]);
 const isAcceptedFieldType = (field: FieldMetadata) => acceptedAutoFormFieldTypes.has(field.fieldType);
 const isAcceptedUpsertFieldType = (field: FieldMetadata) => field.fieldType === FieldType.Id || isAcceptedFieldType(field);
 
+const isNotJoinModelHasManyField = (field: FieldMetadata) =>
+  field.fieldType !== FieldType.HasMany ||
```
danroberts, line 451: I think this would be more clearly expressed in the positive -- isJoinModelHasManyField because some of the double negatives confused me for a second. 

```
(field.fieldType === FieldType.HasMany &&
    field.configuration.__typename === "GadgetHasManyConfig" &&
    field.configuration.isJoinModelHasManyField)
 ```
 
 ...
 
 ```
  // Remove `hasMany` fields that emerge from `hasManyThrough` fields that are not actually model fields
  subset = subset.filter(field => !isJoinModelHasManyField(field));
 ```

MillanWangGadget: Updated such that the filter works as a positive check

packages/react/src/metadata.tsx
MillanWangGadget: removed

danroberts (approved):

alanko0511 (approved):

PR #652, opened 2024-09-24
Added HasManyThrough field support to AutoTable

MillanWangGadget:
- UPDATE
  - Added HasManyThrough field support to AutoTable
  - This is functionally identical to hasMany fields

### DIFF
--- a/packages/react/spec/auto/hooks/useTable.spec.tsx
+++ b/packages/react/spec/auto/hooks/useTable.spec.tsx
@@ -327,7 +327,7 @@ describe("useTable hook", () => {
 
     it("should use default display field if the column string is a relationship field", async () => {
       const result = getUseTableResult({
-        columns: ["name", "hasMany", "hasOne", "belongsTo"],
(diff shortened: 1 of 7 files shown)


packages/react/spec/auto/hooks/useTable.spec.tsx
```diff
@@ -461,7 +461,7 @@ describe("useTable hook", () => {
       `);
     });
 
-    it.each(["password", "hasManyThrough"])("should throw an error if the selected field type is %s", async (fieldType) => {
+    it.each(["password"])("should throw an error if the selected field type is %s", async (fieldType) => {
```
danroberts, line 464: I think we might also want a test like the one above that confirms that the hasmanythrough is loading data correctly.

MillanWangGadget: Good call. Updated with some tests that show `hasManyThrough` in action with the `useTable` hook

alanko0511 (approved):

danroberts (approved):

PR #660, opened 2024-09-29
Update InvalidRecordError comment to match jsDoc syntax

alanko0511:
This PR updates the existing comment block so it becomes a valid jsDoc comment and is available in type declaration.

## PR Checklist


MillanWangGadget (approved):
Good catch

PR #661, opened 2024-09-30
Prevent AutoForm validation with custom children to prevent being blocked unshown required fields

MillanWangGadget:
- **UPDATE**
  - About 2 weeks ago, a change was shipped that prevented local field validations from running on fields that were not rendered as a result of `include/exclude` AutoForm properties
  - When AutoForm components have custom children all validations would still run, which would prevent form submission if one of the fields was required and not registered to react hook form in a child component. 
  - Now, having any child components in `AutoForm` will now avoid all local validations so that the requests are always sent to the server to be validated there. 
    - I don't think we can reasonably parse the child components to determine which validations should be ran


packages/react/src/auto/AutoForm.ts
```diff
@@ -145,13 +147,18 @@ export const useAutoForm = <
...
   const isDeleteAction = metadata && isActionMetadata(metadata) && metadata.action.isDeleteAction;
   const isGlobalAction = action.type === "globalAction";
   const operatesWithRecordId = !!(metadata && isActionMetadata(metadata) && metadata.action.operatesWithRecordIdentity);
   const modelApiIdentifier = action.type == "action" ? action.modelApiIdentifier : undefined;
   const isUpsertMetaAction = metadata && isActionMetadata(metadata) && fields.some((field) => field.metadata.fieldType === FieldType.Id);
   const isUpsertWithFindBy = isUpsertMetaAction && !!findBy;
+  const hasCustomChildren = !!props.children;
+  const fieldPathsToValidate = hasCustomChildren
+    ? [] // With custom children, do not validate fields before sending them to avoid blocking submissions due to missing required fields
+    : fields.map(({ path }) => path);
```
alanko0511, line 161: Should we wrap this inside a `useMemo` with a `[hasCustomChildren]` as the dep array?

packages/react/.changeset/strong-turkeys-worry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@gadgetinc/react": patch
+---
+
+- Added an error message to AutoForm when passing in `include/exclude` options alongside child components
```
alanko0511, line 5: We're throwing a hard error, so maybe it should be "Throw an error" instead.

alanko0511 (approved):

packages/react/src/auto/AutoForm.ts
MillanWangGadget: Good call. Updated to add a `useMemo` to avoid remapping on fields

packages/react/.changeset/strong-turkeys-worry.md
MillanWangGadget: Updated to say that a new error is thrown

danroberts (approved):

PR #663, opened 2024-10-01
Fixed the `findBy` object parameter for the `AutoForm` component such that related records are correctly shown and forms can be properly submitted

MillanWangGadget:
- **UPDATE**
  - Previously, when the `AutoForm` findBy prop was an object representing a unique field, form submissions were broken and relationship fields did not properly load related model data
    - The form submission was broken because we had included the whole object as the id value on submission. This became a type error that could not be shown without advanced `react-hook-form` techniques
    - The relationship fields were broken because we did not have a system for translating those object values into IDs for the model containing the action. The only way to know this is through a lookup, and a hook has been added to do so
  - NOW, the findBy property should work as expected when using objects like `{uniqueFieldApiId: "uniqueValue"}`
- **TESTING**
  - I've added a namespace with inter-related models containing a lot of unique fields here. Great for storybook
    - https://js-clients-test.gadget.app/edit/development/model/nu84_-GxRnkx/
  - Cases of interest 
    - findBy object keys that are not unique model fields
    - Valid unique model field, but the value does not exist in the model
    - Relationships and showing related model records

### DIFF
--- a/packages/react/src/auto/AutoFormActionValidators.ts
+++ b/packages/react/src/auto/AutoFormActionValidators.ts
@@ -1,5 +1,7 @@
 import type { ActionFunction, GlobalActionFunction } from "@gadgetinc/api-client-core";
-import type { ActionMetadata, GlobalActionMetadata } from "../metadata.js";
+import { FieldMetadata, type ActionMetadata, type GlobalActionMetadata } from "../metadata.js";
+import { RecordIdentifier } from "../use-action-form/types.js";
+import { isPlainObject } from "../use-action-form/utils.js";
 import type { useAutoForm } from "./AutoForm.js";
 
 export const validateNonBulkAction = (action: ActionFunction<any, any, any, any, any> | GlobalActionFunction<any>) => {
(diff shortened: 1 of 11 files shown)


packages/react/src/auto/AutoFormActionValidators.ts
```diff
@@ -44,19 +45,48 @@ export const validateAutoFormProps = (props: Parameters<typeof useAutoForm>[0])
...
+  // When an action's API trigger is removed but the referenced API client still uses the old api-triggered action, the presence of the trigger shall be checked in the metadata
+
+  const triggersAsArray =
+    (metadata.__typename === "GadgetGlobalAction"
+      ? metadata.triggers
+      : metadata.__typename === "GadgetModel"
+      ? metadata.action.triggers
+      : []) ?? [];
+
+  const hasApiTrigger = triggersAsArray.some((trigger) => trigger.specID === GadgetApiTriggerSpecId);
```
infiton, line 60: btw I think this might be the source of another issue I've seen with using autoform for auth: the special auth triggers are api triggers but don't have this specific spec id

MillanWangGadget: May I follow up on allowing API auth triggers separately to prevent this `findBy` updated from getting too large?

infiton: yah of soruce; sorry this was just an fyi

packages/react/src/auto/AutoFormActionValidators.ts
```diff
@@ -44,19 +45,48 @@ export const validateAutoFormProps = (props: Parameters<typeof useAutoForm>[0])
...
+
+export const validateFindByObjectWithMetadata = (
+  fields?: readonly { path: string; metadata: FieldMetadata }[],
+  findBy?: RecordIdentifier
+) => {
+  if (!findBy || typeof findBy === "string") {
+    return;
+  }
+
+  if (typeof findBy !== "object" || Array.isArray(findBy) || Object.keys(findBy).length !== 1) {
```
infiton, line 74: we have an `isPlainObject` that might work for this?

MillanWangGadget: Good call. Updated to use `isPlainObject` which means there is no need to check `isArray` since that case is prevented with the prototype checks 👍

infiton (approved):

alanko0511 (approved):

PR #665, opened 2024-10-04
Updated AutoForm enum inputs such that their search value does not get pre-populated

MillanWangGadget:
- UPDATE
  - Pre-populating the search bar for enum inputs makes it such that the options are pre-filtered once you click on the input. 
    - It would be better to see all available options when you clock the input
  - Now, the enum input search input is always initialized as empty

### DIFF
--- a/packages/react/src/auto/hooks/useEnumInputController.tsx
+++ b/packages/react/src/auto/hooks/useEnumInputController.tsx
@@ -21,7 +21,7 @@ export const useEnumInputController = (props: {
     control,
     name: path,
   });
-  const [searchValue, setSearchValue] = useState(typeof fieldProps.value === "string" ? fieldProps.value : "");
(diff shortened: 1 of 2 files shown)


packages/react/src/auto/hooks/useEnumInputController.tsx
```diff
@@ -21,7 +21,7 @@ export const useEnumInputController = (props: {
     control,
     name: path,
   });
-  const [searchValue, setSearchValue] = useState(typeof fieldProps.value === "string" ? fieldProps.value : "");
```
MillanWangGadget, line 24: Unless you are using the `defaultValues` prop in AutoForm, this would always be initially null, which made this empty anyways in most circumstances

alanko0511 (approved):

PR #667, opened 2024-10-08
useActionForm supports unlinking a belongsTo

infiton:
When working on nested forms using `useActionForm` I found that if we set the `id = nuil` on a belongs to relationship we don't call the api correctly to unlink the model

This PR resolves that

## PR Checklist

### DIFF
--- a/packages/react/src/use-action-form/utils.ts
+++ b/packages/react/src/use-action-form/utils.ts
@@ -374,7 +374,9 @@ export const reshapeDataForGraphqlApi = async (client: AnyClient, defaultValues:
           return getParentRelationshipFieldGraphqlApiInput({ input, result });
         case "BelongsTo":
           return inputHasId
-            ? inputHasMoreFields
+            ? input["id"] === null
+              ? { _link: null }
+              : inputHasMoreFields
               ? { update: { id: input["id"], ...rest } }
               : { _link: input["id"] }
             : inputUpdateId // input has no id, but this path was found in the updates object, so we need to delete it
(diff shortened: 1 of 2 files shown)


packages/react/src/use-action-form/utils.ts
```diff
@@ -357,7 +357,9 @@ export const reshapeDataForGraphqlApi = async (client: AnyClient, defaultValues:
           return getParentRelationshipFieldGraphqlApiInput({ input, result });
         case "BelongsTo":
           return inputHasId
-            ? inputHasMoreFields
+            ? input["id"] === null
+              ? { _link: null }
+              : inputHasMoreFields
```
MillanWangGadget, line 362: The only difference I can imagine here is that something like 
`belongsToField: {id:null, otherField:"non-null"}` 
will now unlink the record instead of trying to update a record without an ID. This feels more correct. 

Since the other inputs are ignored when `input["id"] === null && inputHasMoreFields`, should there be  an error/warning to indicate that?

MillanWangGadget (approved):
Change makes sense to me. One lil question but good otherwise

pistachiobaby (approved):
🚢 

🙏 📿 trust in our tests 📿 🙏

packages/react/src/use-action-form/utils.ts
infiton: >trying to update a record without an ID

How would you update a record without an ID?

MillanWangGadget: >> trying to update a record without an ID

> How would you update a record without an ID?

I mean that you are no longer able to attempt an update with `id:null`. It would be a GQL error anyways, even with unique fields on the related

PR #672, opened 2024-10-15
computed view endpoint for api client (GGT-7201)

mkobetic:
This is the counterpart to https://github.com/gadget-inc/gadget/pull/13298

It adds operation builder and runner for computed view as well as generic ComputeView function types and interfaces with metadata. It introduces two variants "with variables" and "without variables" to keep things simpler. The types probably could be merged, but I had enough trouble sorting the types out as is, we can try in a follow up.

Chunk of the work was debugging and tweaking the various utility types to allow TS to infer types correctly. It was a crazy learning curve for me, it's quite possible that I still got some things wrong here. I introduced second versions of the utility types as some of the tweaks broke the dependent packages that use the original ones (react). Not all tweaks are likely necessary for the purposes of this particular PR, but they all are addressing actual issues in the older types, so probably worth keeping. I'll explain more about each inline.

If this passes the review I'll need some guidance on how to roll this out.

## PR Checklist

### DIFF
--- a/packages/api-client-core/spec/default-selection.spec.ts
+++ b/packages/api-client-core/spec/default-selection.spec.ts
@@ -1,11 +1,90 @@
-import type { AssertTrue, IsExact } from "conditional-type-checks";
-import type { DefaultSelection } from "../src/types.js";
-import type { AvailableTestSchemaSelection } from "./TestSchema.js";
+import type { AssertTrue, Has, IsExact } from "conditional-type-checks";
+import type { FieldSelection } from "src/FieldSelection.js";
+import type { ComputedViewFunctionWithVariables, ComputedViewFunctionWithoutVariables } from "src/GadgetFunctions.js";
+import type { AllFieldsSelected, AvailableSelection2, DefaultSelection, DefaultSelection2 } from "../src/types.js";
+import type { AvailableTestSchemaSelection, TestSchema } from "./TestSchema.js";
+import type { ExpandRecursively } from "./helpers.js";
 
-type _NullDefault = DefaultSelection<AvailableTestSchemaSelection, { select: null }, { num: true }>;
-type _TestDefaultsNullToTheDefault = AssertTrue<IsExact<_NullDefault, { num: true }>>;
+test("DefaultSelection", () => {
+  type nullDefault = DefaultSelection<AvailableTestSchemaSelection, { select: null }, { num: true }>;
+  type _TestDefaultsNullToTheDefault = AssertTrue<IsExact<nullDefault, { num: true }>>;
 
-type _NonNullDefault = DefaultSelection<AvailableTestSchemaSelection, { select: { num: false; str: true } }, { num: true }>;
-type _TestRespectsTruthySelections = AssertTrue<IsExact<_NonNullDefault, { num: false; str: true }>>;
+  type undefinedDefault = DefaultSelection<AvailableTestSchemaSelection, { select: undefined }, { num: true }>;
+  type _TestDefaultsUndefinedToTheDefault = AssertTrue<IsExact<undefinedDefault, { num: true }>>;
 
-test("true", () => undefined);
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  type undefinedDefault2 = DefaultSelection<AvailableTestSchemaSelection, {}, { num: true }>;
+  type _TestDefaultsUndefinedToTheDefault2 = AssertTrue<IsExact<undefinedDefault2, { num: true }>>;
+
+  type anyDefault = DefaultSelection2<AvailableTestSchemaSelection, { select: any }, { num: true }>;
+  type _TestDefaultsAnyToTheDefault = AssertTrue<IsExact<anyDefault, { num: true }>>;
+
+  type nonDefault = DefaultSelection<AvailableTestSchemaSelection, { select: { num: false; str: true } }, { num: true }>;
+  type _TestRespectsTruthySelections = AssertTrue<IsExact<nonDefault, { num: false; str: true }>>;
+
+  type nonDefault2 = DefaultSelection<AvailableTestSchemaSelection, { select: { num: false } }, { num: true }>;
+  type _TestRespectsTruthySelections2 = AssertTrue<IsExact<nonDefault2, { num: false }>>;
+});
+
+test("AvailableSelection", () => {
+  type availableTestSchemaSelection = AvailableSelection2<TestSchema>;
+  type _TestAvailableTestSchemaSelection = Has<FieldSelection, availableTestSchemaSelection>;
+
+  type testType = { a: number; b: { c: string; d: { e: boolean } } };
+  type availableSelection = AvailableSelection2<testType>;
+  type _TestAvailableSelection = AssertTrue<
+    IsExact<
+      availableSelection,
+      {
+        a?: boolean | null | undefined;
+        b?:
+          | {
+              c?: boolean | null | undefined;
+              d?:
+                | {
+                    e?: boolean | null | undefined;
+                  }
+                | undefined;
+            }
+          | undefined;
+      }
+    >
+  >;
+});
+
+test("AllFieldsSelected", () => {
+  type testType = { a: number; b: { c: string; d: { e: boolean } } };
+  type availableSelection = AvailableSelection2<testType>;
+  type allFieldsSelected = ExpandRecursively<AllFieldsSelected<availableSelection>>;
+  type _TestAllFieldsSelected = AssertTrue<IsExact<allFieldsSelected, { a: true; b: { c: true; d: { e: true } } }>>;
+});
+
+test("ComputedViewFunction", () => {
+  type testType = { a: number; b: { c: string; d: { e: boolean } } };
+
+  const f: ComputedViewFunctionWithoutVariables<testType> = () => ({} as any);
+
+  const resultWithDefaultSelection = f();
+  type _TestResultWithDefaultSelection = AssertTrue<IsExact<Awaited<typeof resultWithDefaultSelection>, testType>>;
+
+  const resultWithDefaultSelection2 = f({});
+  type _TestResultWithDefaultSelection2 = AssertTrue<IsExact<Awaited<typeof resultWithDefaultSelection2>, testType>>;
+
+  const resultWithDefaultSelection3 = f({ select: null });
+  type _TestResultWithDefaultSelection3 = AssertTrue<IsExact<Awaited<typeof resultWithDefaultSelection3>, testType>>;
+
+  const resultWithSelection = f({ select: { a: true } });
+  type _TestResultWithSelection = AssertTrue<IsExact<Awaited<typeof resultWithSelection>, { a: number }>>;
+
+  const resultWithSelection2 = f({ select: { a: false, b: { c: true } } });
+  type _TestResultWithSelection2 = AssertTrue<IsExact<Awaited<typeof resultWithSelection2>, { b: { c: string } }>>;
+
+  type vars = { n?: number; b: boolean };
+  const f2: ComputedViewFunctionWithVariables<vars, testType> = () => ({} as any);
+
+  const resultWithDefaultSelection4 = f2({ b: false });
+  type _TestResultWithDefaultSelection4 = AssertTrue<IsExact<Awaited<typeof resultWithDefaultSelection4>, testType>>;
+
+  const resultWithSelection5 = f({ select: { a: false, b: { c: true } } });
+  type _TestResultWithSelection5 = AssertTrue<IsExact<Awaited<typeof resultWithSelection5>, { b: { c: string } }>>;
+});
--- a/packages/api-client-core/spec/operationBuilders.spec.ts
+++ b/packages/api-client-core/spec/operationBuilders.spec.ts
@@ -1,6 +1,7 @@
 import {
   actionOperation,
   backgroundActionResultOperation,
+  computedViewOperation,
   enqueueActionOperation,
   findManyOperation,
   findOneByFieldOperation,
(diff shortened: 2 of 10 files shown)


packages/api-client-core/src/GadgetFunctions.ts
```diff
@@ -206,3 +207,14 @@ export interface GlobalActionFunction<VariablesT> {
 
 export type AnyActionFunction = ActionFunctionMetadata<any, any, any, any, any, any> | GlobalActionFunction<any>;
 export type AnyBulkActionFunction = ActionFunctionMetadata<any, any, any, any, any, true>;
+
+export interface ComputedViewFunction<VariablesT extends VariablesOptions, SelectionsT extends FieldSelection, ResultT> {
```
mkobetic, line 211: Let's use DefaultSelection, allows restricting it to fewer fields.

packages/api-client-core/src/GadgetFunctions.ts
```diff
@@ -206,3 +207,14 @@ export interface GlobalActionFunction<VariablesT> {
...
+  (variables: VariablesT): Promise<ResultT>;
```
mkobetic, line 212: (variables: VariablesT, options?: { select?: SelectionsT }): Promise<Select<ResultT, DefaultSelection<DefaultSelectionsT, SelectionsT, DefaultSelectionsT>>>

mkobetic: @airhorns so I gave this a go and it seems I have to declare this as follows for TS to stop complaining about things:
```ts
export interface ComputedViewFunction<VariablesT, DefaultSelectionsT extends FieldSelection, SelectionsT extends DefaultSelectionsT, ResultT> {
  (variables: VariablesT, options?: { select?: SelectionsT }): Promise<Select<ResultT, DefaultSelection<DefaultSelectionsT, { select?: SelectionsT | null }, DefaultSelectionsT>>>;
  /// ...
}
```
The `extends FieldSelection` bit is required because `Select<a,b>` requires b to extend FieldSelection. So far so good. The part that doesn't make sense to me is that I had to do `SelectionsT extends DefaultSelectionsT` because `DefaultSelection<a,b,c>` requires b and c to `extend a`. But that seems backwards, DefaultSelectionsT has the full selection details and SelectionT has some subset so it should be `DefaultSelectionT extends SelectionT` no? But if I do that I can't make DefaultSelection<> to work. This is all very meta and makes my head hurt, so quite likely I'm misunderstanding something.

mkobetic: Hm, maybe I misunderstood what the DefaultSelectionT type should be, looking at this example https://github.com/gadget-inc/js-clients/blob/main/packages/api-client-core/spec/TestSchema.ts#L75. It seems that DefaultSelectionT should have all fields, but optional. So in that sense SelectionT could be extension of DefaultSelectionT. I thought DefaultSelectionT would have all fields required.

airhorns: Thats correct -- the existing default selections have everything optional but I think thats really just happenstance more than anything, no strong reason they are actually that way. The *actual* selection type has to have all the keys optional so that users can pass only some keys and not have to exhaustively list `false` for everything else and I bet the default just matches that cause that was our headspace at the . Because they are out in the wild I think any new ones you produce should match methinks 👍

mkobetic: I think with my improved understanding and leaning heavily on the TestSchema example I'm inclined to settle on this:
```ts
export interface ComputedViewFunction<
  Variables,
  AvailableSelection extends FieldSelection,
  Result
> {
  (variables: Variables, options?: { select?: AvailableSelection }): Promise<
    Select<Result, DefaultSelection<AvailableSelection, Selectable<AvailableSelection>, Required<AvailableSelection>>>
  >;
    // ...
}
```
I was struggling with the two Selection types, given that we can't really know what the actual selection is going to be at runtime. So I think it's fine to say that if the select: option is present the value is of type AvailableSelection and the default is a Required<> derivative of AvailableSelection. Does that make sense?

mkobetic: Hrm, still something off, this give me `never`
```
Select<TotalInStock, DefaultSelection<AvailableTotalInStockSelection, Selectable<AvailableTotalInStockSelection>, Required<AvailableTotalInStockSelection>>>
```

mkobetic: I'm thinking the issue is with the default arg `Required<AvailableSelection>`, we don't want just required fields, we want them all set to true, so I need a type that has all fields and all typed as 'true' I think.

packages/api-client-core/spec/default-selection.spec.ts
```diff
@@ -1,11 +1,90 @@
...
 
-type _NullDefault = DefaultSelection<AvailableTestSchemaSelection, { select: null }, { num: true }>;
-type _TestDefaultsNullToTheDefault = AssertTrue<IsExact<_NullDefault, { num: true }>>;
+test("DefaultSelection", () => {
+  type nullDefault = DefaultSelection<AvailableTestSchemaSelection, { select: null }, { num: true }>;
+  type _TestDefaultsNullToTheDefault = AssertTrue<IsExact<nullDefault, { num: true }>>;
 
-type _NonNullDefault = DefaultSelection<AvailableTestSchemaSelection, { select: { num: false; str: true } }, { num: true }>;
-type _TestRespectsTruthySelections = AssertTrue<IsExact<_NonNullDefault, { num: false; str: true }>>;
+  type undefinedDefault = DefaultSelection<AvailableTestSchemaSelection, { select: undefined }, { num: true }>;
```
mkobetic, line 12: adding more cases where the default should result.

packages/api-client-core/spec/default-selection.spec.ts
```diff
@@ -1,11 +1,90 @@
...
+  type _TestDefaultsUndefinedToTheDefault = AssertTrue<IsExact<undefinedDefault, { num: true }>>;
 
-test("true", () => undefined);
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  type undefinedDefault2 = DefaultSelection<AvailableTestSchemaSelection, {}, { num: true }>;
+  type _TestDefaultsUndefinedToTheDefault2 = AssertTrue<IsExact<undefinedDefault2, { num: true }>>;
+
+  type anyDefault = DefaultSelection2<AvailableTestSchemaSelection, { select: any }, { num: true }>;
```
mkobetic, line 19: Again `any` breaks the original DefaultSelection.

packages/api-client-core/spec/default-selection.spec.ts
```diff
@@ -1,11 +1,90 @@
...
+  type nonDefault2 = DefaultSelection<AvailableTestSchemaSelection, { select: { num: false } }, { num: true }>;
+  type _TestRespectsTruthySelections2 = AssertTrue<IsExact<nonDefault2, { num: false }>>;
+});
+
+test("AvailableSelection", () => {
+  type availableTestSchemaSelection = AvailableSelection2<TestSchema>;
+  type _TestAvailableTestSchemaSelection = Has<FieldSelection, availableTestSchemaSelection>;
+
+  type testType = { a: number; b: { c: string; d: { e: boolean } } };
+  type availableSelection = AvailableSelection2<testType>;
```
mkobetic, line 34: The original AvailableSelection has a bug where boolean fields seem to end up being just `boolean | undefined` instead of `boolean | null | undefined`.

packages/api-client-core/src/types.ts
```diff
@@ -21,15 +21,66 @@ export type LimitToKnownKeys<T, U> = {
...
+ *   nested?: {
+ *     count: boolean | null | undefined
+ *   }
+ * }
+ */
+export type AvailableSelection<Schema> = Schema extends string | number | bigint | null | undefined
+  ? boolean | null | undefined
+  : { [key in keyof Schema]?: AvailableSelection<Schema[key]> };
+
+export type AvailableSelection2<Schema> = Schema extends Array<infer T>
```
mkobetic, line 62: Two issues with the original:
* doesn't handle arrays properly
* converts boolean fields into `boolean | undefined` instead of `boolean | null | undefined` for some reason

packages/api-client-core/src/types.ts
```diff
@@ -21,15 +21,66 @@ export type LimitToKnownKeys<T, U> = {
...
  * Given an options object from a find method, default the type of the selection to a default if no selection is passed
  */
+
 export type DefaultSelection<
   SelectionType,
   Options extends { select?: SelectionType | null },
   Defaults extends SelectionType
 > = Options["select"] extends SelectionType ? Options["select"] : Defaults;
 
+export type DefaultSelection2<
```
mkobetic, line 78: This one I think I do need for the original PR, tightening the types likely allows the TS inference to do the right thing for me (not 100% sure but at this point I just want to get the whole thing into a working state before tweaking things further). It also fixes it to return the Defaults when the Options ends up resolved to `any` instead of returning `any` which screws everything up.

This is also likely the one that breaks the other packages like `react` because the types are now a lot tighter than before. Again would need to dig more to confirm.

packages/api-client-core/src/types.ts
```diff
@@ -74,6 +140,24 @@ type InnerSelect<Schema, Selection extends FieldSelection | null | undefined> =
         : never;
     };
 
+type InnerSelect2<Schema, Selection extends FieldSelection | null | undefined> = IfAny<
```
mkobetic, line 143: The fix here is to return `never` when Selection ends up being `any`. I probably don't need this for the PR, but we should probably fix this regardless.

packages/api-client-core/src/types.ts
```diff
@@ -749,33 +834,6 @@ export type PaginateOptions = {
...
- *   foo?: boolean | null | undefined;
- *   bar?: boolean | null | undefined;
- *   nested?: {
- *     count: boolean | null | undefined
- *   }
- * }
- */
-export type AvailableSelection<Schema> = Schema extends string | number | bigint | null | undefined
-  ? boolean | null | undefined
-  : { [key in keyof Schema]?: AvailableSelection<Schema[key]> };
```
mkobetic, line 777: I just moved this up to be among the generic utility types.

packages/api-client-core/src/GadgetFunctions.ts
```diff
@@ -206,3 +207,63 @@ export interface GlobalActionFunction<VariablesT> {
...
 export type AnyActionFunction = ActionFunctionMetadata<any, any, any, any, any, any> | GlobalActionFunction<any>;
 export type AnyBulkActionFunction = ActionFunctionMetadata<any, any, any, any, any, true>;
+
+// This is a function that represents a computed view that doesn't take any input parameters/variables.
+// Result is an explicit type parameter defining the shape of the full result.
+export type ComputedViewFunctionWithoutVariables<Result> =
+  // Available, Options and Defaults are inferred at call time.
+  <
+    // Available is the full FieldSelection type derived from the shape of the Result type, i.e. all possible selections.
+    Available extends AvailableSelection2<Result> & FieldSelection,
```
mkobetic, line 217: Note the `& FieldSelection`, something about `AvailableSelection<>` makes it not assignable to `FieldSelection` so narrowing it down to what matches.

packages/api-client-core/src/types.ts
airhorns: Why introduce a copy instead of just changing the original? Some backwards compatibility concern?

packages/api-client-core/spec/operationBuilders.spec.ts
```diff
@@ -1491,4 +1492,52 @@ describe("operation builders", () => {
...
+          "boom",
+          { a: true, b: true },
+          { a: { required: false, type: "Int", value: 42 }, b: { required: false, type: "String", value: "fortytwo" } },
+          { a: true }
+        )
+      ).toMatchInlineSnapshot(`
+        {
+          "query": "query boom($a: Int, $b: String) {
+          boom(a: $a, b: $b) {
+            a
```
airhorns, line 1531: 👍

airhorns (approved):
This LGTM after the question about the copied types!

packages/api-client-core/src/types.ts
airhorns: I see I see. To me that suggests that either the react package needs to change and never pass in `any`, or that this change is actually a bad idea and that `any` occurs "naturally" as an input type for good reasons. Can you hunt down if we want to eliminate `any` entirely or not

mkobetic: I can definitely dig into that, I'm just not sure how much of a rabbit hole that's going to be. Does it make sense to ship this with the duplication and then follow up with the attempted merge? Otherwise I'm not sure how much more this is going to drag out.

airhorns: We can merge without the deduplication sure! I think its worth resolving before release though.

mkobetic:
I added some tests and squashed and rebased and apparently I can't build the package anymore for some reason. The error doesn't seem to have anything to do with what I've changed.

```
martin@MacBookPro api-client-core % npm run build

> @gadgetinc/api-client-core@0.15.33 build
> rm -rf dist && mkdir -p dist/cjs dist/esm && echo '{"type": "commonjs"}' > dist/cjs/package.json && echo '{"type": "module"}' > dist/esm/package.json && tsc -b tsconfig.cjs.json tsconfig.esm.json && tsx build/replace-version-constant.ts

src/GadgetConnection.ts:500:47 - error TS2554: Expected 3 arguments, but got 2.

500           this.subscriptionClientOptions?.on?.connected?.(socket, payload);
                                                  ~~~~~~~~~

  node_modules/graphql-ws/lib/client.d.ts:84:98
    84 export type EventConnectedListener = (socket: unknown, payload: ConnectionAckMessage['payload'], wasRetry: boolean) => void;
                                                                                                        ~~~~~~~~~~~~~~~~~
    An argument for 'wasRetry' was not provided.

src/GadgetConnection.ts:501:26 - error TS2554: Expected 3 arguments, but got 2.

501           overrides?.on?.connected?.(socket, payload);
                             ~~~~~~~~~

  node_modules/graphql-ws/lib/client.d.ts:84:98
    84 export type EventConnectedListener = (socket: unknown, payload: ConnectionAckMessage['payload'], wasRetry: boolean) => void;
                                                                                                        ~~~~~~~~~~~~~~~~~
    An argument for 'wasRetry' was not provided.

src/GadgetConnection.ts:500:47 - error TS2554: Expected 3 arguments, but got 2.

500           this.subscriptionClientOptions?.on?.connected?.(socket, payload);
                                                  ~~~~~~~~~

  node_modules/graphql-ws/lib/client.d.ts:84:98
    84 export type EventConnectedListener = (socket: unknown, payload: ConnectionAckMessage['payload'], wasRetry: boolean) => void;
                                                                                                        ~~~~~~~~~~~~~~~~~
    An argument for 'wasRetry' was not provided.

src/GadgetConnection.ts:501:26 - error TS2554: Expected 3 arguments, but got 2.

501           overrides?.on?.connected?.(socket, payload);
                             ~~~~~~~~~

  node_modules/graphql-ws/lib/client.d.ts:84:98
    84 export type EventConnectedListener = (socket: unknown, payload: ConnectionAckMessage['payload'], wasRetry: boolean) => void;
                                                                                                        ~~~~~~~~~~~~~~~~~
    An argument for 'wasRetry' was not provided.


Found 4 errors.

```

airhorns:
Hm, the fact that it is green in CI makes me think its something strange about your local environment! Can I suggest `rm -rf node_modules; pnpm install` to see if its some corruption in there?

mkobetic:
> Hm, the fact that it is green in CI makes me think its something strange about your local environment! Can I suggest `rm -rf node_modules; pnpm install` to see if its some corruption in there?

Yup, looks like wiping the modules helped. I also noticed I managed to mix pnpm and npm, maybe that's what screwed things up.

PR #673, opened 2024-10-15
Make delete typed bulkActions appear red in AutoTable

MillanWangGadget:
- Update
  - `actionType: "delete"` bulk actions used in AutoTable will now appear as red-text buttons to differentiate their destructive nature
  - Unfortunately, the controllability of the buttons in Polaris seems to be limited to that `destructive` bool value


alanko0511 (approved):
:shipit: 👌

PR #674, opened 2024-10-16
Fix setting belongsTo fields with AutoHiddenInput

airhorns:
This fixes using hidden inputs against belongs tos. Previously, the hidden input would set the string ID right into the form values as the value at the field, which doesn't match what `useActionForm` is expecting or what the belongs to controller already does. For belongstos, we need to suffix the path with a `.id` to have useActionForm process the form properly. Without this, the validator complains that the field value is not an object -- it's expecting it to be a `{_link: "123"}` sort of deal. We overcome this by setting the id.

This makes me think there may be other transformations we need to do within AutoHiddenInput -- are yall aware of any others?

### DIFF
--- a/packages/react/spec/auto/inputs/PolarisAutoHiddenInput.spec.tsx
+++ b/packages/react/spec/auto/inputs/PolarisAutoHiddenInput.spec.tsx
@@ -28,6 +28,7 @@ describe("PolarisAutoHiddenInput", () => {
       <PolarisAutoForm action={api.widget.create}>
         <PolarisAutoHiddenInput field="name" value="Bob" />
         <PolarisAutoHiddenInput field="inventoryCount" value={42} />
+        <PolarisAutoHiddenInput field="section" value="123" />
         <PolarisAutoSubmit />
       </PolarisAutoForm>,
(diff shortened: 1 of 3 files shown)


MillanWangGadget (commented):
- In a real app,
  - 🍏 confirmed that `belongsTo` fields get submitted and set properly when  used like this `<AutoHiddenInput field="belongsToFieldApiId" value="123" />`
  - 🍏 Confirmed that `hasMany` fields get submitted and set properly when used like this `<AutoHiddenInput field="hasManyFieldApiId" value={[{id:"1", f:"toUpdate"}, {f:"toCreate"}]} />`
  - 🟡 RichText fields in `AutoHiddenInput` need to have `value={{markdown:"abc"}}` to work. I think that's fine though
  - 🟡 File fields should probably throw an error when used in here. We end up submitting a token in the request which is only valid for 3hours

alanko0511 (approved):
:shipit: 👌

MillanWangGadget (approved):
- 🍏 JSON fields in hidden inputs work with string values and objects 

Looks good 👍

PR #675, opened 2024-10-16
Fix incorrect types for background options priority

thegedge:
As reported in Discord several times now, people write things to fit the types, but then when trying something out will get an error from the GraphQL API:

> [GraphQL] Variable "$backgroundOptions" got invalid value "high" at "backgroundOptions.priority"; Value "high" does not exist in "BackgroundActionPriority" enum. Did you mean the enum value "HIGH"?"

We have two options here:

1. Keep the type the same, and call `toUpperCase` to ensure we send the correct value.
2. Change the type to match the GraphQL value.

(1) is nice, because it avoids any breaking changes in the types, but I got some feedback in Discord that people would be down for (2). Specifically, the following was said:

> Easy to change and is made to cope with an initial error, so it makes sense to remove it

The thing I like about (2) is that our types mirror the GraphQL API vs diverging and that we don't have to do any data munging. The "breaking" change would only be at a type level, where people are currently doing stuff like this:

```
priority: "HIGH" as unknown as "high"
```

If we went with (1), existing apps would continue to have ☝️ or have to change everything to be lowercase. It seems like some of the Discord users would like to and be okay with driving removal of those type casts through type errors.

## PR Checklist

### DIFF
--- a/packages/api-client-core/spec/operationRunners.spec.ts
+++ b/packages/api-client-core/spec/operationRunners.spec.ts
@@ -1237,13 +1237,13 @@ describe("operationRunners", () => {
         MockWidgetCreateAction,
         { widget: { name: "new widget" } },
         {
-          priority: "high",
+          priority: "HIGH",
         }
       );
 
       expect(mockUrqlClient.executeMutation.mock.calls.length).toEqual(1);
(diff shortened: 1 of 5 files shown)


airhorns (approved):
I'm cool with this!

PR #681, opened 2024-10-18
Auto input hoc

infiton:
We want to start moving to an approach in auto components where we make a single selection on models based on the fields that are going to be active in a form. In order to do that we need to know which fields will be needed for inputs; the way that I am proposing to do this is to make all auto inputs higher order components that are wrapped in `autoInput`; doing this will allow us to look at the component tree of an auto form and determine the selection that needs to be made on the model that is being mutated

## PR Checklist

### DIFF
--- a/packages/react/src/auto/AutoInput.tsx
+++ b/packages/react/src/auto/AutoInput.tsx
@@ -1,0 +1,19 @@
+import React from "react";
+
+export interface AutoInputComponent<P> extends React.FC<P> {
+  __autoInput: true;
+}
+
+export function autoInput<P extends { field: string }>(Component: React.FC<P>): AutoInputComponent<P> {
+  const WrappedComponent: React.FC<P> = (props) => {
+    return <Component {...props} />;
+  };
+
+  (WrappedComponent as AutoInputComponent<P>).__autoInput = true;
+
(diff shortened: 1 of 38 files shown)


packages/react/src/auto/AutoInput.tsx
```diff
@@ -0,0 +1,19 @@
...
+  };
+
+  (WrappedComponent as AutoInputComponent<P>).__autoInput = true;
+
+  return WrappedComponent as AutoInputComponent<P>;
+}
+
+export function isAutoInput(component: React.ReactElement): component is React.ReactElement<any, AutoInputComponent<any>> {
+  return typeof component.type === "function" && "__autoInput" in component.type;
+}
```
MillanWangGadget, line 19: How and where would this function be called?

MillanWangGadget (commented):
- 🍏 All inputs have the new `autoInput()` wrapper

packages/react/src/auto/AutoInput.tsx
infiton: its coming in an upcoming PR

MillanWangGadget (approved):
- 🍏 Confirmed that the exported input components still work the same in a real app

PR #684, opened 2024-10-21
Fixed bug with loading selected related model records in AutoForm HasMany inputs in framework version 1.3

MillanWangGadget:
- Update
  - See title
  - This was a consequence of our test environments not being configured with framework version 1.3 and the corresponding changes to related model filtering

### DIFF
--- a/packages/react/.changeset/lazy-pans-push.md
+++ b/packages/react/.changeset/lazy-pans-push.md
@@ -1,0 +1,5 @@
+---
+"@gadgetinc/react": patch
+---
+
+Fixed bug with loading selected related model records in AutoForm HasMany inputs in framework version 1.3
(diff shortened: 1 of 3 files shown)


alanko0511 (approved):

danroberts (approved):
Looks good. As a followup to the fix, could you also add a test that confirms that it is working with both framework version 1.2 + 1.3? I think you could create a test app that stays at v1.2

MillanWangGadget:
> Looks good. As a followup to the fix, could you also add a test that confirms that it is working with both framework version 1.2 + 1.3? I think you could create a test app that stays at v1.2

Yes I will follow up with a 1.2 vs 1.3 test 👍

PR #691, opened 2024-11-06
Updated AutoButton type to accept non-string children

MillanWangGadget:
- Previously, the type was restricted to `string | string[] | undefined` but react components would work
  - This was an unnecessary restriction imposed by Polaris 
- Updated the type to allow for react components


MillanWangGadget:
Closing because sticking with Polaris types feels safer and users can add `ts-ignore` themselves in order to put children into the button

PR #693, opened 2024-11-12
Merge diverged utility types (GGT-7411)

mkobetic:
## PR Checklist


mkobetic:
We're going to revert the changes from api-client-core so this is obsolete now.

PR #695, opened 2024-11-19
Fix record changed() logic for comparing current value Dates

jcao49:

## PR Checklist


packages/api-client-core/spec/GadgetRecord.spec.ts
```diff
@@ -442,4 +442,165 @@ describe("GadgetRecord", () => {
...
+      };
+
+      const product = new GadgetRecord<SampleBaseRecord>(productBaseRecord);
+      expect(product.changed()).toBe(false);
+
+      product.publishedAt = new Date(dateStr);
+      expect(product.changed()).toBe(false);
+    });
+
+    it("correctly calculates record changed when the current value is a Date and is the same date as the previous value", () => {
```
jasong689, line 464: maybe add that this test and the previous are different since one uses `changed` and the other is `changed("publishedAt")`. i was a bit confused since they looked almost identical

packages/api-client-core/spec/GadgetRecord.spec.ts
```diff
@@ -442,4 +442,165 @@ describe("GadgetRecord", () => {
...
+        body: "A description of why it's cool",
+        publishedAt: new Date(newDateStr),
+      };
+
+      const product = new GadgetRecord<SampleBaseRecord>(productBaseRecord);
+      expect(product.changed("publishedAt")).toBe(false);
+
+      product.publishedAt = oldDateStr;
+      expect(product.changed("publishedAt")).toBe(true);
+    });
```
jasong689, line 604: Could we add test case for the following:
- invalid date
- 2 `Date`
- date without time

jasong689 (commented):
changes look good. will 🎩

pistachiobaby (approved):
🎩 ✅ 

Published Product Update webhook no longer incorrectly returns true for `record.changed("publishedAt") ✅ 

Unpublished -> Published product does return true for both scenarios ✅ 
CleanShot 2024-11-25 at 09 19 10@2x

jasong689 (approved):

PR #696, opened 2024-11-20
Revert "computed view endpoint for api client (GGT-7201)"

mkobetic:
Reverts gadget-inc/js-clients#672.

We've pulled all these changes over to the client generator side, so these are now unnecessary here.


airhorns (approved):
LGTM!

PR #697, opened 2024-11-20
Propagate graphql api errors when live

garymardell:
Currently if you use the api in live mode and there is an error (such as a permission error) you get back:

```
"[Network] Missing data."
```

this is because the response you get back is:

```
{
  "id": "b5202ab5-d0fb-4a99-baef-be784ae67296",
  "type": "next",
  "payload": {
    "data": null,
    "errors": [
      {
        "message": "GGT_PERMISSION_DENIED: Permission has been denied...",
        "locations": [{ "line": 2, "column": 3 }],
        "path": ["posts"]
      }
    ],
    "revision": 1
  }
}
```

where the `data` value is `null`, which is expected, however the api-core patch handler throws an error when it is missing rather than publishing the response with the error. 

## PR Checklist


packages/react/package.json
```diff
@@ -48,7 +48,7 @@
   },
   "dependencies": {
     "@0no-co/graphql.web": "^1.0.4",
-    "@gadgetinc/api-client-core": "^0.15.34",
+    "@gadgetinc/api-client-core": "^0.15.35",
```
MillanWangGadget, line 51: These also have references to 0.15.34

packages/shopify-extensions/package.json
packages/react-shopify-app-bridge/package.json
packages/react-bigcommerce/package.json

garymardell:
Will update this once https://github.com/gadget-inc/js-clients/pull/698 has been merged and the new version of api-client-core is released.

airhorns (approved):

PR #701, opened 2024-11-28
Added an error in AutoTable for adding `sortable:true` to fields that are not sortable

MillanWangGadget:
- UPDATE
  - Previously, adding `sortable:true` in a column configuration for an unsortable field would properly render the page, but an error would be thrown if you tried to sort the column
  - Now, the error is thrown upon rendering the component instead of upon sorting so that it is easier to detect this problem 
  - Relationship columns are always marked as non-sortable by the API even if the displayed field on the related model is sortable. 
    - Adding that functionality to the column sorting system would depend on the Gadget app's framework version and shall be considered a separate concern 
  - In storybook, `sortable:false` has been added to some of the fields so that it is easier to test this new error


alanko0511 (approved):
:shipit: 👌

PR #702, opened 2024-12-02
Added `searchValue` property to `AutoTable` component to control the table's search value

MillanWangGadget:
UPDATE
- Added `searchValue` prop to `AutoTable` to control the search value of the table from outside of the component itself
- This prop is compatible with `searchable:true` in AutoTable. In this case, the new `searchValue` prop will only be used to power the search when the user does not manually add their own search 

CleanShot 2024-12-02 at 14 28 20


alanko0511 (approved):
🎩 
:shipit: 👌

PR #703, opened 2024-12-02
Update js-clients to process action responses

garymardell:
The response from an upsert action was not being properly handled by `useAction` and therefore the data was not being returned to the user. This was because the `action.hasReturnType` looks like:

```
{
  "... on CreatePostResult": { hasReturnType: false },
  "... on UpdatePostResult": { hasReturnType: true },
}
```

in the useAction processResult it would just check if `action.hasReturnType` is truthy and if so use `mutationData.result`. However in this case the value is truthy but an object not a boolean value. 

We already have the utility inside the `api-client-core` to process the action results (which is used both by the core action runner and background actions) which handles this case correctly so this updates the code to use that instead.

## PR Checklist

### DIFF
--- a/packages/react-bigcommerce/package.json
+++ b/packages/react-bigcommerce/package.json
@@ -27,7 +27,7 @@
     "prerelease": "gitpkg publish"
   },
   "dependencies": {
-    "@gadgetinc/api-client-core": "^0.15.35"
+    "@gadgetinc/api-client-core": "^0.15.36"
   },
   "devDependencies": {
     "@gadgetinc/api-client-core": "workspace:*",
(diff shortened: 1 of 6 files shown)


garymardell:
The vite tests fail on this PR too as its trying to use the version of api-client-core that would be released by this PR...

airhorns:
Hm, thats new! If you don't bump the version in this PR, does it work ok?

airhorns (approved):
Changes look good to me

garymardell:
>  If you don't bump the version in this PR, does it work ok?

No, because the package doesn't export the function for react client to use without version bump. Going to merge a PR that exposes the function and bumps version first and then will rebase this.

PR #708, opened 2024-12-06
Removed `allowMultiple` from `AutoEnumInput` props type

MillanWangGadget:
- Updated `AutoEnumInput` props type to no longer include `allowMultiple` as it could not override the configuration of the corresponding Gadget field
- This prop was included by the Polaris Combobox props spread, and gave users the wrong idea about what could be configured in the input


alanko0511 (approved):
:shipit: 👌

PR #709, opened 2024-12-06
Fix double delete on use action form

infiton:
we had a nasty issue with how we determined which ids have been updated/created/deleted when there are grandchildren. The ids for each array element were being addressed by the array index which the form can change (by reordering the children or deleting and appending)

This PR fixes that by addressing the existing ids not by array index but by the id of the array element

## PR Checklist

### DIFF
--- a/packages/react/src/use-action-form/utils.ts
+++ b/packages/react/src/use-action-form/utils.ts
@@ -157,13 +157,18 @@ export const disambiguateDefaultValues = (data: any, initialData: any, action: a
   return result;
 };
 
-function getUpdates(data: any): Record<string, number[]> {
+function getRecordIdsAtPath(data: any): Record<string, number[]> {
   const updateList: Record<string, number[]> = {};
 
   function traverse(input: any, path: string | undefined = undefined, depth = 0, isInArray?: boolean): any {
     if (Array.isArray(input)) {
-      return input.map((item: any, index) => {
-        const currentPath = path ? `${path}.${index}` : index.toString();
+      return input.flatMap((item: any) => {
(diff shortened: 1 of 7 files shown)


packages/react/src/use-action-form/utils.ts
```diff
@@ -365,7 +388,7 @@ export const reshapeDataForGraphqlApi = async (client: AnyClient, defaultValues:
 
       const inputHasId = "id" in input;
       const inputHasMoreFields = Object.keys(input).filter((field) => field !== "__typename").length > 1;
-      const inputUpdateId = updates[path!]?.[0];
+      const inputUpdateId = idPath && recordIdsAtPath[idPath] ? recordIdsAtPath[idPath][0] : undefined;
```
MillanWangGadget, line 391: Consider adding a check for `recordIdsAtPath[idPath].length` in the off chance that `recordIdsAtPath[idPath]` is an empty array

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.18.4",
+  "version": "0.18.5",
```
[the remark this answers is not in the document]
infiton: I'll remove the version bump and then we can merge that PR on top

packages/react/src/use-action-form/utils.ts
infiton: :+1:

infiton:
going to merge; but this won't be released until https://github.com/gadget-inc/js-clients/pull/704 merges

PR #713, opened 2025-01-03
Updated AutoRichTextInput to eager-load dependencies instead of in a useEffect

MillanWangGadget:
- **UPDATE**
  - `AutoRichTextInput` imported the `mdxEditor` dependency through an asynchronous function in a useEffect, and the components was stuck in a loading state until the import was resolved
    - This "lazy loaded" component was then used in `LazyLoadedPolarisAutoRichTextInput.tsx`. We were effectively doing a double lazy load when we only needed to do it once
    - `LazyLoadedPolarisAutoRichTextInput.tsx` was a later addition after the original `AutoRichTextInput` component was added. When I added that, I think I neglected to remove the useEffect based import, which was originally used to prevent errors when using AutoComponents without `mdxEditor`
  - Now the `mdxEditor` import is directly eager loaded into the `AutoRichTextInput` component. 
    - We will now rely only on the frontend library specific lazy loaded implementation of the `AutoRichTextInput` to make `mdxEditor` optional 

BEFORE - The form loads in and RichText inputs are stuck in a loading state for a short while
CleanShot 2025-01-03 at 13 16 10

NOW - Rich text editor loads immediately with the rest of the form
CleanShot 2025-01-03 at 13 57 06

## 🎩 Testing Instructions

1. Use this prerelease in a new app that DOES NOT have the mdxEditor package installed 
2. `"@gadgetinc/react": "https://codeload.github.com/gadget-inc/js-clients/tar.gz/@gadgetinc/react-v0.18.5-gitpkg-45ef0664",`
3. Add an `AutoForm` on an action that has some fields. Observe everything is good when there are no rich text fields
4. Add a rich text field, and observe the big error for the missing mdxEditor package
5. Add `"@mdxeditor/editor": "^3.8.0",` to the app's `package.json`
6. Observe that using `AutoForm` no longer has errors for the missing import 
7. Observe that reloading the page does not show the `loading` state like it did previously

### DIFF
--- a/packages/react/spec/auto/PolarisAutoForm.spec.tsx
+++ b/packages/react/spec/auto/PolarisAutoForm.spec.tsx
@@ -301,7 +301,7 @@ describe("PolarisAutoForm", () => {
         category: [],
         color: null,
         description: {
-          markdown: "example _rich_ **text**",
+          markdown: "example *rich* **text**",
         },
         embedding: null,
         inventoryCount: 1234,
--- a/packages/react/src/auto/shared/AutoRichTextInput.tsx
+++ b/packages/react/src/auto/shared/AutoRichTextInput.tsx
@@ -1,4 +1,6 @@
-import React, { useEffect, useRef, useState } from "react";
+import * as mdxModule from "@mdxeditor/editor";
+import "@mdxeditor/editor/style.css";
+import React, { useEffect, useRef } from "react";
 import { useFormContext } from "../../useActionForm.js";
 import { get } from "../../utils.js";
 import { autoInput } from "../AutoInput.js";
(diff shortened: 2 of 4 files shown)


packages/react/spec/auto/PolarisAutoForm.spec.tsx
```diff
@@ -301,7 +301,7 @@ describe("PolarisAutoForm", () => {
         category: [],
         color: null,
         description: {
-          markdown: "example _rich_ **text**",
+          markdown: "example *rich* **text**",
```
MillanWangGadget, line 304: Surrounding text with `*` or `_` are equivalent in markdown for indicating italics. 

MDX Editor automatically replaces `_`->`*` for all italics text. I could not find a way to disable this functionality, so I simply did the replacement to avoid test failures

The tests were previously passing because the `useEffect` would not actually load `mdxEditor` in time for it to do this replacement on the text, thus leaving it unaffected in the form state. We were accidentally passing the test 🙃

packages/react/src/auto/shared/AutoRichTextInput.tsx
```diff
@@ -1,64 +1,46 @@
...
+  thematicBreakPlugin,
+  markdownShortcutPlugin,
+  linkDialogPlugin,
+  diffSourcePlugin,
+  toolbarPlugin,
+  DiffSourceToggleWrapper,
+  UndoRedo,
+  BlockTypeSelect,
+  Separator,
+} = mdxModule;
```
MillanWangGadget, line 29: I did `import * as mdxModule from "@mdxeditor/editor";` and then dereferenced everything here because the `packages/react/spec/imports/imports-vite-project.spec.ts` test fails when importing the dereferenced stuff directly due to erroneous `no exported property` errors. 

Through manual testing, I've confirmed that AutoComponents works with||without the `mdxEditor` dependency

infiton (approved):

alanko0511 (approved):
:shipit: 👌

PR #717, opened 2025-01-07
Fix unable to delete values in hasMany field in useActionForm

alanko0511:
Previously, if the edge ID value had more than one digit, it triggered a bug that prevented the edge item from being deleted. 

When we generate a map that maps the default values and their related paths, we set the new path by using the string of the current path - 2 characters. Therefore, if the ID has more than one digit, the new path becomes `foo.bar.`, which causes the transform to fail to determine whether it should update or delete its value.

This PR fixes the issue by doing the `string.split(".")` and skipping the last value in the array to remove the ID.

## PR Checklist

### DIFF
--- a/packages/react/spec/useActionFormNested.spec.tsx
+++ b/packages/react/spec/useActionFormNested.spec.tsx
@@ -2177,6 +2177,62 @@ describe("useActionFormNested", () => {
                       },
                     },
                   },
+                  {
+                    node: {
+                      id: "11",
+                      name: "John Doe",
+                      followerFriendships: {
+                        edges: [
+                          {
+                            node: {
+                              id: "11",
+                              follower: {
+                                id: "1100",
+                                name: "Jane Doe",
+                              },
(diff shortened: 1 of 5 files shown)


packages/react/spec/useActionFormNested.spec.tsx
```diff
@@ -2177,6 +2177,62 @@ describe("useActionFormNested", () => {
                       },
                     },
                   },
+                  {
+                    node: {
+                      id: "11",
```
alanko0511, line 2182: This updated test fails when I bring back the old `substring` approach because I introduce edge items with more than one digit.

MillanWangGadget (approved):
Change makes sense. I've confirmed that there are no other potentially problematic uses of `substring` in the repo

MillanWangGadget:
I think that this should be released as a new package immediately to help the original bug reporter.

infiton (approved):

PR #720, opened 2025-01-11
Don't reexecute live queries when their cache is invalidated

scott-rc:
This prevents the cache exchange from re-executing live queries when their cache is invalidated.

This change is to prevent the following scenario:
1. component uses `findMany` with `live: true`
2. component executes a mutation that touches the `__typename` the `findMany` references
3. cache exchange re-executes the `findMany` because the mutation invalidated its cache

Step 3 doesn't stop the existing live query either, causing live queries to leak on the client and cause the server to keep track and push changes to live queries that aren't being used.

---

I tried to implement this by "wrapping" the cache exchange using the answer [in this discussion](https://github.com/urql-graphql/urql/discussions/3559) but the suggested code doesn't work and the original author of the discussion [says that they had to fork the exchange](https://github.com/urql-graphql/urql/discussions/3559#discussioncomment-9338149) in order to skip the cache.

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/exchanges/cacheExchange.ts
+++ b/packages/api-client-core/src/exchanges/cacheExchange.ts
@@ -1,0 +1,237 @@
+/**
+ * This file was copied from: https://github.com/urql-graphql/urql/blob/25d114d25807f0676dbf453732602753279ba0db/packages/core/src/exchanges/cache.ts
+ * Any meaningful changes are documented with comments starting with "GADGET:"
+ */
+/* eslint-disable @typescript-eslint/no-use-before-define */
+import {
+  formatDocument,
+  makeOperation,
+  makeResult,
+  type Client,
+  type Exchange,
+  type Operation,
+  type OperationContext,
+  type OperationResult,
+} from "@urql/core";
+import { filter, map, merge, pipe, tap } from "wonka";
+import { isLiveQueryOperationDefinitionNode } from "../graphql-live-query-utils/index.js";
+
+type ResultCache = Map<number, OperationResult>;
+type OperationCache = Map<string, Set<number>>;
+
+const shouldSkip = ({ kind }: Operation) => kind !== "mutation" && kind !== "query";
+
+/** Adds unique typenames to query (for invalidating cache entries) */
+export const mapTypeNames = (operation: Operation): Operation => {
+  const query = formatDocument(operation.query);
+  if (query !== operation.query) {
+    const formattedOperation = makeOperation(operation.kind, operation);
+    formattedOperation.query = query;
+    return formattedOperation;
+  } else {
+    return operation;
+  }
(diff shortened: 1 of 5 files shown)


packages/api-client-core/src/exchanges/cacheExchange.ts
```diff
@@ -0,0 +1,237 @@
...
+          }
+
+          for (const key of pendingOperations.values()) {
+            if (resultCache.has(key)) {
+              operation = (resultCache.get(key) as OperationResult).operation;
+              // GADGET: added the below line to skip reexecuting live queries when their data is invalidated since they should receive updates from the server
+              if (!operation.query.definitions.some(isLiveQueryOperationDefinitionNode)) {
+                resultCache.delete(key);
+                reexecuteOperation(client, operation);
+              }
```
scott-rc, line 162: This `if` statement is the only meaningful change of this entire PR.

packages/api-client-core/src/exchanges/cacheExchange.ts
```diff
@@ -0,0 +1,237 @@
+/**
+ * This file was copied from: https://github.com/urql-graphql/urql/blob/25d114d25807f0676dbf453732602753279ba0db/packages/core/src/exchanges/cache.ts
+ * Any meaningful changes are documented with comments starting with "GADGET:"
+ */
```
thegedge, line 4: We can keep this comment, but should do one of two things to adhere to the MIT license: either copy https://github.com/urql-graphql/urql/blob/main/LICENSE into this directory (maybe we put it in a subdirectory so the license is properly scoped to just the files that matter) or we copy the LICENSE contents into the top of all files we copied.

thegedge (approved):
😭 🚀

packages/api-client-core/src/exchanges/cacheExchange.ts
```diff
@@ -0,0 +1,237 @@
...
+            for (const key of operations.values()) pendingOperations.add(key);
+            operations.clear();
+          }
+
+          for (const key of pendingOperations.values()) {
+            if (resultCache.has(key)) {
+              operation = (resultCache.get(key) as OperationResult).operation;
+              // GADGET: added the below line to skip reexecuting live queries when their data is invalidated since they should receive updates from the server
+              if (!operation.query.definitions.some(isLiveQueryOperationDefinitionNode)) {
+                resultCache.delete(key);
```
thegedge, line 160: Should we still delete, or is there potentially a race condition? Just wondering if there's any opportunity for stale cache.

airhorns (approved):
🚢

scott-rc:
We never want to re-execute live queries so I'm gonna go with this instead: https://github.com/gadget-inc/js-clients/pull/723

PR #721, opened 2025-01-12
shadcn skeleton

airhorns:
WIP implementation of AutoButton for shadcn, showing the factory pattern and example test layout


packages/react/src/auto/shadcn/inputs/ShadcnAutoTextInput.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import type { InputHTMLAttributes } from "react";
+import type { Control } from "../../../useActionForm.js";
+import { getPropsWithoutRef } from "../../../utils.js";
+import { autoInput } from "../../AutoInput.js";
+import { useStringInputController } from "../../hooks/useStringInputController.js";
+import type { ShadcnElements } from "../elements.js";
+
+// should we move this to a shared location?
+import { cn } from "../../../../spec/auto/shadcn-defaults/utils.js";
```
airhorns, line 10: I think the move would be to expect this to be passed in as an element that you can then get from the context, as users are likely gonna have their own version that they use in their shadcn components, and we wnat to use the same one. generally speaking, any import from the `spec` dir in the `src` dir is busted

jadewale: @airhorns This import was to merge the styles that came with the users component and add an error state if the input had an error.

packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { makeSubmitResultBanner } from "./submit/ShadcnSubmitResultBanner.js";
+
+/**
+ * Renders a form for an action on a model automatically using Shadcn
+ */
+export const makeAutoForm =
+  <Elements extends ShadcnElements>({ Form, Input, Button, Alert, Skeleton, AlertTitle, AlertDescription, Label, Checkbox }: Elements) =>
+  <GivenOptions extends OptionsType, SchemaT, ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any>>(
+    props: AutoFormProps<GivenOptions, SchemaT, ActionFunc> & ComponentProps<typeof Form>
+  ) => {
```
airhorns, line 21: just realizing that we should probably use named functions instead of anonymous functions in these factories here so that stack traces and the React dev tools can use the right name for the components in development

packages/react/src/auto/shadcn/inputs/ShadcnAutoTextInput.tsx
airhorns: Right, I understant we need the utility, but I am saying we can't import it ourselves from the version in `spec`. I think this utility needs to be provided to us as one of the elements, and then referred to from the elements at runtime. The user will need their own version of this utility in their app, and will use it in all their shadcn components, we should re-use theirs, not import from our own. In general, any code in `src` importing from `spec` is an antipattern as well, if we intend to ship the code at runtime it needs to be in `src`. But, that's why I put the default components in `spec` -- we don't intend to ship any of that at runtime, its just a copy of shadcn we use for testing.

packages/react/cypress/component/auto/form/AutoForm.cy.tsx
```diff
@@ -129,6 +129,14 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
   });
 
   it("can render a form to update model and submit it", () => {
+
+    /**
+     * This test is disabled for Shadcn because it's not supported yet. We need to have a list component for this to work properly
+     */
+    if (name === "Shadcn") {
+      return;
+    }
```
airhorns, line 138: tiny nit: I think this is going to be easy to forget to fix later cause its going to show up as a green test when we look at the list of tests. can you file a github issue to ensure we come back to re-enable this?

packages/react/spec/auto/shadcn-defaults/components/Alert.tsx
```diff
@@ -0,0 +1,36 @@
...
+  <h5 ref={ref} className={cn("mb-1 font-medium leading-none tracking-tight", className)} {...props} />
+));
+AlertTitle.displayName = "AlertTitle";
+
+const AlertDescription = React.forwardRef<HTMLParagraphElement, React.HTMLAttributes<HTMLParagraphElement>>(
+  ({ className, ...props }, ref) => <div ref={ref} className={cn("text-sm [&_p]:leading-relaxed", className)} {...props} />
+);
+AlertDescription.displayName = "AlertDescription";
+
+export { Alert, AlertDescription, AlertTitle };
```
airhorns, line 36: I highly recommend configuring your editor to use the eslint / prettier config that is set up in the repo already so that all the code you write matches the formatting rules we have set already and you dont have to waste time reformatting stuff manually 👍 see the root level package scripts for the commands to run to fix lint errors automatically too, like `pnpm lint:fix`

packages/react/cypress/component/auto/form/AutoFormInputLabels.cy.tsx
```diff
@@ -1,15 +1,26 @@
...
+import { Input } from "../../../../spec/auto/shadcn-defaults/components/Input.js";
+import { Label } from "../../../../spec/auto/shadcn-defaults/components/Label.js";
+
 const AutoInput = (props: { suiteName: string; field: string; label?: string }) => {
   if (props.suiteName === "Polaris") {
     return <PolarisAutoInput {...props} />;
   }
 
+  if (props.suiteName === "Shadcn") {
+    const ShadcnAutoInput = makeShadcnAutoInput({ Input: Input, Label: Label, Checkbox: Checkbox, Button: Button });
```
airhorns, line 20: nit: can you move this outside of this function? Otherwise, it's going to re-create the input component every render and for every test, which is kinda slow, and also not how production uses of this stuff will work so I think it will test it a bit differently. the factory functions should generally be invoked in the module scope

packages/react/src/auto/shadcn/inputs/ShadcnAutoNumberInput.tsx
```diff
@@ -0,0 +1,37 @@
...
+        ? getStepFromNumberOfDecimals(metadata.configuration.decimals)
+        : value
+        ? getStepFromNumberOfDecimals(countNumberOfDecimals(`${value}`))
+        : 1;
+
+    return <ShadcnAutoTextInput {...props} step={step} type={"number"} />;
+  });
+};
+
+// TODO: move to a shared location
```
airhorns, line 26: 👍 definitely -- as well as the stuff above with the `step` if that is shared with polaris

packages/react/src/auto/shadcn/inputs/ShadcnAutoPasswordInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const EncryptedInput = makeShadcnAutoEncryptedStringInput({ Input, Label, Button });
+
+  return autoInput((props: { field: string; control?: Control<any>; className?: string }) => {
+    const { findBy } = useAutoFormMetadata();
+    const { path } = useFieldMetadata(props.field);
+    const { field: fieldProps } = useController({ name: path });
+
+    const [isEditing, setIsEditing] = useState(!findBy);
+
+    const startEditing = () => {
```
airhorns, line 27: should probably be a `useCallback`

packages/react/spec/auto/shadcn-defaults/components/Alert.tsx
jadewale: Noted. Thanks

packages/react/package.json
```diff
@@ -44,6 +49,7 @@
   "dependencies": {
     "@0no-co/graphql.web": "^1.0.4",
     "@gadgetinc/api-client-core": "^0.15.38",
+    "@gadgetinc/react": "workspace:^",
```
airhorns, line 52: This is gonna break stuff -- the package can't depend on itself, and can't depend on a workspace version of the package as that won't be available on npm. imports within the package should be relative to avoid referring to it by name

PR #723, opened 2025-01-13
Don't reexecute live queries

scott-rc:
This prevents our urql client from re-executing live queries.

This change is to prevent the following scenario:
1. component uses `useFindMany` with `live: true`
2. component executes a mutation that touches the `__typename` the `useFindMany` references
3. cache exchange re-executes the `useFindMany` because the mutation invalidated its cache

Step 3 doesn't stop the existing live query, causing live queries to leak on the client and cause the server to keep track and push changes to live queries that aren't being used.

My [first attempt](https://github.com/gadget-inc/js-clients/pull/720) at fixing this issue was to prevent the cache exchange from re-executing live queries, however we realized this issue could happen again if something else decided to start re-executing live queries, so we decided to prevent live queries from ever being re-executed.

---

I've 🎩'd this change locally so I know it works, but I still need to figure out how to write a test for this.

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -441,6 +441,18 @@ export class GadgetConnection {
       requestPolicy: this.requestPolicy,
     });
     (client as any)[$gadgetConnection] = this;
+
+    const reexecuteOperation = client.reexecuteOperation.bind(client);
+    client.reexecuteOperation = (operation) => {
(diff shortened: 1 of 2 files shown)


airhorns (approved):
LGTM after some sort of test. I think you could probs mount a live query in react land and spy on something to see how many times it gets called?

thegedge (approved):
Makes sense to me (and verified `cacheExchange` calls `client.reexecuteOperation`).

PR #668, opened 2024-10-10
auto relationship forms

infiton:
... a description that explains what, why, and how ...

## PR Checklist

- [ ] Important or complicated code is tested
- [ ] Any user facing changes are documented in the Gadget-side changelog
- [ ] Any immediate changes are slated for release in Gadget via a generated package dependency bump
- [ ] Versions within this monorepo are matching and there's a valid upgrade path

PR #726, opened 2025-01-15
Single lookup for selected related records in autocomponents

MillanWangGadget:
A small subset of https://github.com/gadget-inc/js-clients/pull/668 without changes to the experience of using these components 

- **UPDATE**
  - Storybook stories updated so that they work again 
  - Existing record `findBy` selection system
    - Now, relationship fields have their values included in the initial `findBy` lookup for existing records. This eliminates the need to do additional lookups to get the related model records in relationship inputs
    - packages/react/src/use-table/helpers.tsx
  - `pause` param added to `useActionFrom`
  - Fixes from the other PR
    - Global actions fixed in auto form
    - exclude password fields from selection across the board
    - Submission result banner will now stay around until manually closed
    - BelongsTo inputs 
      - Selected record value is now shown even if the corresponding record is not loaded on the first page of related records
      - The raw ID value in the record is now retrieved in the initial selection
      - Dangling relationship records are now shown in red as an ID that cannot be resolved
- **Temporarily disabled functionality**
  - HasOne / HasManyThrough field inputs
  - HasMany fields -> Only showing related model records that are unlinked or already related to the current record
    - This would have to be an option to avoid becoming a breaking change 
    - Will need to integrate with future related model record filtering 
  - Determining the fields to include in the form state based on potentially deeply nested child components
    - The WIP approach to do this with `React.Children` does not work through components and would be a breaking change.
    - A different approach must be used to allow for dynamic deep nested inputs

It is expected that there will be no user facing changes experienced from these code changes


packages/react/src/use-action-form/utils.ts
```diff
@@ -429,23 +419,87 @@ export const reshapeDataForGraphqlApi = async (client: AnyClient, defaultValues:
...
+    if (inputHasMoreNonIdFields) {
+      return { update: { id: input["id"], ...rest } };
+    }
+
+    return { _link: input["id"] };
+  }
+
+  const inputUpdateId = idPath && recordIdsAtPath[idPath] && recordIdsAtPath[idPath].length > 0 ? recordIdsAtPath[idPath][0] : undefined;
+  if (inputUpdateId && !inputHasMoreNonIdFields) {
+    return { delete: { id: inputUpdateId } };
```
MillanWangGadget, line 458: This is a potentially controversial area that differs from the other PR. 

The other PR has this case as `{ _link: null }` but that will fail a test in packages/react/spec/useActionFormNested.spec.tsx

I kept this the same as main to make sure that this PR does not change any functionality

packages/react/src/auto/AutoForm.ts
```diff
@@ -69,6 +74,29 @@ const useValidationResolver = (metadata: ModelWithOneActionMetadata | GlobalActi
   }, [metadata, pathsToValidate]);
 };
 
+const isMetadataForUpsertAction = (metadata: ModelWithOneActionMetadata | GlobalActionMetadata | undefined | null) => {
```
infiton, line 77: I think we have `isUpsertMetaAction` on the meta schema now; should we use that instead of attempting to infer it?

infiton: I'm cool to do this in a follow up though

packages/react/src/use-action-form/utils.ts
infiton: I think the test/current behaviour is wrong; I think we'll release this as `0.19` so I'm o making that breaking change

lets talk about the test thats failing tomorrow

infiton (approved):
:pray: for getting all those tests working! Let's chat about the the two comments tomorrow, but I'm cool if you want to ship this and address the comments in a follow up

PR #729, opened 2025-01-20
Smart field selection from nested AutoInputs in custom components in AutoForm

MillanWangGadget:
- UPDATES
  - Added `findResult` to the return value of `useActionForm` for direct access to the existing record found when using the `findBy` prop
  - Added `select` prop to `useAutoForm` and `AutoForm` to allow for conditional selection of fields on the existing record found when using the `findBy` prop
  - Added `AutoFormFieldsFromChildComponents` context to contain registered fields from custom child components
    - These registered fields are used to control the form state for existing record value selection and submission
  - Behaviour should be unchanged when there are no child components in AutoForm


packages/react/src/auto/polaris/PolarisAutoForm.tsx
```diff
@@ -131,9 +133,12 @@ const PolarisAutoFormComponent = <
...
       <FormProvider {...originalFormMethods}>
-        <Form {...rest} onSubmit={submit}>
-          <BlockStack gap="400">{formContent}</BlockStack>
-        </Form>
+        {isLoading && <PolarisAutoFormSkeleton />}
+        <div hidden={isLoading}>
+          <Form {...rest} onSubmit={submit}>
+            <BlockStack gap="400">{formContent}</BlockStack>
+          </Form>
+        </div>
```
MillanWangGadget, line 141: When the form has inputs added conditionally, the selection will change and refetch the `findBy` record in `useActionForm`. This will cause `isLoading` to be true again. 

This area now makes the form `hidden` so that the children can stay rendered and so that they don't get their `useStates` reset

infiton: :+1: nice

infiton (approved):

alanko0511 (approved):

PR #732, opened 2025-01-22
Feature/shadcn skeleton

jadewale:
The Shadcn Skeleton which gives us the ability to integrate Shadcn and Gagdet

## PR Checklist

Occasionally that error was coming up so added `width` & `height`
Screenshot 2025-01-22 at 08 22 46

Also added a click & Type command as was experiencing issues with Shadcn input components in cypress
https://github.com/cypress-io/cypress/issues/5830
 https://github.com/cypress-io/cypress/issues/5827


packages/react/cypress/component/auto/form/AutoFormDynamicFormInputChanges.cy.tsx
```diff
@@ -12,8 +12,8 @@ describeForEachAutoAdapter("AutoForm", ({ name, adapter: { AutoForm }, wrapper }
         <AutoForm action={api.widget.update} findBy={findBy} />
 
         <div>
-          <button id="setFindById1" onClick={() => setFindBy("1")} />
-          <button id="setFindById2" onClick={() => setFindBy("2")} />
+          <button style={{ width: 1, height: 1 }} id="setFindById1" onClick={() => setFindBy("1")} />
+          <button style={{ width: 1, height: 1 }} id="setFindById2" onClick={() => setFindBy("2")} />
```
airhorns, line 16: Shouldn't buttons have a default style that makes them have dimensions and be findable? this works but i wouldn't have thought it was necessary

packages/react/cypress/support/auto.tsx
```diff
@@ -3,16 +3,23 @@ import "@shopify/polaris/build/esm/styles.css";
...
+import { SUITE_NAMES } from "./constants.js";
 
 interface AutoSuiteConfig {
   name: string;
   adapter: AutoAdapter;
   wrapper: ComponentType<{ children: ReactNode }>;
 }
 
+const ONLY_RUN_SUITES = {
+  [SUITE_NAMES.SHADCN]: ["AutoButton"],
```
airhorns, line 20: I am down for this, but it does mean all the autoinputs stuff that is landing in this PR is untested for now, right?

airhorns (approved):
LGTM

packages/react/cypress/component/auto/form/AutoFormDynamicFormInputChanges.cy.tsx
jadewale: Because there is no label or text inbetween the button, it won't have anything.
If we did <button>hello</button> then the button will take up the space of the `hello`

PR #734, opened 2025-01-22
Shadcn Encrypted Auto Text Inputs

jadewale:
Integrating Shadcn AutoInputs for Encrypted Texts e.g Password Inputs

## PR Checklist

### DIFF
--- a/packages/react/src/auto/shadcn/inputs/ShadcnAutoEncryptedStringInput.tsx
+++ b/packages/react/src/auto/shadcn/inputs/ShadcnAutoEncryptedStringInput.tsx
@@ -1,0 +1,32 @@
+import { EyeIcon, EyeOffIcon } from "lucide-react";
+import React, { useState } from "react";
+import type { Control } from "../../../useActionForm.js";
+import { autoInput } from "../../AutoInput.js";
+import type { ShadcnElements } from "../elements.js";
+import { makeShadcnAutoTextInput } from "./ShadcnAutoTextInput.js";
+
+export const makeShadcnAutoEncryptedStringInput = ({ Input, Label, Button }: Pick<ShadcnElements, "Input" | "Label" | "Button">) => {
+  const TextInput = makeShadcnAutoTextInput({ Input, Label });
+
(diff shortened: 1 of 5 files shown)


packages/react/src/auto/shadcn/inputs/ShadcnAutoEncryptedStringInput.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { makeShadcnAutoTextInput } from "./ShadcnAutoTextInput.js";
+
+export const makeShadcnAutoEncryptedStringInput = ({ Input, Label, Button }: Pick<ShadcnElements, "Input" | "Label" | "Button">) => {
+  const TextInput = makeShadcnAutoTextInput({ Input, Label });
+
+  function ShadcnAutoEncryptedStringInput(props: { field: string; control?: Control<any>; className?: string; suffix?: React.ReactNode }) {
+    const [isShown, setIsShown] = useState(false);
+    const { suffix, ...restProps } = props;
+
+    const showHideToggleButton = (
```
jadewale, line 15: @airhorns should users be supplying their password input mask / unmask buttons ? or this okay

packages/react/src/auto/shadcn/inputs/ShadcnAutoPasswordInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <EncryptedInput
+        {...props}
+        {...(isEditing
+          ? { placeholder: "Password" }
+          : {
+              placeholder: existingPasswordPlaceholder,
+              disabled: true,
+              className: "pr-20",
+              suffix: (
+                <Button
```
jadewale, line 42: as per last comment with supplying custom mask/unmask

packages/react/src/auto/shadcn/inputs/ShadcnAutoEncryptedStringInput.tsx
airhorns: No, its ok that they can't. As long as you use the Button element I am ok with this kind of thing. Its a kinda arbitrary decision we've made on a case by case basis, but this is fine for now and we can accept future PRs adding more customization.

airhorns (approved):

PR #735, opened 2025-01-22
Integerate Autoform

jadewale:

## PR Checklist


packages/react/src/auto/shadcn/inputs/ShadcnAutoInput.tsx
```diff
@@ -15,40 +16,44 @@ export const makeShadcnAutoInput = ({
...
+    () => ({
+      idInput: makeShadcnAutoIdInput({ Input, Label }),
+      textInput: makeShadcnAutoTextInput({ Input, Label }),
+      numberInput: makeShadcnAutoNumberInput({ Input, Label }),
+      encryptedInput: makeShadcnAutoEncryptedStringInput({ Input, Label, Button }),
+      passwordInput: makeShadcnAutoPasswordInput({ Input, Label, Button }),
+      booleanInput: makeShadcnAutoBooleanInput({ Checkbox, Label }),
+    }),
+    [Input, Label, Checkbox, Button]
+  );
```
airhorns: I don't think its legal to use `useMemo` outside a component, and this factory function is outside a component. I think you can just make these local variables though, and then the component below will close over them!

packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+      Omit<Partial<FormProps>, "action"> & { findBy: any };
+
+    validateAutoFormProps(props);
+
+    // Component key to force re-render when the action or findBy changes
+    const componentKey = `${action.modelApiIdentifier ?? ""}.${action.operationName}.${JSON.stringify(findBy)}`;
+
+    const ShadcnAutoInput = makeShadcnAutoInput({ Input, Label, Button, Checkbox });
+
+    const ShadcnAutoSubmit = makeShadcnAutoSubmit({ Button });
```
airhorns, line 33: I think this could be the source of your rendering woes -- this is re-defining and re-creating the components for rendering each time. I think you can define them once in the factory function, and then return the component that just refers to them like this:


```suggestion
export const makeAutoForm =
  <Elements extends ShadcnElements>({ Form, Input, Button, Alert, Skeleton, AlertTitle, AlertDescription, Label, Checkbox }: Elements) => {
      const ShadcnAutoInput = makeShadcnAutoInput({ Input, Label, Button, Checkbox });

    const ShadcnAutoSubmit = makeShadcnAutoSubmit({ Button });


  return <GivenOptions extends OptionsType, SchemaT, ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any>>(
    props: AutoFormProps<GivenOptions, SchemaT, ActionFunc> & ComponentProps<typeof Form>
  ) => {
    const { action, findBy } = props as AutoFormProps<GivenOptions, SchemaT, ActionFunc> &
      Omit<Partial<FormProps>, "action"> & { findBy: any };

    validateAutoFormProps(props);

    // Component key to force re-render when the action or findBy changes
    const componentKey = `${action.modelApiIdentifier ?? ""}.${action.operationName}.${JSON.stringify(findBy)}`;

```

packages/react/src/auto/shadcn/inputs/ShadcnAutoInput.tsx
```diff
@@ -15,39 +16,40 @@ export const makeShadcnAutoInput = ({
   Checkbox,
   Button,
 }: Pick<ShadcnElements, "Input" | "Label" | "Checkbox" | "Button">) => {
-  function ShadcnAutoInput(props: { field: string; label?: string }) {
+  const inputMakers = {
```
jadewale, line 19: @airhorns I switched from memoizing here and did the invocation outside. Is this okay?

packages/react/cypress/component/auto/form/AutoFormInputLabels.cy.tsx
```diff
@@ -11,9 +11,13 @@ import { api } from "../../../support/api.js";
 import { describeForEachAutoAdapter } from "../../../support/auto.js";
 import { SUITE_NAMES } from "../../../support/constants.js";
 
-const ShadcnAutoInput = makeShadcnAutoInput({ Input, Label, Checkbox, Button });
+function useShadcnAutoInput() {
+  return React.useMemo(() => makeShadcnAutoInput({ Input, Label, Checkbox, Button }), []);
+}
```
airhorns, line 16: I don't think this is super realistic -- users aren't gonna wrap their factories in a memo, you know? I think they are just gonna have one big file that defines all their autocomponents once as module scope variables, and then they're gonna import from there. Does it still work if you leave it as it was?

jadewale: We don't need the hook and memoization. I removed it. Thanks

packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
jadewale: Thanks . This was actually fine as it was the `autotextInput` but still added the optimisation you suggested

packages/react/src/auto/shadcn/inputs/ShadcnAutoInput.tsx
airhorns: Yep this looks right to me! I think the `inputMakers` name is just a little confusing in that these arent the makers anymore, they are the made components. I might call this `inputComponents` or something like that instead!

packages/react/src/auto/shadcn/inputs/ShadcnAutoTextInput.tsx
```diff
@@ -39,7 +39,7 @@ export const makeShadcnAutoTextInput = ({ Input, Label }: Pick<ShadcnElements, "
             name={name}
             type={type || "text"}
             id={id}
-            value={value}
+            value={value ?? ""}
```
airhorns, line 42: 👍

airhorns (approved):
LGTM, can address that one comment in the next PR if you want!

PR #737, opened 2025-01-24
Fix initial form flash for autoform with findby

MillanWangGadget:
- https://github.com/gadget-inc/js-clients/pull/729 -> introduced a small regression where a quick flash would happen when the form would render, but all of the inputs were null since they had not been registered in the form context.
- As the fields were getting registered, they would contribute to a selection change that would control the `findby` lookup. During this lookup, the form would go back to the skeleton until the values were loaded, causing this flash
- Previously  (slo mo)
  - CleanShot 2025-01-23 at 21 07 28
- Now (slo mo)
  - CleanShot 2025-01-23 at 21 08 17

### DIFF
--- a/packages/react/cypress/component/auto/form/PolarisAutoJSONInput.cy.tsx
+++ b/packages/react/cypress/component/auto/form/PolarisAutoJSONInput.cy.tsx
@@ -72,7 +72,6 @@ describe("PolarisJSONInput", () => {
       {
         method: "POST",
         url: `${api.connection.options.endpoint}?operation=widget`,
-        times: 1,
       },
       {
         body: {
@@ -91,7 +90,7 @@ describe("PolarisJSONInput", () => {
           },
         },
       }
-    );
+    ).as("widget");
 
     cy.mountWithWrapper(
       <PolarisAutoForm action={api.widget.update} findBy="1">
@@ -101,6 +100,7 @@ describe("PolarisJSONInput", () => {
       PolarisWrapper
     );
 
+    cy.wait("@widget");
     cy.get(`textarea[name="widget.metafields"]`).should(
       "have.value",
       `{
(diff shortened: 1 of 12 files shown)


packages/react/src/auto/AutoFormContext.tsx
```diff
@@ -63,8 +63,14 @@ export const useFieldsFromChildComponents = () => {
 
 export const AutoFormFieldsFromChildComponentsContext = React.createContext<AutoFormFieldsFromChildComponents | undefined>(undefined);
 
-export const AutoFormFieldsFromChildComponentsProvider = ({ children }: { children: React.ReactNode }) => {
-  const fieldsFromChildComponents = useInitializeFieldsFromChildComponents(React.Children.count(children) > 0);
```
MillanWangGadget, line 67: Big mistake from me here. This wrapper will always wrap the `AutoForm` component so it previously would always have children. What this should really do is only be true when there are children within the form. 

This is now passed in as a prop

packages/react/cypress/component/auto/form/PolarisAutoJSONInput.cy.tsx
```diff
@@ -250,22 +251,23 @@ describe("PolarisJSONInput", () => {
...
+    cy.wait("@widget");
+
     cy.get(`textarea[name="widget.metafields"]`).clear();
     cy.get(`textarea[name="widget.metafields"]`).should("have.value", "");
 
     cy.intercept(
       {
         method: "POST",
-        url: `${api.connection.options.endpoint}?operation=createWidget`,
-        times: 1,
```
MillanWangGadget, line 260: `times: 1` can cause flakes when auto forms have custom children. This is because the registration of the custom components will update the field selection in the `findBy` which often cancels the initial findBy and refetches with an updated selection

infiton (approved):

PR #738, opened 2025-01-24
Feature/relationship auto inputs

jadewale:
Load More / Pagination has not been Implemented.

Screenshot 2025-01-27 at 09 55 27
Screenshot 2025-01-27 at 09 55 20
Screenshot 2025-01-27 at 09 55 14
Screenshot 2025-01-27 at 09 55 07
Screenshot 2025-01-27 at 09 54 54
Screenshot 2025-01-27 at 09 54 49


packages/react/src/auto/shadcn/inputs/relationships/RelatedModelOption.tsx
```diff
@@ -0,0 +1,92 @@
+import React from "react";
+import type { Option } from "../../../interfaces/AutoRelationshipInputProps.js";
+import type { ShadcnElements } from "../../elements.js";
+import { makeShadcnListMessages } from "./ShadcnListMessages.js";
+
+// TODO: This is Polaris's AutoSelection enum, but it might not be needed for shadcn.
+export declare enum AutoSelection {
```
airhorns, line 7: I am into this prop, but, is it actually used in the implementation? I can't see it being used so if not I think we should :fire: and if so, is the `declare` keyword actually needed here?

```suggestion
export enum AutoSelection {
```

packages/react/src/auto/shadcn/inputs/relationships/RelatedModelOption.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export const makeRelatedModelOption = ({
+  CommandItem,
+  CommandList,
+  CommandEmpty,
+  CommandGroup,
+  Checkbox,
+  Label,
+}: Pick<ShadcnElements, "CommandItem" | "CommandList" | "CommandEmpty" | "CommandGroup" | "Checkbox" | "Label">) => {
+  const RelatedModelOption = React.memo(function RelatedModelOption(props: RelatedModelOptionsProps) {
```
airhorns, line 37: I don't think this needs to be memoized, it can just be a named function

packages/react/src/auto/shadcn/inputs/relationships/RelatedModelOption.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const { ListMessage, NoRecordsMessage, ShadcnSelectableOption, getErrorMessage } = React.useMemo(
+      () =>
+        makeShadcnListMessages({
+          CommandItem,
+          CommandEmpty,
+          Checkbox,
+          Label,
+        }),
+      [CommandItem, CommandEmpty, Checkbox, Label]
+    );
```
airhorns, line 49: I think this should be moved out of the inner component body and only defined once and not memoized

packages/react/src/auto/shadcn/inputs/ShadcnAutoRolesInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const roleOptions = React.useMemo(
+      () =>
+        options?.map((option) => ({
+          id: option.value,
+          label: option.label,
+        })) ?? [],
+      [options]
+    );
+
+    const recordOptions = React.useMemo(
```
airhorns, line 84: tiny thing but it'd be good to stay consistent between `React.useMemo` and `useMemo` by itself, I kinda prefer the latter. you dont have to go and fix it, just going forward 👍

packages/react/src/auto/shadcn/inputs/relationships/SelectedRelatedRecordTags.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }) {
+    const { selectedRecords, optionLabel, onRemoveRecord } = props;
+
+    const options = getRecordsAsOptions(selectedRecords, optionLabel);
+
+    if (!options.length) {
+      return null;
+    }
+
+    //TODO: why is the key not the id?
```
airhorns, line 21: Can you explain?

packages/react/src/auto/shadcn/inputs/relationships/ShadcnListMessages.tsx
```diff
@@ -0,0 +1,87 @@
+import React from "react";
```
airhorns, line 1: The AutoHasOneInput file above is empty -- that coming in another PR?

packages/react/src/auto/shadcn/inputs/relationships/ShadcnListMessages.tsx
```diff
@@ -0,0 +1,87 @@
...
+        </CommandItem>
+      );
+    }
+
+    //TODO: ask for an example of what a listbox action looks like
+    /**
+     *  <Listbox.Action key={id} value={id} selected={selected}>
+          {label}
+        </Listbox.Action>
+     */
```
airhorns, line 72: Huh?

packages/react/src/auto/shadcn/inputs/relationships/ShadcnListMessages.tsx
jadewale: Yes. I have it done already, although it seems it might have been modified

packages/react/src/auto/shadcn/inputs/relationships/SelectedRelatedRecordTags.tsx
jadewale: once again, me asking myself questions to come back to. Please ignore @airhorns

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoHasManyInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+  Badge,
+  Button,
+  Command,
+  CommandItem,
+  CommandInput,
+  Label,
+  CommandList,
+  CommandEmpty,
+  CommandGroup,
+  Checkbox,
```
MillanWangGadget, line 20: I don't see this checkbox showing up in storybook for hasMany in the same way that there is a checkbox in the roles input. Could you make those look consistent?

packages/react/src/auto/shadcn/inputs/ShadcnAutoRolesInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+    if (loading || !options || options.length === 0) {
+      // Don't render until role options exist. There must always be at least one role option `unauthenticated`
+      return null;
+    }
+
+    return (
+      <ShadcnComboInput
+        {...props}
+        options={roleOptions}
+        path={""}
```
MillanWangGadget, line 110: Out of curiousity, why is this an empty string?

jadewale: @MillanWangGadget I didn't see it being used for the RolesInput, but just looked deeper and saw it is in the `useRolesInputController` so add it as path={fieldProps.name}. Thanks

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoHasManyInput.tsx
jadewale: good catch 
Screenshot 2025-01-27 at 14 41 21

packages/react/src/auto/shadcn/inputs/relationships/ShadcnListMessages.tsx
```diff
@@ -0,0 +1,87 @@
...
+  CommandItem,
+  Checkbox,
+  Label,
+}: Pick<ShadcnElements, "CommandEmpty" | "CommandItem" | "Checkbox" | "Label">) => {
+  function NoRecordsMessage(props: { message?: string }) {
+    const { message = "No records found" } = props;
+    return <CommandEmpty>{message}</CommandEmpty>;
+  }
+
+  function ShadcnSelectableOption(props: SelectableOptionProps) {
```
MillanWangGadget, line 24: CleanShot 2025-01-27 at 15 04 28

in here, there's a small clickable area between the checkbox and the label where you can change the form state without changing the checkbox state

packages/react/src/auto/shadcn/inputs/relationships/ShadcnListMessages.tsx
```diff
@@ -0,0 +1,87 @@
...
+            e.stopPropagation();
+          }}
+          onSelect={() => {
+            onSelect?.(id);
+          }}
+          onKeyDown={(e: React.KeyboardEvent) => {
+            if (e.key === "Enter") {
+              onSelect?.(id);
+            }
+          }}
```
MillanWangGadget, line 45: This does not change the checkbox

jadewale: My commit should address this change and the change above @MillanWangGadget

MillanWangGadget (approved):
- 🍏 Storybook is setup well
- 🍏 hasMany input renders properly and makes correct API requests 
- 🍏 ^ Roles input - also good
- 🍏 ^ belongsTo input - also good

Initial pass of these inputs are good. To keep PRs smaller, I think this is ok to ship now and follow up with the more intense relationship stuff like pagination and search-by-api for the related model record selectors

PR #741, opened 2025-01-27
Update @gadgetinc/react version to be 18.10 alpha

MillanWangGadget:
New version 0.19.0 

Currently as alpha until further notice

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.18.9",
+  "version": "0.18.10-alpha.1",
   "files": [
     "README.md",
     "dist/**/*",
--- a/scripts/check-dependency-versions.ts
+++ b/scripts/check-dependency-versions.ts
@@ -18,7 +18,9 @@ for (const filepath of await globby("packages/*/package.json")) {
     for (const [target, requiredVersion] of Object.entries(targets)) {
       const dependencyVersion = manifest[key]?.[target];
       if (dependencyVersion) {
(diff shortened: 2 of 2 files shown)


packages/react-bigcommerce/package.json
```diff
@@ -40,7 +40,6 @@
   },
   "peerDependencies": {
     "react": "^18.0.0",
-    "react-dom": "^18.0.0",
-    "@gadgetinc/react": "^0.18.4"
```
MillanWangGadget, line 44: CI version check fails if this peer dep is left behind with the react pack going to 19

airhorns: Sadly that's a no go -- the peer dep matters to end users who install, and the dev dep only matters to us who install the monorepo. I think we maybe aughta make the CI version check work with the alpha version?

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/react",
-  "version": "0.18.9",
+  "version": "0.18.10-alpha.1",
```
airhorns, line 3: Does a major break things?

packages/react-bigcommerce/package.json
MillanWangGadget: - Updates
  - Changed CI check to work around `-alpha` in the version

packages/react/package.json
MillanWangGadget: yes. The CI breaks with `0.19.0 does not work with >0.18.4`

airhorns: ok

MillanWangGadget: Am I good to merge this ?

PR #742, opened 2025-01-27
Remove Autoform  HasManyThrough fields till it is implemented

jadewale:
This Input which was skipped initially is causing the main to fail


MillanWangGadget (approved):
Only test related changes. Good to ship on green 👍

PR #743, opened 2025-01-27
Feature/shadcn auto inputs enum hasone

jadewale:

## PR Checklist


MillanWangGadget (commented):
- 🟡 When `allowOther` is enabled, the `add option` element is only visible when there are no search matches. Could you make it such that it always appears when there is a search value, regardless of matches? 
  - CleanShot 2025-01-28 at 11 18 45
- 🍏 Select multiple, defined options only works and makes correct API requests
- 🍏 Select multiple, other strings allowed ^same 
- 🍏 Select one, defined options only ^same 
- 🍏 Select one, other strings allowed ^same

packages/react/cypress/component/auto/form/ShadcnAutoEnumInput.cy.tsx
```diff
@@ -0,0 +1,395 @@
...
+import { elements } from "../../../../spec/auto/shadcn-defaults/index.js";
+import { apiTriggerOnly } from "../../../../spec/auto/support/Triggers.js";
+import { getStadiumRecord } from "../../../../spec/auto/support/stadiumModel.js";
+import { makeAutocomponents } from "../../../../src/auto/shadcn/index.js";
+import { api } from "../../../support/api.js";
+import { ShadcnWrapper } from "../../../support/auto.js";
+
+const ShadcnAutoForm = makeAutocomponents(elements).AutoForm;
+
+describe("ShadcnAutoEnumInput", () => {
```
airhorns, line 11: Are we able to use shared suites here that test that this works roughly the same as polaris? I feel like its gonna be a real pain in the ass for the next person to deduplicate

airhorns (approved):
code LGTM

packages/react/cypress/component/auto/form/ShadcnAutoEnumInput.cy.tsx
jadewale: The element targets are quite different, so it might not be possible to use a shared suite . Although maybe if we switched away from targeting things with id and used either `data-cy="submit"`

jadewale: @airhorns I had a similar case in `AutoRolesInput` where we used a shared suite both I target elemets differently per suite . would you prefer that going forward?

` it("can render a role input, make a selection, and submit the role key", () => {
 
    if (name === SUITE_NAMES.POLARIS) {
      polarisAutoRoleInput();
    }

    if (name === SUITE_NAMES.SHADCN) {
      shadcnAutoRoleInput();
    }


    submit("Widget");
  });
});`

MillanWangGadget (approved):
Previous concerns are addressed. Good to merge 👍

PR #745, opened 2025-01-30
Initial readonly pass of Shadcn AutoTable

MillanWangGadget:
Initial readonly pass of Shadcn AutoTable - Simply to start the momentum on the new AutoTable. Many further updates to come


packages/react/src/auto/polaris/tableCells/PolarisAutoTableFileCell.tsx
```diff
@@ -1,38 +1,14 @@
...
-      }
-
-      // Join the name and extension back together
-      return [name, extension].join(".");
-    } else {
-      // Return the original name if there's no extension
-      return originalFileName;
-    }
-  }, [originalFileName]);
+  const { originalFileName, formattedFileName, imageFileIconUrl } = useAutoTableFileCell(props);
```
airhorns, line 8: 🙏 thanks for refactoring

packages/react/src/auto/shadcn/table/cells/ShadcnAutoTableStyles.css
```diff
@@ -0,0 +1,23 @@
+.shadcn-auto-table-cell {
```
airhorns, line 1: Do we do other css files like this in the polaris implementation?

packages/react/src/auto/shadcn/table/cells/ShadcnAutoTableTextCell.tsx
```diff
@@ -0,0 +1,39 @@
...
+            </TooltipTrigger>
+          </Tooltip>
+        </TooltipProvider>
+      </div>
+    );
+  }
+
+  return (props: { value: any }) => {
+    return <ShadcnAutoTableTextCell {...props} />;
+  };
```
airhorns, line 38: Is this wrapper necessary? I think we could just return the function above?

packages/react/src/auto/shadcn/table/cells/ShadcnAutoTableTagCell.tsx
```diff
@@ -0,0 +1,52 @@
...
+            </Tooltip>
+          </TooltipProvider>
+        </div>
+      </>
+    );
+  }
+
+  return (props: { value: any }) => {
+    return <ShadcnAutoTableTagCell {...props} />;
+  };
```
airhorns, line 51: same wrapper question as below

airhorns (approved):
code LGTM after comments, i feel like we can tophat and fix once its all working together?

packages/react/src/auto/shadcn/ShadcnAutoTable.tsx
```diff
@@ -0,0 +1,138 @@
...
+                          <ShadcnAutoTableCellRenderer column={column} value={row[column.identifier] as ColumnValueType} />
+                        )}
+                      </TableCell>
+                    ))}
+                  </TableRow>
+                ))
+              ) : (
+                <TableRow>
+                  <TableCell colSpan={columns.length} className="h-24 text-center">
+                    No results.
```
infiton, line 108: this is what we want for empty state?

packages/react/src/auto/shadcn/table/cells/ShadcnAutoTableStyles.css
MillanWangGadget: yes in `packages/react/src/auto/polaris/styles/rich-text.css`, but that might be a special case

packages/react/src/auto/shadcn/table/cells/ShadcnAutoTableTextCell.tsx
MillanWangGadget: Removed unnecessary wrappers in return statements across the board 👍

packages/react/src/auto/shadcn/table/cells/ShadcnAutoTableTagCell.tsx
MillanWangGadget: Removed unnecessary wrappers in return statements across the board 👍

packages/react/src/auto/shadcn/ShadcnAutoTable.tsx
MillanWangGadget: Updated this empty state message to have the same text as the Polaris equivalent

PR #749, opened 2025-01-30
Separated `autoInput` and new `autoRelationshipForm` to prevent contribution of `selectPaths` to all input props

MillanWangGadget:
- **UPDATE**
  - Previously, the `selectPaths` prop was appearing in the types for ALL AutoInputs, but it is only relevant to `AutoRelationshipForm` wrapper components 
  - Now, `autoInput` and `autoRelationshipForm` are separate, and `selectPaths` only appears on the relationship form wrapper props type


infiton (approved):
makes sense

PR #750, opened 2025-01-31
Export all the shadcn autoform subcomponents from the factory

airhorns:
Before, the factory for shadcn auto components only returned the top level `<AutoForm/>` component, but now, we export all the other little inner input components as well. This allows users to use the inner components by name when they want to take control, like we do in a bunch of the tests. This also matches the kind of barrel-file export we do with Polaris as well.

[no-changelog-required]


jadewale (approved):
Nice !!

MillanWangGadget (commented):
🍏 Changes make sense. Send it 👍

MillanWangGadget (approved):

PR #752, opened 2025-01-31
Updated AutoForm type to have a type error when given `title` or `submitLabel` when form has children

MillanWangGadget:
- **UPDATE**
  - When an AutoForm has child components, adding `title` or `submitLabel` will do nothing because those elements won't even be rendered. All that will be rendered is the user provided children to `AutoForm`
  - Since those props do nothing in that case, they should be type errors instead to reduce confusion and `what does the prop do` moments


airhorns (approved):
Hnoice

airhorns:
Oh does the shadcn implementation need to be updated as well?

MillanWangGadget:
> Oh does the shadcn implementation need to be updated as well?

Good call, the type check CI step passed because the props were actually `any` on the Shadcn side. I've updated the Shadcn props to have the same type as the main `AutoForm` input

PR #756, opened 2025-02-04
feature/shadcn-auto-relationship-forms

jadewale:

## PR Checklist
Screenshot 2025-02-04 at 16 56 56
Screenshot 2025-02-04 at 16 56 52
Screenshot 2025-02-04 at 16 56 41
Screenshot 2025-02-04 at 16 56 30
Screenshot 2025-02-04 at 16 56 26


packages/react/cypress/component/auto/form/AutoHasOneForm.cy.tsx
```diff
@@ -16,177 +12,184 @@ const originalDoodadLinkedToWidget = {
...
+      cy.wait("@widget");
 
-    cy.get(".Polaris-Button__Icon").first().click();
-    cy.contains("Edit doodad").click();
+      cy.contains("Doodad 1"); // primary label
+      cy.contains("Weight:333"); // secondary label
+      cy.contains("Large"); // tertiary label
 
-    cy.get('input[id="widget.doodad.name"]').should("exist").click().type(" - updated");
-    cy.get('input[id="widget.doodad.weight"]').should("exist").click().type("123");
```
jadewale, line 136: Polaris was able to open a modal which had "Har" and type "ry" which resulted in "Harry" while Shadcn Modal open resulted in the first input taking autofocus, so the click and type was replace the first element with the other element . e.g "Har" and type "ry" was resulting in "ry". Real Users in that case will click at the end of the input to keep typing so it won't be an issue. below is what the UI looked like when a user opens the dialog
Screenshot 2025-02-04 at 16 49 41

packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
```diff
@@ -85,6 +100,39 @@ export const makeAutoForm = <Elements extends ShadcnElements>({
...
+    PopoverTrigger,
+    Dialog,
+    DialogContent,
+    DialogClose,
+    DialogFooter,
+    DialogHeader,
+    DialogTitle,
+    DialogDescription,
+    DialogTrigger,
+  });
```
airhorns, line 134: I think it would result in a slightly smaller bundle to just pass `elements` here, and save us the trouble of updating this list all the time 👍

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoHasOneForm.tsx
```diff
@@ -0,0 +1,274 @@
...
+    // that comes from the response to the action mutation
+    useEffect(() => {
+      if (isSubmitSuccessful && record && !record.id && !("_link" in record) && !("_unlink" in record) && defaultRecordId) {
+        setValue(path + ".id", defaultRecordId);
+      }
+    }, [record, defaultRecordId, path, setValue, submitCount, isSubmitSuccessful]);
+
+    const primaryLabel = useOptionLabelForField(field, props.primaryLabel);
+    const hasRecord = !!(record && !("_unlink" in record && record._unlink));
+    const recordOption = record ? getRecordAsOption(record, primaryLabel, props.secondaryLabel, props.tertiaryLabel) : null;
```
airhorns, line 140: Think we can push some of this logic into the controller or a new utility function? It all seems pretty smart such that duplicating it between the two implementations seems like asking for bugs in one but not the other

packages/react/spec/auto/shadcn-defaults/inputs/form/ShadcnAutoRelationshipForm.stories.jsx
```diff
@@ -0,0 +1,75 @@
...
+import { makeAutocomponents } from "../../../../../src/auto/shadcn/index.ts";
+import { elements } from "../../index.tsx";
+
+// More on how to set up stories at: https://storybook.js.org/docs/writing-stories#default-export
+
+const { AutoForm, AutoHasOneForm, AutoInput, AutoSubmit, SubmitResultBanner } = makeAutocomponents(elements);
+
+const Component = (props) => {
+  return (
+    <AutoForm {...props} action={api.widget.create}>
```
MillanWangGadget, line 16: the `action` prop here is overriding the action that you pass in from the stories. It's safe to remove since you're passing in the action in both cases

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoHasOneForm.tsx
jadewale: okay, working on it.

packages/react/src/auto/shadcn/inputs/relationships/RelatedModelOption.tsx
```diff
@@ -88,7 +102,75 @@ export const makeRelatedModelOption = (
...
+            />
+          </ScrollArea>
+        </PopoverContent>
+      </Popover>
+    );
+  }
+
+  function RelatedModelOptionsSearch(props: {
+    modelName: string;
+    label?: string;
```
MillanWangGadget, line 149: Label is currently unused. Is it needed?

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoHasOneForm.tsx
```diff
@@ -0,0 +1,274 @@
...
+    const { path, metadata } = useAutoRelationship({ field });
+    const {
+      setValue,
+      getValues,
+      formState: { defaultValues, submitCount, isSubmitSuccessful },
+    } = useFormContext();
+
+    const { record, relatedModelOptions } = useHasOneController(props);
+    const [actionsOpen, setActionsOpen] = useState(false);
+    const [searchOpen, setSearchOpen] = useState(false);
```
MillanWangGadget, line 116: unused

MillanWangGadget (commented):
Some minor things but looks good otherwise after merge conflict resolution

packages/react/src/useHasOneForm.ts
```diff
@@ -0,0 +1,81 @@
...
+  tertiaryLabel?: OptionLabel;
+}) => {
+  const { path, metadata } = useAutoRelationship({ field: props.field });
+  const {
+    setValue,
+    getValues,
+    formState: { defaultValues, submitCount, isSubmitSuccessful },
+  } = useFormContext();
+
+  const { record, relatedModelOptions } = useHasOneController(props);
```
airhorns, line 23: Do we use this other hook anywhere else? If not i think we should merge them

jadewale: I use it with Polaris. I changed PolarisAutoHasForm to use `useHasOneForm.ts`

PR #758, opened 2025-02-05
Feature/Shadcn-AutoBelongsForm

jadewale:

## PR Checklist

Screenshot 2025-02-04 at 22 55 53
Screenshot 2025-02-04 at 22 55 46
Screenshot 2025-02-04 at 22 55 41

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
@@ -1,170 +1,173 @@
 import React from "react";
-import { PolarisAutoForm } from "../../../../src/auto/polaris/PolarisAutoForm.js";
-import { PolarisAutoInput } from "../../../../src/auto/polaris/inputs/PolarisAutoInput.js";
-import { PolarisAutoBelongsToForm } from "../../../../src/auto/polaris/inputs/relationships/PolarisAutoBelongsToForm.js";
-import { PolarisAutoSubmit } from "../../../../src/auto/polaris/submit/PolarisAutoSubmit.js";
-import { PolarisSubmitResultBanner } from "../../../../src/auto/polaris/submit/PolarisSubmitResultBanner.js";
 import { api } from "../../../support/api.js";
-import { PolarisWrapper } from "../../../support/auto.js";
+import { describeForEachAutoAdapter } from "../../../support/auto.js";
+import { SUITE_NAMES } from "../../../support/constants.js";
 
 const originalSectionLinkedToWidget = { id: "1", name: "Section 1", label: "Label 1" };
 
-describe("PolarisAutoBelongsToForm", () => {
-  const interceptModelUpdateActionMetadata = () => {
-    cy.intercept({ method: "POST", url: `${api.connection.endpoint}?operation=ModelActionMetadata` }, RealWidgetMetadata).as(
-      "ModelCreateActionMetadata"
-    );
-  };
+describeForEachAutoAdapter(
+  "AutoBelongsToForm",
+  ({ name, adapter: { AutoForm, AutoInput, AutoSubmit, SubmitResultBanner, AutoBelongsToForm }, wrapper }) => {
+    const interceptModelUpdateActionMetadata = () => {
+      cy.intercept({ method: "POST", url: `${api.connection.endpoint}?operation=ModelActionMetadata` }, RealWidgetMetadata).as(
+        "ModelCreateActionMetadata"
+      );
+    };
 
-  const expectUpdateActionSubmissionVariables = (expectedQueryValue?: any) => {
-    cy.intercept({ method: "POST", url: `${api.connection.endpoint}?operation=updateWidget` }, (req) => {
-      // eslint-disable-next-line
-      expect(req.body.variables).to.deep.equal(expectedQueryValue);
-      req.reply({ data: { updateWidget: { success: true, errors: null, x: {} } } });
-    }).as("updateWidget");
-  };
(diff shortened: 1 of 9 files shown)


packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
```diff
@@ -1,170 +1,173 @@
...
+      });
+      cy.get('[id="submit"]').click();
+      cy.wait("@updateWidget");
     });
-    cy.get('[id="submit"]').click();
-    cy.wait("@updateWidget");
-  });
-});
+  }
+);
```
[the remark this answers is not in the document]
MillanWangGadget: We currently do not have a test for this. The API currently does not return nested action inputs to power a potential `show all possible fields` approach. Right now, Polaris oneliner relationship form components work, but they have weird behaviour with empty modals. I'm gonna investigate this today

packages/react/spec/auto/shadcn-defaults/inputs/form/ShadcnAutoRelationshipForm.stories.jsx
```diff
@@ -9,23 +9,33 @@ import { elements } from "../../index.tsx";
...
-      >
-        <div className="flex flex-col gap-4">
-          <AutoInput field="name" />
-          <AutoInput field="weight" />
-          <AutoInput field="active" />
-          <AutoInput field="size" />
+      <div className="flex flex-col gap-4">
+        <div>
+          <AutoBelongsToForm field="section" primaryLabel="name"
+            renderParent={(record) => <Text>this is a custom belongsTo render for {record.name}</Text>}
```
MillanWangGadget, line 20: This should be `renderSelectedRecord` since `renderParent` is not a prop on the component

packages/react/src/useHasBelongsForm.ts
```diff
@@ -0,0 +1,78 @@
...
+}) => {
+  const { field } = props;
+  const { path, metadata } = useAutoRelationship({ field });
+  const {
+    setValue,
+    getValues,
+    formState: { defaultValues, submitCount, isSubmitSuccessful },
+  } = useFormContext();
+
+  const { record, relatedModelOptions } = useBelongsToController(props);
```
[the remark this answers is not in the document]
jadewale: the `useBelongsToController` and `useHasOneController` are being used in other hooks.  I don't know if we should combine the useForm and the controllers. That will be adding other hooks and props on this hooks that probably don't need them @airhorns  
Screenshot 2025-02-05 at 15 08 26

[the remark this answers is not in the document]
airhorns: Ah ok if they are used in multiple places then keeping them separate makes sense to me!

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoBelongsToForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+              willLoadMoreOptions={pagination.hasNextPage && searchFilterOptions.length >= optionRecordsToLoadCount}
+              onScrolledToBottom={handleScrolledToBottom}
+              onChange={search.set}
+              defaultValue={search.value}
+            />
+          )}
+        </div>
+
+        <Dialog open={modalOpen} onOpenChange={() => setModalOpen(!modalOpen)}>
+          <RelationshipContext.Provider value={{ transformPath: (path) => pathPrefix + "." + path }}>
```
MillanWangGadget, line 205: This is now going to be a type error once you rebase due to a bug fix I shipped here.
https://github.com/gadget-inc/js-clients/pull/760


The fix is easy, you just gotta do something similar to what I did in that PR to the `hasOneForm` hook. Add this to the new hook you added and exported
```ts
const metaDataPathPrefix = relationshipContext?.transformMetadataPath ? relationshipContext.transformMetadataPath(props.field) : props.field;
```

Then do this in the belongsToForm components 
```tsx
       <RelationshipContext.Provider
           value={{
             transformPath: (path) => pathPrefix + "." + path,
             transformMetadataPath: (path) => metaDataPathPrefix + "." + path,
           }}
         >
```

packages/react/src/useHasBelongsForm.ts
```diff
@@ -0,0 +1,78 @@
+import { useAutoRelationship, useRelationshipContext } from "./auto/hooks/useAutoRelationship.js";
+import { useBelongsToController } from "./auto/hooks/useBelongsToController.js";
+import { getRecordAsOption, useOptionLabelForField } from "./auto/hooks/useRelatedModel.js";
+import type { OptionLabel } from "./auto/interfaces/AutoRelationshipInputProps.js";
+import { useFormContext } from "./useActionForm.js";
+import { get } from "./utils.js";
+
+import { useEffect, useState } from "react";
+export const useHasBelongsForm = (props: {
```
MillanWangGadget, line 9: What do you think about renaming this to something like `useBelongsToForm` to be more consistent with the existing `useHasOneForm`?

The word `has` in this hook name could confuse people into thinking that it is related to `hasOne / hasMany / hasManyThrough` fields and not just `belongsTo` fields

MillanWangGadget (commented):
- 🍏 PolarisAutoBelongsToForm behaviour is unchanged 
- 🍏 ShadcnAutoBelongsToForm has equivalent functionality 

Looks good after rebase and comments

packages/react/src/useHasBelongsForm.ts
jadewale: works

MillanWangGadget (approved):
Changes look good 👍

PR #760, opened 2025-02-05
Updated AutoFormRelationshipForms to use metadataPath instead form state path

MillanWangGadget:
- **UPDATE**
  - `autoRelationshipForm` wrapper previously registered `selectPaths` incorrectly by prepending the current field API id in front of the selected field instead of prepending the whole path of the current field. 
    - This would make nested relationship forms unable to show the selected paths in labels without having them as AutoInputs or updating the select 
  - Nesting `AutoHasOneForm` and `AutoBelongsToForms` inside of an `AutoHasManyForm` would cause throw an exception when used
    - This is because those forms included indexes from their parent `HasManyForm` into the paths used for their own relationship form context
      - Eg. Their paths were `parent.children.0.grandchildren.0....` instead of `parent.children.grandchildren...`
      - Fixed by updating the context to use the metadata path, which does not include indexes for the position in arrays in the form

### DIFF
--- a/packages/react/spec/useActionFormNested.spec.tsx
+++ b/packages/react/spec/useActionFormNested.spec.tsx
@@ -2274,88 +2274,52 @@ describe("useActionFormNested", () => {
             "name": "Ottawa",
             "tweeters": [
               {
-                "delete": {
-                  "id": "1",
-                },
+                "followerFriendships": [
+                  {
+                    "follower": {
+                      "id": "102",
+                      "name": "Jim Doe",
+                    },
+                    "id": "3",
+                  },
+                  {
+                    "follower": {
+                      "id": "103",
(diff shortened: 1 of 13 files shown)


packages/react/spec/useActionFormNested.spec.tsx
MillanWangGadget, line 1: The snapshot change in this test is a result of an update to the API client. I updated the `js-clients-test` app to add a series of models that are interconnected via nested relationships, and I encountered an old bug where having model with the same api ID in different namespaces broke the GQL generation. I had to upgrade the app's FWVersion to work around this
 
Updating the api client version and running this test on main will yield the same result

infiton (approved):
relationship master!

PR #762, opened 2025-02-05
Revert "Give records a reference to the model manager which loaded them"

garymardell:
Reverts gadget-inc/js-clients#754


packages/api-client-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/api-client-core",
-  "version": "0.15.39",
+  "version": "0.15.38",
```
garymardell, line 3:
```suggestion
  "version": "0.15.40",
```

PR #765, opened 2025-02-06
Allow sibling edit in AutoHasManyThroughForm

MillanWangGadget:
- UPDATE
  - It was previously impossible to edit the sibling record from within a AutoHasManyThroughForm as the variable passed to the `joinModel-belongsTo-relatedModel` field would always be `_link` instead of a nested action
  - now, that join model belongs to field gets a nested action formed if it is called in the context of a AutoHasManyThroughForm.

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
@@ -92,17 +92,79 @@ describeForEachAutoAdapter(
       cy.get('[id="deleteButton_students.0"]').click();
 
       expectUpdateActionSubmissionVariables({
-        id: "3",
         course: {
           registrations: [
-            // Deleted first
+            //  Deleted first
             { delete: { id: "1" } },
             // Updated second
-            { update: { effectiveFrom: "2024-03-17", effectiveTo: "2025-08-21", id: "25", student: { _link: "11" } } },
+            {
+              update: {
+                effectiveFrom: "2025-02-18",
+                effectiveTo: "2025-02-22",
+                id: "50",
+                student: { update: { firstName: "Benjamin", id: "43", lastName: "Martin" } },
+              },
+            },
             // Created third
             { create: { student: { _link: "10" } } },
           ],
         },
+        id: "3",
+      });
+      cy.get('[id="submit"]').click();
+      cy.wait("@updateCourse");
+    });
+
+    it("can update fields on the sibling model", () => {
+      cy.mountWithWrapper(
+        <AutoForm action={api.university.course.update} findBy="3">
+          <AutoHasManyThroughForm
+            field="students"
+            selectPaths={["firstName", "lastName", "year", "department"]}
+            primaryLabel={["firstName", "lastName"]}
+            secondaryLabel={(record: any) => `Year: ${record.year}`}
+            tertiaryLabel="department"
+          >
+            <AutoInput field="firstName" />
+            <AutoInput field="lastName" />
+          </AutoHasManyThroughForm>
+          <AutoSubmit id="submit" />
+        </AutoForm>,
+        wrapper
+      );
+      cy.wait("@ModelCreateActionMetadata");
+      cy.wait("@course");
+      cy.wait("@students");
+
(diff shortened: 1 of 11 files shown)


packages/react/spec/auto/polaris/form/PolarisAutoRelationshipForm.stories.jsx
```diff
@@ -197,6 +197,7 @@ const ExampleCourseCreateRelatedForm = (props) => {
             <InlineStack>
               <PolarisAutoInput field="registration.effectiveFrom" />
               <PolarisAutoInput field="registration.effectiveTo" />
+              <PolarisAutoInput field="registration.student.firstName" />
```
infiton, line 200: didn't we want to address this as `field="firstName"`?

MillanWangGadget: Good point. This currently does not load in the data for existing records and it makes a GQL error on submit, but I think it can work if I add some extra stuff to the relationship context wrapper to adjust the pathing accordingly

MillanWangGadget: I can confirm that the `edit the join model fields` case requires the joinModel API ID and not the api id of a field associated with the join model with this storybook entry. I left this out because it does not prove anything that existing entries don't prove already, and the file is already quite large 


```jsx
export const HasManyThroughWhereFieldApiIdsAreUnrelatedToModelNames = {
  render: (args) => (
    <>
      <Page>
        <PolarisAutoForm {...args}>
          <PolarisSubmitResultBanner />
          <Card>
            <BlockStack gap="300">
              <FormLayout>
                <PolarisAutoInput field="baseModelName" />
              </FormLayout>
            </BlockStack>
          </Card>

          <Card>
            <PolarisAutoHasManyThroughForm field="baseModelHmtField" selectPaths={["siblingName"]} primaryLabel={"siblingName"}>
              <InlineStack>
                <PolarisAutoInput field="joinerModel.joinModelName" />
                <PolarisAutoInput field="siblingName" />
              </InlineStack>
            </PolarisAutoHasManyThroughForm>
          </Card>
          <PolarisAutoSubmit />
        </PolarisAutoForm>
      </Page>
    </>
  ),
  args: {
    action: api.hasManyThrough.baseModel.update,
    findBy: "1",
    debug: true,
  },
};

```

packages/react/src/auto/polaris/inputs/relationships/PolarisAutoHasManyThroughForm.tsx
```diff
@@ -107,6 +107,10 @@ export const PolarisAutoHasManyThroughForm = autoRelationshipForm(
                                 transformPath: (path) => `${joinModelField}.${idx}.${path.replace(`${joinModelApiIdentifier}.`, "")}`,
                                 transformMetadataPath: (path) => `${metaDataPathPrefix}.${path}`,
                                 fieldArray,
+                                hasManyThrough: {
+                                  joinModelApiIdentifier: joinModelApiIdentifier ?? "",
+                                  inverseRelatedModelField: inverseRelatedModelField ?? "",
```
airhorns, line 112: Why the coalesces to the empty string here?

packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
```diff
@@ -92,17 +92,79 @@ describeForEachAutoAdapter(
...
+                    firstName: "Benjamin- updated",
+                    id: "43",
+                    lastName: "Martin- updated",
+                  },
+                },
+              },
+            },
+          ],
+        },
+        id: "3",
```
airhorns, line 167: 👍

airhorns (approved):

packages/react/src/auto/polaris/inputs/relationships/PolarisAutoHasManyThroughForm.tsx
MillanWangGadget: I originally added these because those values are potentially undefined, but I've updated this so that the hook asserts those values before returning them. The form would not work anyways if the relationship is misconfigured

PR #773, opened 2025-02-08
AutoHasOneForm - UI overhaul, allow selected record replacement, consolidate shared functions with BelongsTo

MillanWangGadget:
- UPDATES
	- removed `renderSelectedRecord`. This is redundant with the primary/secondary/ternary record label system 
	- `useSingleRelatedRecordRelationshipForm` 
		- There was a lot of overlap between the form hooks for belongsTo and hasOne. I've consolidated their shared logic into this one hook and added extensions where needed
			- Ex - HasOne has more complicated logic for creating a new record
	- `SearchableSingleRelatedModelRecordSelector`
		- In Polaris and Shadcn, we had an effectively identical component that was used to search and select for records on the related model. It has now been consolidated into a single component in both Polaris and Shadcn 
	- `hasOne` - GQL argument shaping
		- Previously, the inclusion of `_unlink` would ignore all of the other params in the form state. This made it impossible to unlink a record and replace it with a newly created one at the same time
	- UI overhaul for `AutoHasOneForm`
		- After discussing with designers, we decided that `AutoHasOneForm` should look basically the same as `AutoHasManyForm`, but there is only one related record available
			- Disabled `select existing related model record` functionality. It is now only possible to create new records and edit existing linked records.
			- Replaced the modal system for edit/create with a single unit accordion just like the hasManyForm

### DIFF
--- a/packages/react/src/auto/polaris/inputs/relationships/SearchableSingleRelatedModelRecordSelector.tsx
+++ b/packages/react/src/auto/polaris/inputs/relationships/SearchableSingleRelatedModelRecordSelector.tsx
@@ -1,0 +1,63 @@
+import { AutoSelection, Icon, InlineStack, Listbox } from "@shopify/polaris";
+import { PlusCircleIcon } from "@shopify/polaris-icons";
+import React from "react";
+import { type useSingleRelatedRecordRelationshipForm } from "src/useSingleRelatedRecordRelationshipForm.js";
+import { RelatedModelOptionsPopover, RelatedModelOptionsSearch } from "./RelatedModelOptions.js";
+
+export const SearchableSingleRelatedModelRecordSelector = (props: {
+  form: ReturnType<typeof useSingleRelatedRecordRelationshipForm>;
+  override?: {
+    addNewRecord?: () => void;
+  };
+}) => {
+  const {
+    search,
+    searchOpen,
+    setSearchOpen,
+    pagination,
+    records,
+    isLoading,
+    searchFilterOptions,
+    relatedModelName,
+    path,
+    setValue,
+    setIsEditing,
+  } = props.form;
+  return (
+    <>
+      <RelatedModelOptionsPopover
+        active={searchOpen}
+        activator={
+          <RelatedModelOptionsSearch
+            modelName={relatedModelName}
+            value={search.value}
+            onChange={search.set}
+            onFocus={() => setSearchOpen(true)}
+          />
+        }
+        onClose={() => setSearchOpen(false)}
+        onScrolledToBottom={pagination.loadNextPage}
+        actions={[
+          <Listbox.Action key="add-new-record" value="add-new-record" divider>
+            <InlineStack gap="200">
+              <Icon source={PlusCircleIcon} />
+              Add {relatedModelName}
+            </InlineStack>
+          </Listbox.Action>,
+        ]}
+        options={searchFilterOptions}
+        records={records}
+        onSelect={(record) => {
+          if (record.id === "add-new-record") {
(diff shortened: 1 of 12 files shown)


packages/react/src/auto/polaris/inputs/relationships/SearchableSingleRelatedModelRecordSelector.tsx
```diff
@@ -0,0 +1,63 @@
+import { AutoSelection, Icon, InlineStack, Listbox } from "@shopify/polaris";
+import { PlusCircleIcon } from "@shopify/polaris-icons";
+import React from "react";
+import { type useSingleRelatedRecordRelationshipForm } from "src/useSingleRelatedRecordRelationshipForm.js";
+import { RelatedModelOptionsPopover, RelatedModelOptionsSearch } from "./RelatedModelOptions.js";
+
+export const SearchableSingleRelatedModelRecordSelector = (props: {
+  form: ReturnType<typeof useSingleRelatedRecordRelationshipForm>;
```
airhorns, line 8: 👍

packages/react/src/use-action-form/utils.ts
```diff
@@ -468,8 +468,22 @@ const getHasOneGraphqlApiInput = (props: { input: any; result: any; defaultValue
...
+    } = input;
+
+    const additionalActions =
+      !__replace || Object.keys(nonUnlinkParams).length === 0
+        ? {}
+        : "id" in nonUnlinkParams && nonUnlinkParams.id
+        ? { update: nonUnlinkParams }
+        : { create: nonUnlinkParams };
+
+    return { _unlink, ...additionalActions };
```
airhorns, line 486: Makes sense, also, this must have been real annoying

airhorns (approved):
🎩 🍏

packages/react/src/auto/shadcn/inputs/relationships/SearchableSingleRelatedModelRecordSelector.tsx
```diff
@@ -0,0 +1,107 @@
+import { CommandSeparator } from "cmdk";
+import { PlusIcon } from "lucide-react";
+import { default as React, useCallback } from "react";
+import { type useSingleRelatedRecordRelationshipForm } from "src/useSingleRelatedRecordRelationshipForm.js";
+import { debounce } from "../../../../utils.js";
+import { optionRecordsToLoadCount } from "../../../hooks/useRelatedModel.js";
+import type { ShadcnElements } from "../../elements.js";
+import { makeShadcnAutoComboInput } from "../ShadcnAutoComboInput.js";
+
+export const makeSearchableSingleRelatedModelRecordSelector = ({
```
jadewale, line 10: Everything looks good although I don't really see the need for this extracted component.

jadewale (approved):

packages/react/src/use-action-form/utils.ts
MillanWangGadget: Annoying once now, and never annoying for the end user 😊

packages/react/src/auto/shadcn/inputs/relationships/SearchableSingleRelatedModelRecordSelector.tsx
MillanWangGadget: The intention was to make it easily used as a one liner in the belongsTo and hasOne cases. While it is currently not enabled in The has one case, it will be easier to reenable in the future like this

jadewale: ooh okay ! that works

PR #774, opened 2025-02-08
shadcn fixes 2

airhorns:
- **Fix need for a bunch of custom casting with explicitly always included findBy type**
- **Don't use shadcn form -- it overrides react-hook-form context that we need**
- **Make the shadcn form layout a bit nicer**

### DIFF
--- a/packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
+++ b/packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
@@ -1,13 +1,13 @@
 import type { ActionFunction } from "@gadgetinc/api-client-core";
 import type { ComponentProps } from "react";
-import React from "react";
+import React, { forwardRef } from "react";
 import { FormProvider } from "../../useActionForm.js";
 import { humanizeCamelCase, type OptionsType } from "../../utils.js";
 import type { AutoFormProps } from "../AutoForm.js";
 import { useAutoForm } from "../AutoForm.js";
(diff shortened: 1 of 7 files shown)


packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
```diff
@@ -39,29 +41,27 @@ export const makeAutoForm = <Elements extends ShadcnElements>(elements: Elements
...
   const { SubmitSuccessfulBanner, SubmitErrorBanner, SubmitResultBanner } = makeSubmitResultBanner(elements);
-
-  const { Form, Skeleton } = elements;
-
   const AutoHasOneForm = makeShadcnAutoHasOneForm(elements);
   const AutoBelongsToForm = makeShadcnAutoBelongsToForm(elements);
   const AutoHasManyForm = makeShadcnAutoHasManyForm(elements);
   const AutoHasManyThroughForm = makeShadcnAutoHasManyThroughForm(elements);
+
+  const FormContainer = forwardRef<HTMLFormElement, React.FormHTMLAttributes<HTMLFormElement>>(({ className, ...props }, ref) => {
```
airhorns, line 49: FYI @jadewale -- I had to switch us away from using the `Form` component from shadcn, as it messes with the Autoform context because it also uses `react-hook-form` by default. The version we were using for testing in this repo was edited to not do that, but we can't expect our users to make the same edits, so I just switched to a hardcoded container here instead of accepting a form element.

PR #775, opened 2025-02-09
Explicitly handle null values for Polaris and Shadcn AutoDatePickers

jadewale:
Fix issue where null dates in both Shadcn and Polaris were defaulting to 1970 .

Screenshot 2025-02-09 at 18 17 16

### DIFF
--- a/packages/react/src/auto/shadcn/inputs/ShadcnAutoDateTimePicker.tsx
+++ b/packages/react/src/auto/shadcn/inputs/ShadcnAutoDateTimePicker.tsx
@@ -68,7 +68,12 @@ export const makeShadcnAutoDateTimePicker = ({
 
     const localTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
     const localTime = useMemo(() => {
-      return value ? value : isValidDate(new Date(fieldProps.value)) ? new Date(fieldProps.value) : undefined;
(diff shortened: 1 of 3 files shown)


packages/react/src/auto/shadcn/inputs/ShadcnAutoDateTimePicker.tsx
```diff
@@ -68,7 +68,12 @@ export const makeShadcnAutoDateTimePicker = ({
...
     const localTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
     const localTime = useMemo(() => {
-      return value ? value : isValidDate(new Date(fieldProps.value)) ? new Date(fieldProps.value) : undefined;
+      if (value) return value;
+
+      if (fieldProps.value == null) return undefined; // Prevents null from becoming 1970
+
+      const date = new Date(fieldProps.value);
+      return isValidDate(date) ? date : undefined;
     }, [value, fieldProps.value]);
```
airhorns, line 77: Since we have to make this change in two places, I think it'd be good to refactor this out into a common utility so that the next design system just gets these smarts by default!

airhorns (approved):
LGTM after comments

MillanWangGadget (commented):
Looks good 👍

PR #777, opened 2025-02-10
Feature/shadcn fileinput upload

jadewale:

## PR Checklist


MillanWangGadget (commented):
- 🍏 Adding a file to the input makes the correct API call and has equivalent with Polaris
- 🍏 Retrieving existing records containing files works as well
- 🍏 Image size validation works well 
- 🟡 Could you add a generic file icon when a non image file is added? 
  - You can test this easily by adding an early `return;` to `useFileInputController->onlyImagesValidation`
  - CleanShot 2025-02-10 at 16 46 29@2x

MillanWangGadget:
https://github.com/gadget-inc/js-clients/pull/814 
Replacement here

PR #778, opened 2025-02-10
Update autoRelationship form field type to be object based for displaying related model records

MillanWangGadget:
- UPDATE
   - After consulting design, we will replace the old `primaryLabel`, `secondaryLabel`, and `ternaryLabel` with a new prop called `displayRecord:{primary, secondary, ternary}`
   - The `selectPaths` prop has been removed in favour of inferring the used fields directly from the field name strings used in the `displayRecord` object. 
     - If a callback function is used, we will then try to smartly select fields on the related model
       - This matches the behaviour of `AutoTable` when we have a custom column configuration. If the given columns are based on field names, we use those directly. If a single callback for a custom cell renderer is detected, then we expand the selection to include all non-relationship fields

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
@@ -102,7 +102,13 @@ describeForEachAutoAdapter(
       cy.mountWithWrapper(
         <AutoForm action={api.widget.update} findBy="42">
           <SubmitResultBanner />
-          <AutoBelongsToForm field="section" primaryLabel="name" secondaryLabel="label">
+          <AutoBelongsToForm
+            field="section"
+            displayRecord={{
+              primary: "name",
+              secondary: "label",
+            }}
+          >
             <AutoInput field="name" />
             <AutoInput field="label" />
           </AutoBelongsToForm>
@@ -139,7 +145,13 @@ describeForEachAutoAdapter(
       cy.mountWithWrapper(
         <AutoForm action={api.widget.update} findBy="42">
           <SubmitResultBanner />
-          <AutoBelongsToForm field="section" primaryLabel="name" secondaryLabel="label">
+          <AutoBelongsToForm
+            field="section"
+            displayRecord={{
+              primary: "name",
+              secondary: "label",
+            }}
+          >
             <AutoInput field="name" />
             <AutoInput field="label" />
           </AutoBelongsToForm>
--- a/packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
@@ -104,7 +104,7 @@ describeForEachAutoAdapter(
           <SubmitResultBanner />
           <AutoBelongsToForm
             field="section"
-            displayRecord={{
+            recordLabel={{
               primary: "name",
               secondary: "label",
             }}
@@ -147,7 +147,7 @@ describeForEachAutoAdapter(
           <SubmitResultBanner />
           <AutoBelongsToForm
             field="section"
-            displayRecord={{
+            recordLabel={{
               primary: "name",
               secondary: "label",
             }}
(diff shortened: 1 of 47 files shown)


packages/react/src/auto/hooks/useSelectedPathsFromDisplayRecord.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  if (typeof optionLabel === "string") {
+    return [optionLabel];
+  }
+
+  // Callback instead of explicit selection
+  return getFieldsToSelectOnDisplayRecordCallback?.().filter((field) => !displayRecordCallbackOmittedFields.includes(field)) ?? [];
+};
+
+const displayRecordCallbackOmittedFields = ["createdAt", "updatedAt"];
```
MillanWangGadget, line 61: Having these auto-included in the selection will lead to API errors since they are not visible in the API. There may be more cases like this for other special case fields. Ex. `roles` on user model, `storeHash` on `bigcommerce.store`, probably some Shopify examples etc. 

I'd like to follow up on making this field exclusion system more robust to handle those cases

packages/react/src/useHasManyThroughForm.ts
```diff
@@ -1,25 +1,15 @@
...
-  primaryLabel?: OptionLabel;
-  secondaryLabel?: OptionLabel;
-  tertiaryLabel?: OptionLabel;
-}) => {
+export const useHasManyThroughForm = (props: AutoRelationshipFormProps) => {
   const { field, children } = props;
   const { metadata } = useAutoRelationship({ field });
   const { setValue } = useFormContext();
-  const childPaths = children && extractPathsFromChildren(children);
-  const hasChildForm = childPaths && childPaths.length > 0;
```
MillanWangGadget, line 22: This was the cause of a bug where `AutoHasManyThroughForms` would not render their children at all unless there was an `AutoInput` directly within. This feels incorrect and the forms would never render custom children containing `AutoInputs` due to the `React.children.forEach` approach. 

This has been fixed so that the children will appear regardless of if they have an `autoInput` or not

packages/react/cypress/component/auto/form/AutoBelongsToForm.cy.tsx
```diff
@@ -102,7 +102,13 @@ describeForEachAutoAdapter(
       cy.mountWithWrapper(
         <AutoForm action={api.widget.update} findBy="42">
           <SubmitResultBanner />
-          <AutoBelongsToForm field="section" primaryLabel="name" secondaryLabel="label">
+          <AutoBelongsToForm
+            field="section"
+            displayRecord={{
```
airhorns, line 107: What do you think about calling this prop `display` everywhere? 

```
display="name"
```

```
display={{primary: "name"}}
```

`displayRecord` just feels a bit clunky for something that I think a lot of folks are going to use all the time

MillanWangGadget: I'd be open to that but I think it is nice to have the term `record` in the prop name to really highlight the fact that this controls related records. To somebody on boarding, it may initially give them the impression that this is related to the `display` CSS property, but the type is probably a good clue that it's not.

MillanWangGadget: - As we discussed offline
  - Renamed to `recordLabel`
  - Updated the type to be `recordLabel: OptionLabel | {primary?: OptionLabel, secondary?..., ternary?...}`
    - Receiving an `OptionLabel` will be treated the same as `{primary: OptionLabel}`
  - Made changed the arg in the `OptionLabel`  callback
    - See changeset entry for the breaking change 
    - WE GOTTA BE VERY CAREFUL TO RELEASE A NEW VERSION OF API CLIENT CORE - We'd have to back out this change

packages/react/src/auto/AutoForm.ts
```diff
@@ -224,31 +226,36 @@ export const useAutoForm = <
...
+        getFieldsToSelectOnRecordLabelCallback: (path) =>
+          getAllRelatedModelFieldApiIdentifiers({ path, rootFieldsMetadata: getRootFieldsFromMetadata(metadata) }),
+      })
     : [];
 
   useEffect(() => {
     registerFields(registeredFieldsFromChildren);
   }, [registeredFieldsFromChildren.join(","), registerFields]);
 
+  console.log("registeredFieldsFromChildren :", registeredFieldsFromChildren);
```
airhorns, line 253: :fire:

packages/react/src/useHasManyThroughForm.ts
airhorns: Nice!

infiton (approved):

PR #779, opened 2025-02-12
Add colors to tailwind config for shadcn

jadewale:

## PR Checklist

### DIFF
--- a/packages/react/globals.css
+++ b/packages/react/globals.css
@@ -15,5 +15,9 @@
 
     --muted: 210 40% 96.1%;
     --muted-foreground: 215.4 16.3% 46.9%;
+
+    --accent-foreground: 0 0% 9%;
+    --destructive: 0 84.2% 60.2%;
+    --destructive-foreground: 0 0% 98%;
   }
 }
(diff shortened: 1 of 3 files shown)


packages/react/spec/auto/shadcn-defaults/components/Calendar.tsx
```diff
@@ -26,7 +26,7 @@ function Calendar({ className, classNames, showOutsideDays = true, ...props }: C
         head_cell: "text-muted-foreground rounded-md w-9 font-normal text-[0.8rem]",
         row: "flex w-full mt-2",
         cell: "h-9 w-9 text-center  text-sm p-0 relative [&:has([aria-selected].day-range-end)]:rounded-r-md [&:has([aria-selected].day-outside)]:bg-accent/50 [&:has([aria-selected])]:bg-accent first:[&:has([aria-selected])]:rounded-l-md last:[&:has([aria-selected])]:rounded-r-md focus-within:relative focus-within:z-20",
-        day: cn("", "h-9 w-9 p-0 font-normal aria-selected:opacity-100 text-justify"),
```
jadewale, line 29: I messed with this originally, so ignore. pretty sure their default datepickers will work. doing some integration with Gadget so will test more there.

packages/react/globals.css
```diff
@@ -15,5 +15,9 @@
 
     --muted: 210 40% 96.1%;
     --muted-foreground: 215.4 16.3% 46.9%;
+
+    --accent-foreground: 0 0% 9%;
+    --destructive: 0 84.2% 60.2%;
+    --destructive-foreground: 0 0% 98%;
```
airhorns, line 21: This is only used for tests right? If so, can we move it to the spec dir?

airhorns: ping on this one @jadewale -- this file isn't required to be imported by end users right, just used for our tests?

jadewale: I think end users will need this as part of the tailwind set up. The tailwind config refers to these colors in your global.css

PR #780, opened 2025-02-12
Fix type error for global actions in AutoForm

MillanWangGadget:
- UPDATE
  - Global actions caused type errors when used in `AutoForm`. This was not a problem in older versions, and I'm not certain what caused this.
  - The types for `AutoForm` components have been updated to account for Global actions.
  - The `AutoFormProps` type has been simplified to remove the last 2 type arguments. 
    - They did not contribute anything that could not be inferred from the other type args
    - They resulted in the `defaultValues` type being `Record<string, unknown>` instead of the real type from the API client

### DIFF
--- a/packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
+++ b/packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
@@ -1,4 +1,4 @@
-import type { ActionFunction } from "@gadgetinc/api-client-core";
+import type { ActionFunction, GlobalActionFunction } from "@gadgetinc/api-client-core";
 import type { ComponentProps } from "react";
 import React, { forwardRef } from "react";
 import { FormProvider } from "../../useActionForm.js";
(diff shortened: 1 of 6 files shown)


packages/react/src/auto/shadcn/ShadcnAutoForm.tsx
```diff
@@ -50,14 +50,18 @@ export const makeAutoForm = <Elements extends ShadcnElements>(elements: Elements
...
   });
 
-  function AutoForm<GivenOptions extends OptionsType, SchemaT, ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any>>(
-    props: AutoFormProps<GivenOptions, SchemaT, ActionFunc> & Omit<ComponentProps<typeof FormContainer>, "action">
-  ) {
+  function AutoForm<
+    GivenOptions extends OptionsType,
+    SchemaT,
+    ActionFunc extends ActionFunction<GivenOptions, any, any, SchemaT, any> | GlobalActionFunction<any>
+  >(props: AutoFormProps<GivenOptions, SchemaT, ActionFunc> & Omit<ComponentProps<typeof FormContainer>, "action" | "defaultValue">) {
```
MillanWangGadget, line 57: `Omit<ComponentProps<typeof FormContainer>, "action">`
is now 
`Omit<ComponentProps<typeof FormContainer>, "action" | "defaultValue">`

because otherwise there would be `defaultValue` and `defaultValues` in the props type. Extremely confusing and I think we should only allow people to use the Gadget defaultValues system

PR #782, opened 2025-02-12
reset useActionForm form state to match GQL response for update actions

MillanWangGadget:
- **UPDATE**
  - Previously in AutoRelationshipForms, this happened
    - Workflow 
      - Load form with an update action for an existing record
      - create a new related record in the AutoRelationshipForm component 
      - Submit
      - Click submit again - This resubmits the form such that you create dupes of the records created on the first submit
      - The record now has duplicates of the records first submission. 
    - This is because the IDs of those newly created related  records are not known after the first submission. When the second submit runs, the GQL argument reshaper assumes that the 2nd submit is for more creates since there are no ID
  - Now 
    - ~~The existing related record is refetched after successful submit when the AutoForm has a relationship form child component.~~
    - The form value is reset to match the response of the API 
  - When the form is submitting or refetching, the opacity of the form component is dropped instead of hidden. This prevents big height changes after submitting or when refetching due to conditionally appearing `AutoInputs`


packages/react/src/auto/AutoInput.tsx
```diff
@@ -45,7 +45,11 @@ export function autoRelationshipForm<P extends { field: string }>(
   Component: React.FC<P>
 ): AutoInputComponent<P & { selectPaths?: string[] }> {
   const WrappedComponent: React.FC<P> = (props) => {
-    const { hasCustomFormChildren, registerFields, fieldSet } = useFieldsFromChildComponents();
+    const { hasCustomFormChildren, registerFields, fieldSet, setHasRelationshipForms } = useFieldsFromChildComponents();
+
+    useEffect(() => {
+      setHasRelationshipForms(true);
+    }, [setHasRelationshipForms]);
```
airhorns, line 52: It might be a bit simpler to model this as a boolean argument to the `useFieldsFromChildComponents` function or something like that

airhorns (approved):
This makes good sense to me but I think its complicated enough that it deserves a test -- maybe a mocked response one would make it easy to test sending a weird response back from the server that changed data that you can assert shows up after?

infiton:
hmm I thought that in `useActionForm` we run the same selection against the mutation; but I guess we aren't doing that same thing for related records; I think we could add the id selection into the mutation result and use that instead?

packages/react/src/auto/AutoInput.tsx
MillanWangGadget: Updated such that we don't need to do this in the first place and the API response with the record will now populate the form values

MillanWangGadget:
>  complicated enough that it deserves a test
- Updated to add a test in the `hasManyForm` cy-test to show that double submitting makes arguments of the correct shape assuming proper API responses

> hmm I thought that in `useActionForm` we run the same selection against the mutation; but I guess we aren't doing that same thing for related records; I think we could add the id selection into the mutation result and use that instead?

https://github.com/gadget-inc/js-clients/pull/468/files#diff-9bf476067a84c3481348d15c1e1f4fb9b23373e31c29e6b9df23e57e166a5e45R239-R256

It turns out that `keepValues:true` in this area is the real cause of this bug. This option makes it such that the  action response does not get applied to the form state. Simply removing this fixes the bug, and makes the minor change of form-inputs potentially getting changed after a successful submission so that they match the DB value

infiton (approved):

packages/react/cypress/component/auto/form/AutoBelongsToInput.cy.tsx
MillanWangGadget, line 1: The Shadcn side of this test failed CI since there was a missing `e.preventDefault()` on the button to unlink an existing record. This resulted in a cross test pollution situation where there was an extra `submit form` API request with expected values that were not expected in the next test 💀

Cypress does not auto cancel pending API request between tests 💀 

While battling this situation I merged the Polaris and Shadcn tests together. I will follow up for all of the other split tests

PR #783, opened 2025-02-13
Feature/test shadcn export

jadewale:

## PR Checklist


packages/react/package.json
```diff
@@ -54,6 +54,8 @@
     "pluralize": "^8.0.0",
     "react-fast-compare": "^3.2.2",
     "react-hook-form": "~7.48.2",
+    "cmdk": "^1.0.4",
```
airhorns, line 57: Do we actually import stuff from cmdk, or just rely on components that use it under the hood?

packages/react/cypress/component/auto/form/ShadcnAutoBelongsToInput.cy.tsx
```diff
@@ -0,0 +1,212 @@
+import React from "react";
+import { elements } from "../../../../spec/auto/shadcn-defaults/index.js";
+import { apiTriggerOnly } from "../../../../spec/auto/support/Triggers.js";
+import { makeAutocomponents } from "../../../../src/auto/shadcn/index.js";
+import { api } from "../../../support/api.js";
+import { ShadcnWrapper } from "../../../support/auto.js";
+
+const { AutoForm, AutoSubmit, SubmitResultBanner, AutoBelongsToInput } = makeAutocomponents(elements);
+
+describe("ShadcnAutoBelongsToInput", () => {
```
airhorns, line 10: I think Millan recently made a version of this test that worked against both design libs

airhorns (approved):

packages/react/package.json
jadewale: I don't think we need this actually

PR #785, opened 2025-02-14
Fix jumpiness in AutoRelationshipInput related model records dropdowns when paginating

MillanWangGadget:
- **UPDATE**
  - Previously, scrolling in the related model record drop down was jumpy. It did not always feel like a smooth scroll like it did in the gizmo hasMany records drop down in storybook
  - this issue was caused by the fact that the records are sorted by ID, however, they are sorted by ID as strings instead of as numbers. 
    - eg. `1,10,11,12,13,14,15,16,17,18,19,   2,20,21,22.....`
    - this was probably never noticed because this behaviour was generally tested in the gizmos has many field on the widget model, and all records in the gizmo model have IDs greater than 1000. This is an issue that is more appearance when the model records start at 1
  - Now, the ID strings are converted to number before they are sorted. You can now observe smooth scrolling in all of the auto relationship input components when paginating

### DIFF
--- a/packages/react/src/auto/polaris/inputs/relationships/PolarisListMessages.tsx
+++ b/packages/react/src/auto/polaris/inputs/relationships/PolarisListMessages.tsx
@@ -7,7 +7,7 @@ export const SelectableOption = (props: { primary?: React.ReactNode; id: string;
 
   return !primary ? (
     <Listbox.Option key={id} value={id} selected={selected}>
-      id: {id}
+      {`id: ${id}`}
     </Listbox.Option>
   ) : typeof primary === "string" ? (
     <Listbox.Option key={id} value={id} selected={selected}>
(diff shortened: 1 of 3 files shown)


packages/react/src/auto/polaris/inputs/relationships/PolarisListMessages.tsx
```diff
@@ -7,7 +7,7 @@ export const SelectableOption = (props: { primary?: React.ReactNode; id: string;
 
   return !primary ? (
     <Listbox.Option key={id} value={id} selected={selected}>
-      id: {id}
+      {`id: ${id}`}
```
MillanWangGadget, line 10: In polaris, this looks completely different when it is a single literal string versus raw HTML text.

airhorns (approved):
🤘 🤘 🤘

PR #786, opened 2025-02-14
Add AutoRelationshipForms and Shadcn to the index files

MillanWangGadget:
Now that the react pack version is alpha until the auto relationship forms are ready to go public, I've added the components to the exported index files


airhorns (approved):
:shipit: i had the same thing on a branch but its not typechecking so send it!

Edit i think the only diff was this: https://github.com/gadget-inc/js-clients/pull/769/files#diff-fa8a9bfb172ee9ce6bcb9888e94681bd16ebeb885875400f18edbccbb704d81d

PR #787, opened 2025-02-14
Loosen type of AutoForm `defaultValues` prop

MillanWangGadget:
- Unfortunately when I first tried to tighten up the type in this PR, I did not realize that the types for relationship fields would be based on the GQL nested action types instead of JS types. 
  - https://github.com/gadget-inc/js-clients/pull/780
  - if you have `childModel.parentModel.name` in the defaultValues prop, you would get a type error saying `only "create", 'update', 'delete', '_link' are valid property names`. This is wrong because those would be treated like field names instead of nested actionss in useActionForm
- Reshaping the GQL nested action based type to be more like the JS type using only TS type operations is gonna be very hard. 
  - Fixing this type to reflect the JS expected type is probably gonna easier to implement and maintain as an `api-client-core` change


airhorns (approved):
Ah true yes, makes sense

PR #792, opened 2025-02-18
Fix relationship form searching with custom rendered labels, and added `Click to edit...` placeholder to AutoHasOneForm

MillanWangGadget:
- **UPDATES**
  - Related model record searching with custom renderer callbacks 
    - The `search` system that we use for related model records relies on the fullText search that we use in the API. Since the fullText match can result in records with labels that disagree with the user provided search value, there is also a client side filter that ensures we only see options from the API that contain the user's search value. 
      - This client side filter worked well when the record labels were strings, but the filter removed EVERYTHING when the labels were ReactNodes from a callback function. 
        - `[object Object` as a string does not contain the user's search query
    - Now, the client side filter serializes react nodes before checking if they contain the user's search value. Searching now works with custom renderer callbacks 😎 
  - `Click to edit...` placeholder
    - Updated such that the placeholder only shows if there is nothing available across all the `primary, secondary, ternary` 
      - It was weird if you put a valid value into an input corresponding to the secondary label, but were still unable to see that value after confirming the value
    - Added the same placeholder to `AutoHasOneForm` to be consistent
      - Previously, there was no placeholder and there would simply be a completely empty button

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
@@ -70,7 +70,7 @@ describeForEachAutoAdapter(
             field="students"
             recordLabel={{
               primary: ["firstName", "lastName"],
-              secondary: (record: any) => `Year: ${record.year}`,
+              secondary: ({ record }: any) => `Year: ${record.year}`,
               tertiary: "department",
             }}
           >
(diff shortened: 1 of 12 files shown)


packages/react/src/auto/polaris/PolarisAutoForm.tsx
```diff
@@ -103,7 +103,7 @@ const PolarisAutoFormComponent = <
     fields,
   };
 
-  if (fetchingMetadata || (findBy && pauseExistingRecordLookup)) {
+  if (fetchingMetadata) {
```
MillanWangGadget, line 106: If a form component had no only conditionally appearing AutoInputs, we would be stuck in this state with no way of rendering the inputs initially to register them. 

Now, we rely on the recently added `reduce opacity while fetching` system to provide a `loading` type of feedback.

airhorns (approved):

PR #796, opened 2025-02-20
Add error for nesting AutoForm inside another AutoForm

MillanWangGadget:
Nesting an autoForm within another AutoForm causes the submission to break completely. This should be an error case in the first place


airhorns (approved):
hnoice

PR #800, opened 2025-02-21
Error for relationship fields in `AutoHasManyThrough` form + New join model AutoInput system

MillanWangGadget:
- UPDATES
  - Banning Relationships in HMT forms
    - The resulting GQL request is completely incorrect when relationship fields are used within `AutoHasManyThrough` forms. 
      - It seems that the form state does not get updated properly for relationships.
      - This is a problem on both the join model relationships and sibling model relationships
    - For now, an error will prevent users from thinking that this is supported. In the future, it is feasible to add support for this smoothly without breaking change
    - No changes for non-relationship fields being used in a `AutoHasManyThroughForm`. Access to sibling and join model fields is still good
  - New HMT form join record input system
    - previously, we relied on having `field="joinModelName.joinModelField"` inside of AutoInputs inside of HMT form components to refer to join model inputs. This system suggests that it is OK to have `dot.separated.paths` to traverse relationships in autoform, but this does not work well outside of this context.
    - Now, I have added a validation that assures that "." is not included inside in the `field` prop string. I think it is bad to give users the idea that this notation will work consistently.
    -  I added two new components to give an alternate way to use inputs on the join model without dot.separated.paths
        - `<AutoHasManyThroughJoinModelForm> ` - wrapper  component around regular AutoForm inputs to ensure that they are applied to the join model

### DIFF
--- a/packages/react/.changeset/rotten-news-rule.md
+++ b/packages/react/.changeset/rotten-news-rule.md
@@ -1,0 +1,5 @@
+---
+"@gadgetinc/react": patch
+---
+
+BEAKING CHANGE - Added error for using dot.path.fields in the `field` prop of AutoInputs to traverse record relationships. This functionality was inconsistent and is now better handled by the new relationship form components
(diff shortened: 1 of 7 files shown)


airhorns (approved):
LGTM after changes discussed IRL to remove input component, rename wrapper component to match form nomenclature

PR #801, opened 2025-02-21
Strengthen internal .deleteMany type to require being passed a filter

airhorns:
A user on discord invoked `api.thing.internal.deleteMany(["1", "2", "3"])` which is wrong -- the function accepts an options object with a filter and a sort, but the type had those properties marked as optional, which meant that the array actually passed the typecheck! The right invocation is `api.thing.internal.deleteMany({filter: { ...}})`. Since this is a really sensitive operation, I think we should be more annoying about the required inputs so that users are forced to pass a filter of some sort. They can still delete everything by doing `api.thing.internal.deleteMany({filter: {}})`, but I think we should make that done explicitly instead of by accident if you don't pass the right options.

This is a breaking types change which I am ok to just YOLO -- functionality is unaffected but I think we can consider it a bugfix.

### DIFF
--- a/packages/api-client-core/src/InternalModelManager.ts
+++ b/packages/api-client-core/src/InternalModelManager.ts
@@ -22,7 +22,13 @@ import {
   namespacify,
   sortTypeName,
 } from "./support.js";
-import type { InternalFieldSelection, InternalFindListOptions, InternalFindManyOptions, InternalFindOneOptions } from "./types.js";
+import type {
+  AnyFilter,
+  InternalFieldSelection,
+  InternalFindListOptions,
+  InternalFindManyOptions,
+  InternalFindOneOptions,
+} from "./types.js";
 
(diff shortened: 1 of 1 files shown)


MillanWangGadget (approved):
- 🍏 arrays are now type errors in `api.internal.x.deleteMany`

Changes look good to me 👍

MillanWangGadget:
BTW - The CI cypress fail is a CI only thing. It is good locally and I've been chasing that down for a while. You're good to merge on red

airhorns:
> this will only type error? should we throw a runtime error too?

No I don't think so, because I think deleting everything is actually a valid use case, and I don't super want to change runtime behaviour on people all of a sudden. This just sands down the rough edge to give you a squiggle if you do the wrong thing by accident which I think is sufficient.

infiton (approved):

PR #805, opened 2025-02-24
Background action upsert responses

infiton:
Right now this code breaks in a gadget app:

```
const handles = await api.enqueue(api.todo.bulkUpsert, input)

return await Promise.all(handles.map((handle) => handle.result()))
```

This is because we are not correctly building the graphql query in cases where the `hasReturnType` between an action and its bulkAction are different.

This PR changes this so that if a bulkAction has a reference to its `singleAction` we will use that to generate the query for background action results; things will continue to work for existing clients and non-upsert actions

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/operationBuilders.ts
+++ b/packages/api-client-core/src/operationBuilders.ts
@@ -206,25 +206,32 @@ export const backgroundActionResultOperation = <Action extends AnyActionFunction
   options?: Options
 ) => {
   let fields: FieldSelection = {};
-  let operationName = action.operationName;
   let resultType: string;
 
-  if (action.isBulk) {
-    operationName = action.operationName.replace(/^bulk/, "").replace(/s$/, "");
+  const backgroundAction = action.isBulk && action.singleAction ? action.singleAction : action;
+
+  let operationName = backgroundAction.operationName;
+  if (backgroundAction.isBulk) {
+    operationName = backgroundAction.operationName.replace(/^bulk/, "").replace(/s$/, "");
   }
 
-  if (!action.operationReturnType) {
+  if (!backgroundAction.operationReturnType) {
     resultType = `${camelize(operationName)}Result`;
   } else {
-    resultType = `${action.operationReturnType}Result`;
+    resultType = `${backgroundAction.operationReturnType}Result`;
   }
 
-  switch (action.type) {
+  switch (backgroundAction.type) {
     case "action": {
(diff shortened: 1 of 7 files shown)


packages/api-client-core/src/operationBuilders.ts
```diff
@@ -206,25 +206,32 @@ export const backgroundActionResultOperation = <Action extends AnyActionFunction
   options?: Options
 ) => {
   let fields: FieldSelection = {};
-  let operationName = action.operationName;
   let resultType: string;
 
-  if (action.isBulk) {
-    operationName = action.operationName.replace(/^bulk/, "").replace(/s$/, "");
+  const backgroundAction = action.isBulk && action.singleAction ? action.singleAction : action;
```
infiton, line 211: before we make changes in core, `action.singleAction` will always be `undefined` and this will be equivalent to the current case

packages/api-client-core/src/operationRunners.ts
```diff
@@ -403,7 +403,7 @@ export async function enqueueActionRunner<SchemaT, Action extends AnyActionFunct
   options: EnqueueBackgroundActionOptions<Action> = {}
 ): Promise<Result | Result[]> {
   const normalizedVariableValues = action.isBulk
-    ? disambiguateBulkActionVariables(action as ActionFunctionMetadata<any, any, any, any, any, true>, variables)
+    ? disambiguateBulkActionVariables(action, variables)
```
infiton, line 406: type assertion no longer needed due to change in types made above

packages/react/src/useEnqueue.ts
```diff
@@ -69,9 +64,7 @@ export const useEnqueue = <SchemaT, Action extends AnyActionFunction>(
     state,
     useCallback(
       async (input: Action["variablesType"], options?: EnqueueBackgroundActionOptions<Action>) => {
-        const variables = action.isBulk
-          ? disambiguateBulkActionVariables(action as ActionFunctionMetadata<any, any, any, any, any, true>, input)
-          : disambiguateActionVariables(action, input);
+        const variables = action.isBulk ? disambiguateBulkActionVariables(action, input) : disambiguateActionVariables(action, input);
```
infiton, line 67: type assertion no longer needed due to change in types made above

MillanWangGadget (approved):
Change makes sense to get a real reference to the single action to make the operation. Seems safer than trying to make things work based on the existing `singleActionFunctionName`

PR #807, opened 2025-02-26
export AutoFormProps and AutoTableProps for auto/polaris

infiton:
@pistachiobaby has run into an issue when generating reference cards for AutoForm and AutoTable that the props type for these are not exported; this PR adds these as exports

I also changed the type of AutoForm props that make it clear that for actions that take record identity that either a findBy or record prop is required. I also added some more test coverage for passing a record instead of a findBy

cc @airhorns 

## PR Checklist


packages/react/src/auto/AutoForm.ts
```diff
@@ -68,17 +66,31 @@ export type AutoFormProps<
...
+            findBy?: never;
+            /** A record for this form to act on */
+            record: GadgetRecord<any>;
+          }
     : // eslint-disable-next-line @typescript-eslint/ban-types
       {
         /** This action doesn't run against existing records, so you can't pass a findBy option */
         findBy?: never;
+        /** A record for this form to act on */
+        record?: GadgetRecord<any>;
```
MillanWangGadget, line 93: This comment contradicts the above comment saying
```This action doesn't run against existing records, so you can't pass a findBy option```

I'm realizing now that this record prop clashes with the `defaultValues` prop for prepopulating the form  values. Should we deprecate one of them to avoid confusion? 

Also I think it'd be better to have a action input dependent type passed into the GadgetRecord. I tried doing something like that for defaultValues previously, but it breaks for relationships where the type is based on nested actions like `belongsToField: {create: {...`. I think that would need another API client core change. That could be considered separately tho

**Review by MillanWangGadget — COMMENTED** [review]
- 🍏 Types getting exported for the Form and Table look reasonable
- 🟡 It feels weird that you can pass in `record` and `defaultValues` at the same time given they do similar things in slightly different ways. Perhaps the type should make these mutually exclusive. 

I'm gonna come back to this with fresh eyes to test deeply selected records in deep AutoRelationshipForms.

File: packages/react/spec/auto/polaris/form/PolarisAutoFormExistingRecord.stories.jsx [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+import translations from "@shopify/polaris/locales/en.json";
+import React from "react";
+import { Provider } from "../../../../src/GadgetProvider.tsx";
+import { PolarisAutoForm } from "../../../../src/auto/polaris/PolarisAutoForm.tsx";
+import { useFindOne } from "../../../../src/index.ts";
+import { FormProvider, useForm } from "../../../../src/useActionForm.ts";
+import { testApi as api } from "../../../apis.ts";
+
+const AutoFormWithExistingRecord = () => {
+  const [{ data: widget }] = useFindOne(api.widget, "1145");
```
MillanWangGadget, line 11: Interesting

When this fetch has a `select: {...}` that does not include all of the fields in the form, the unselected fields will get nullified in the submit. 

I think this is correct because the resulting record matches the state of the form. Not selecting in the fetch is basically a `reset field` operation

packages/react/src/auto/polaris/index.ts
```diff
@@ -2,7 +2,9 @@ export { useAutoFormMetadata } from "../AutoFormContext.js";
 export { AutoHasManyThroughJoinModelForm } from "../hooks/useHasManyThroughController.js";
 export { PolarisAutoButton as AutoButton } from "./PolarisAutoButton.js";
 export { PolarisAutoForm as AutoForm, PolarisAutoFormSkeleton as AutoFormSkeleton } from "./PolarisAutoForm.js";
+export type { PolarisAutoFormProps as AutoFormProps } from "./PolarisAutoForm.js";
 export { PolarisAutoTable as AutoTable } from "./PolarisAutoTable.js";
+export type { PolarisAutoTableProps as AutoTableProps } from "./PolarisAutoTable.js";
```
pistachiobaby, line 7: 💯

pistachiobaby (approved):

MillanWangGadget (commented):
- `record` prop
  - 🍏 Passing in a record without `id` selected makes an error
  - 🍏 `defaultValues` wins over `record` for root 
  - 🟡 Passing in a record with hasMany children does not load them in the form
    - I suspect that `...edges.node...` not getting translated properly to the form state is the reason. I think this could be followed up though as unblocking Nick feels more imporatant 


packages/react/src/auto/shadcn/unreleasedIndex.ts
^ in here, it'd be nice to add the type exports aswell for the Shadcn side so we don't gotta remember to do it later

packages/react/spec/auto/polaris/form/PolarisAutoFormExistingRecord.stories.jsx
infiton: yah I think that's the correct behaviour

packages/react/src/auto/AutoForm.ts
```diff
@@ -68,17 +66,30 @@ export type AutoFormProps<
...
+            findBy?: never;
+            /** A record for this form to act on; will be merged with the default values */
+            record: GadgetRecord<any>;
+          }
     : // eslint-disable-next-line @typescript-eslint/ban-types
       {
         /** This action doesn't run against existing records, so you can't pass a findBy option */
         findBy?: never;
+        /** This action doesn't operate with a record, so you can't pass a record option */
+        record?: never;
```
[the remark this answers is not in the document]
infiton: this is true of the findBy as well; taking these out led to a bunch of errors downstream where we assume this key is in the props

packages/react/src/auto/AutoForm.ts
```diff
@@ -616,17 +649,30 @@ const removeIdFieldsUnlessUpsertWithoutFindBy = (isUpsertWithFindBy?: boolean) =
...
 
   if (isUpsertMetaAction) {
     return; // optional findBy value for upsert meta actions
   }
 
-  if (operatesWithRecordId && !hasFindBy) {
-    throw new Error("The 'findBy' prop is required for actions that operate with a record identity.");
-  } else if (!operatesWithRecordId && hasFindBy) {
-    throw new Error("The 'findBy' prop is only allowed for actions that operate with a record identity.");
+  if (record) {
```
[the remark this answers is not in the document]
infiton: :+1: fixed and added a test for this

packages/react/src/auto/AutoForm.ts
```diff
@@ -616,17 +649,30 @@ const removeIdFieldsUnlessUpsertWithoutFindBy = (isUpsertWithFindBy?: boolean) =
...
+    if (operatesWithRecordId && !record.id) {
+      throw new Error("Passing a record to an action that operates with a record identity requires the record to have an id.");
+    } else if (!operatesWithRecordId && record.id) {
```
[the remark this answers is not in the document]
infiton: 👍

MillanWangGadget:
- 🍏 Error for using findBy and record at the same time
- 🍏 Error for passing in `record` to create and global actions
- 🍏 When `defaultValues` and `record` are used at the same time, `defaultValues` will win
- 🍏 Record passed in with selected relationships will populate the relationship inputs. They submit properly
- ❔ Interestingly, doing `findBy={diffRecordId} defaultValues={{modelName: record}}` kinda works. Simple field types from the passed in record work, but hasMany relationships seem to get overwritten by the findBy. 
- 🟡 The id in a record does not prepopulate the `AutoIdInput` in upsert actions. This works for `defaultValues` though with something like `modelName.id: "1"`
- 🟡 In the new storybook, I'm unable to submit due the the roles field. We send it as an array of strings, but the passed in record puts them in as objects.

PR #810, opened 2025-02-26
Bring back shadcn to react pack export

MillanWangGadget:
Simply re-exporting the Shadcn auto components so that they are visible in the package


packages/react-bigcommerce/package.json
```diff
@@ -41,6 +41,6 @@
   "peerDependencies": {
```
infiton, line 41: do we need to bump the versions of these packages too?

MillanWangGadget: By that do you mean react 18->19 across the whole repo?

airhorns: No I think he means the version of these packages themselves so they release new versions also that then depend on 20, like react-bigcommerce needs to go to v0.3.0

airhorns (approved):
WOOO!

packages/react-bigcommerce/package.json
MillanWangGadget: Aha that makes sense.

Updated to boost version of all package.json files that have `@gadgetinc/react` as a peer dep

PR #819, opened 2025-03-11
Reworked Shadcn AutoComponents DateTime picker

MillanWangGadget:
CleanShot 2025-03-11 at 17 40 18

- **UPDATE**
  - Changed the default `Calendar` component file to be based on the more recent version of `react-day-picker`. 
    - Shadcn uses an outdated version of this component in their docs and this is called out by the `react-day-picker` docs that they point out. The version provided by the Shadcn docs did not have year selection, and it also had a lot of style issues that would need a lot of advanced styling out of the box to look reasonable. 
    - https://date-picker.luca-felix.com/
  - Made the state of the date and time pickers completely dependent on the form state
    - There were previously some `useStates` that were used to track the time and date independently of the form state. Those had the possibility of getting out of sync
  - Replaced the previous `input with time validation` approach with a `input type="time"` approach
    - This input won't even let you enter bad time values
    - This has a consequence of making the clock 24h instead of 12h
    - This avoids a lot of oddities that occurred with the validation when bad times were entered. It's also a bit of friction for the user  to clear out a bad value
  - Added a `clear` button to the input
    - Previously, there was no way to clear the value
    - There still is no way to clear the value on the polaris side. It is a bigger design challenge over there

### DIFF
--- a/packages/react/spec/auto/shadcn-defaults/components/Calendar.tsx
+++ b/packages/react/spec/auto/shadcn-defaults/components/Calendar.tsx
@@ -1,50 +1,408 @@
+/**
+ * From https://date-picker.luca-felix.com/
+ * The default component provided by react-day-picker
+ * They mention that the default component in the Shadcn docs is outdated and suggests this as an alternative
+ *
+ */
+
+import { differenceInCalendarDays } from "date-fns";
 import { ChevronLeft, ChevronRight } from "lucide-react";
 import * as React from "react";
-import { DayPicker } from "react-day-picker";
+import { DayPicker, labelNext, labelPrevious, useDayPicker, type DayPickerProps } from "react-day-picker";
 import { cn } from "../utils.js";
-import { buttonVariants } from "./Button.js";
+import { Button, buttonVariants } from "./Button.js";
 
-export type CalendarProps = React.ComponentProps<typeof DayPicker>;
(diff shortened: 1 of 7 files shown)


packages/react/spec/auto/shadcn-defaults/components/Calendar.tsx
```diff
@@ -1,50 +1,409 @@
...
+            disabled={navView === "years" ? isDisabled : undefined}
+          >
+            {displayYears.from + i}
+          </Button>
+        );
+      })}
+    </div>
+  );
+}
+
```
airhorns, line 408: when we go to integrate this new version into gadget we may need to refresh our component templates to drop this new one, FYI there is a `development/refresh-shadcn.mjs` script for that

packages/react/src/auto/shadcn/inputs/ShadcnAutoDateTimePicker.tsx
```diff
@@ -82,47 +91,27 @@ export const makeShadcnAutoDateTimePicker = ({
       [fieldProps, value, localTz, onChange]
     );
 
-    const [date, setDate] = useState<Date>();
```
airhorns, line 85: 🙏 good call

airhorns (approved):

PR #822, opened 2025-03-13
make shadcn dateTime picker use Lazyload to avoid SSR issues with css files

MillanWangGadget:
- Update
  - `import "./something.css"` does not play nice with SSR in Gadget apps. 
  - This was affecting the DateTime picker because of the required css file to make the `input type="time"` dropdown not appear
  - The part that requires css has now been moved to a new file that will be lazy loaded. This was inspired from the `AutoRichTextInput`  system of getting in the custom css file


airhorns (approved):
Makes sense to me, and is also annoying

PR #823, opened 2025-03-13
Add computed view hook

airhorns:
This adds a react `useView` hook to accompany the imperative computed view functions we're already generating in fv1.4 . `useView` works pretty similarly to a finder in that we grab the function off the api client and run half of it to get the operation plan, and then execute the actual thing using the urql hook. Computed views have no built in support for pagination so its simpler than a findMany or what have you, but they are typed so we do the ole fish-out-the-type-from-a-fake-property-on-the-function-thing.

The plan for releasing this is to stick it in the next react version but not announce support for a while yet while we dogfood and do an internal alpha methinks.

[no-changelog-required]

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@gadget-client/app-with-no-user-model": "^1.10.0",
     "@gadget-client/bulk-actions-test": "^1.113.0",
     "@gadget-client/full-auth": "^1.9.0",
-    "@gadget-client/js-clients-test": "1.511.0-development.2409",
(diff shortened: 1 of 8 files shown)


MillanWangGadget (approved):
- 🍏 Everything in the new hook seems to have similar vibes to things that already exist in other hooks like `useFindMany`

I think this is good to stealth ship now to get started on some real testing. 

FYI, i got another small change I want to sneak into the next release as well

infiton (approved):

PR #825, opened 2025-03-13
Add a ping to live queries to ensure socket is active

garymardell:
Without the `keepAlive` option the websocket used for live queries does not send out ping messages. This is a problem as after the initial connection and subscription messages no more messages are sent to the server. This means that the connection could no longer be active without realising it.

Following the example in:

[ttps://github.com/enisdenjo/graphql-ws/blob/5161bd91388677893badb8dbaca5c6587095199b/src/client.ts#L266-L303](https://github.com/enisdenjo/graphql-ws/blob/5161bd91388677893badb8dbaca5c6587095199b/src/client.ts#L266-L303)

I have set the `keepAlive` parameter with a 5 second interval and waiting at most 2 seconds to get a response from the server. This is slightly lower than the example which uses 10 second intervals, and a 5 second timeout as I expect live queries to be more real time than this. 

## Testing

I have tested this locally by:

* Make a page use a live query and print results
* Use the network developer tools to set to offline
* Make an update to some data displayed by the live query
* Then disable the offline mode again

Previously the websocket would continue to appear active however receive no further messages. With this change you see a new websocket connection is established after the timeout.

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -468,9 +468,13 @@ export class GadgetConnection {
       url = addUrlParams(url, overrides.urlParams);
     }
 
+    let activeSocket: globalThis.WebSocket;
+    let timedOut: NodeJS.Timeout;
+
     return this.createSubscriptionClient({
       url,
       webSocketImpl: this.websocketImplementation,
+      keepAlive: 5_000,
       connectionParams: async () => {
         // In the browser, we can't set arbitrary headers on the websocket request, so we don't use the same auth mechanism that we use for normal HTTP requests. Instead we use graphql-ws' connectionParams to send the auth information in the connection setup message to the server.
         const connectionParams: Record<string, any> = { environment: this.environment, auth: { type: this.authenticationMode } };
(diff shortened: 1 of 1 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -511,6 +515,19 @@ export class GadgetConnection {
...
           overrides?.on?.connected?.(socket, payload);
+          activeSocket = socket as globalThis.WebSocket;
+        },
+        ping: (received) => {
+          if (!received) {
+            timedOut = setTimeout(() => {
+              if (activeSocket.readyState === WebSocket.OPEN) {
+                activeSocket.close(4408, "Request Timeout");
+              }
+            }, 2_000);
```
airhorns, line 526: I think this might be a bit aggressive for people on planes or trains or what have you with low latency links. Maybe 7s / 3s or something like that? Very finger to the wind

airhorns (approved):
LGTM after comments

packages/api-client-core/src/GadgetConnection.ts
garymardell: Happy to bump up the limits a bit and we can adjust based on future feedback.

PR #826, opened 2025-03-14
Shadcn autocomponent QA 2

MillanWangGadget:
- UPDATE
  - Revamped autoTable to match designs. Now have sticky headers and 1st columns
  - Updated submission result banners to be better styled 
  - Updated dropdown in Shadcn combo input to be like a popover instead of rendered in-line in the page, and also updated the `load more on scroll to bottom` behaviour to work with this new system


packages/react/src/auto/shadcn/ShadcnAutoTable.tsx
```diff
@@ -105,44 +133,72 @@ export const makeAutoTable = (elements: ShadcnElements) => {
...
-          ))}
+          {columns.map((column, i) => {
+            const isSticky = i === 0;
+            return (
+              <TableCell
+                key={column.identifier}
+                style={column.style}
+                className={`${
+                  isSticky ? `sticky left-${canSelectRecords ? "6" : "0"} z-10` : ``
+                } max-w-[200px] overflow-hidden text-ellipsis whitespace-nowrap ${getCellBackgroundColor(isSticky, isSelected, isHovered)}`}
```
airhorns, line 188: Tailwind ninja!

airhorns (approved):
LGTM!

PR #828, opened 2025-03-18
Added dark/light mode selector and design system selectors to all most storybooks

MillanWangGadget:
- **UPDATE**
  - Many of the issues I've found with the Shadcn auto comps have been from `light mode good, dark mode bad` issues. Since we default RRv7 apps to be created in dark mode and storybook was previously only in light mode, these issues were easy to miss, and visible to many 
  - Now, I've added a dark/light mode selector to most of the Shadcn stories so it's easier to find these `light mode good, dark mode bad` moments 
  - AutoForm stories have also been updated to have a built in design mode selector for Polaris/Shadcn. I had done this for AutoTable and found that is was really nice to helping to check feature parity
  - The Tailwind config has been updated to be basically the same as a real created RRv7 app. There were other issues that came from subtle differences between the two


MillanWangGadget:
Talked offline about releasing in light mode only for now. I'm gonna merge as the `polaris / Shadcn` switcher is gonna be really useful for getting all of the shadcn relationship forms working well

PR #829, opened 2025-03-19
Update Shad autocomponents relationship forms & other QA

MillanWangGadget:
- UPDATES
  - Removed dark mode toggle for auto comps in storybook. It was decided that this will come later as full dark mode support will be a big mission 
  - Updated all relationship forms to be more visually similar to the Polaris equivalents. This replaces the base shadcn accordion stuff that was used previously 
  - Replaced AutoTable neutral/hover/selected color overlay system with a brightness control. 
    - This was originally done to avoid partial opacity sticky column issues, but did not look consistent enough
  - Updated FileInput component to be similarly sized to all other inputs and to have a border when a file is selected
  - Added an exported list of all tailwind classes used in the shadcn side. This is to be used as a safelist on the Gadget side to ensure that the necessary classes are loaded so that the auto comps look reasonable


packages/react/spec/auto/storybook/form/AutoRelationshipForm.stories.tsx
```diff
@@ -53,16 +53,18 @@ const ExampleWidgetAutoRelatedForm = (props: any) => {
...
+              <AutoInput field="active" />
+              <AutoInput field="size" />
+            </div>
           </AutoHasOneForm>
         </Card>
 
         <Card>
           <AutoHasManyForm
-            label={<p>{`Has Many Form -- Gizmos`}</p>}
+            // label={`Has Many Form -- Gizmos`}
```
airhorns, line 67: 🔥 ?

packages/react/src/auto/shadcn/GadgetShadcnTailwindSafelist.ts
```diff
@@ -98,7 +114,6 @@ export const GadgetShadcnTailwindSafelist = [
   "whitespace-nowrap",
   "z-0",
   "z-10",
-  "z-100",
   "z-20",
   "z-30",
   "z-50",
```
airhorns, line 119: not for right now, but someday it'd be nice to have tailwind generate this for us, or a test that ensures its accurate, I imagine folks are gonna forget to update it when they use other classes

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoHasManyForm.tsx
```diff
@@ -109,42 +109,53 @@ export const makeShadcnAutoHasManyForm = ({
...
     );
   }
 
+  function AccordionSection(props: { position?: "top" | "bottom" | "middle" | "only"; children: React.ReactNode; onClick?: () => void }) {
+    const { position, children, onClick } = props;
+
+    return (
+      <Button
+        type="button"
+        variant="outline"
```
airhorns, line 141: Do you think folks might add intense styles to their outline buttons in their own shadcn components? And if so, do we want them applied here, or should we use our own thingy that is clickable but isn't a button you think? I could go either way

airhorns (approved):

packages/react/src/auto/shadcn/inputs/relationships/ShadcnAutoHasManyForm.tsx
MillanWangGadget: I feel like I'd rather have this be a button so that the user has some control over it, although I agree that they may not want to have their button styles make changes everywhere. I think I'd like to follow up with adding a prop to the button that lets the user do different styling depending on where/how the button is used. I'm thinking of calling it something like `usageContext` and making it an enum

packages/react/spec/auto/storybook/form/AutoRelationshipForm.stories.tsx
MillanWangGadget: removed 🫡

PR #830, opened 2025-03-19
Remove PolarisAutoTable `F` keyboard shortcut

MillanWangGadget:
This `F` keyboard shortcut applies to the whole page and would hijack focus, even in other textboxes. I've simply removed this keyboard shortcut as always clicking to search seems a lot less bad than focus highjacking


MillanWangGadget:
cc @pistachiobaby

PR #833, opened 2025-03-19
updated shadcn safelist and added test ot ensure added classes stay in sync

MillanWangGadget:
- **UPDATE**
  - Having the safelist potentially get out of sync with the added classes is a unfortunate possibility that the added test aims to prevent
  - The added test relies on the tailwind css tool that creates all of the css classes and keeps them in a file. That result will be used to power the safelist
  - Why are there still special cases? 
    - It seems that the tailwind cli tool misses some classes for reasons that I can't explain. For those that I've identified, I've added a special case class string array


airhorns (approved):
Damn, megadope!

PR #834, opened 2025-03-22
Remove onChange and value props from AutoComps

MillanWangGadget:
- UPDATE
  - `value` and `onChange` props in `Auto_FieldType_Input` components were exposed and could lead to bad bugs where the inputs look connected to the overall form state, but actually do nothing
  - Simplified `PolarisAutoDateTimePicker` now that `value` and `onChange` will come only from the form state. Way simpler now that those extra props do not need to be accounted for 
  - Added new optional `afterChange` prop to many field type specific `AutoInput` components to run a callback after the input value is changed


packages/react/src/auto/polaris/inputs/PolarisAutoDateTimePicker.tsx
```diff
@@ -124,17 +121,14 @@ export const PolarisAutoDateTimePicker = autoInput((props: PolarisAutoDateTimePi
           />
         </div>
       </Popover>
-      {(props.includeTime ?? (config as GadgetDateTimeConfig).includeTime) && (
+      {(config as GadgetDateTimeConfig).includeTime && (
         <div style={{ width: "130px" }}>
           <PolarisAutoTimePicker
             fieldProps={fieldProps}
             id={props.id}
-            localTime={localTime}
```
airhorns, line 132: Didn't need this one?

packages/react/src/auto/polaris/inputs/PolarisAutoNumberInput.tsx
```diff
@@ -21,8 +21,7 @@ type PolarisAutoNumberInputProps = AutoNumberInputProps & Partial<TextFieldProps
  * @returns The AutoNumberInput component
  */
 export const PolarisAutoNumberInput = autoInput((props: PolarisAutoNumberInputProps) => {
-  const { field, control } = props;
-  const { type, metadata, value } = useStringInputController({ field, control });
+  const { type, metadata, value } = useStringInputController(props);
```
airhorns, line 24: 🙏

packages/react/src/auto/shadcn/inputs/ShadcnAutoJSONInput.tsx
```diff
@@ -23,12 +26,12 @@ export const makeShadcnAutoJSONInput = ({ Label, Textarea }: Pick<ShadcnElements
         <Textarea
           {...getPropsWithoutRef(controller)}
           {...getPropsWithoutRef(focusProps)}
+          id={id}
+          className="font-mono"
```
airhorns, line 30: 👍

airhorns (approved):

packages/react/src/auto/polaris/inputs/PolarisAutoDateTimePicker.tsx
MillanWangGadget: `localTime` is just a safely memoized copy of `fieldProps.value`. This is done in `packages/react/src/auto/hooks/useDateTimeField.ts`

PR #844, opened 2025-04-03
Release new react pack 0.20.1

MillanWangGadget:
See changelog


packages/shopify-extensions/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/shopify-extensions",
-  "version": "0.5.0",
+  "version": "0.6.0",
```
airhorns, line 3: I think this could be a minor bump no? Same with the others that arent react itself

MillanWangGadget: I've updated such that the non-react package.json files get their last number updated instead of the middle number. 

Although if the last number is the minor version and the middle number is the major version, what is the first number ?

airhorns (approved):

PR #845, opened 2025-04-03
Add support for inline views and fix operation name on model namespaced views

airhorns:
This adds support for inline views (views that are plain strings instead of .gelly files on the backed) and fixes support for named views mounted on models in the `useView` hook. See the tests!

[no-changelog-required]

### DIFF
--- a/packages/react/src/useView.ts
+++ b/packages/react/src/useView.ts
@@ -110,7 +110,18 @@ export function useView<VariablesT, F extends ViewFunction<VariablesT, any>>(
   }
 
   const memoizedVariables = useStructuralMemo(variables);
-  const memoizedOptions = useStructuralMemo(options);
+  const memoizedOptions = useStructuralMemo({
+    ...options,
+    context: {
+      ...options?.context,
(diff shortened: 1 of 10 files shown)


packages/react/src/useView.ts
```diff
@@ -75,17 +111,26 @@ export function useView<VariablesT, F extends ViewFunction<VariablesT, any>>(
...
     return { ...rawResult, data, error };
-  }, [view, options?.pause, rawResult]);
+  }, [dataPath, options?.pause, rawResult]);
 
   return [result, refresh];
 }
+
+const inlineViewQuery = `query InlineView($query: String!, $variables: JSONObject) { 
+  gellyView(query: $query, variables: $variables) 
+}`;
```
airhorns, line 136: this is a hardcoded query we make available in the GraphQLSchemaGenerator for running inline views, it is weird lookin but users never have to see it

infiton: :+1:

infiton (approved):

PR #850, opened 2025-04-07
Fix pathing of validations on AutoHasManyThroughForm

MillanWangGadget:
- **UPDATE**
  - `AutoHasManyThroughForm` had a bug where the fields within were registered on paths that were different from their paths on the real form state
    - Fetching and submitting were always pathed based on the implicit join model hasMany field
      - Eg. Course hasMany students through registration - course.registrations.student.name, course.registration.startDate
    - The resulting nested action was also pathed based on this `implicit join model hasMany field` system.
    - The problem was that the metadata pathing was based on the actual HMT field instead of the implicit joinModelHasMany field. This mismatch of paths caused the following issues 
      - The path of the yupResolver did not agree with the form state path for AutoInputs within AutoHasManyThroughForm.
        - Validations were not run on those AutoInput fields, and bad validations did not block submission like other form contexts
        - The resolver is responsible for type transformations, and it would be impossible to user a number field in a AutoHasManyThrough field because the form state holds numbers a strings, and we relied on this resolver to transform them. This resulted in errors like `"123" is not assignable to type number`
        - `defaultValues` did not work for AutoHasManyThroughForm regardless of if the values were pathed based on the real HMT path or the implicitJoinHasMany path. Now the implicitJoinHasMany path works to set default values

### DIFF
--- a/packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
+++ b/packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
@@ -76,6 +76,7 @@ describeForEachAutoAdapter(
             <AutoHasManyThroughJoinModelForm>
               <AutoInput field="effectiveFrom" />
               <AutoInput field="effectiveTo" />
+              <AutoInput field="attempt" />
             </AutoHasManyThroughJoinModelForm>
           </AutoHasManyThroughForm>
           <AutoSubmit id="submit" />
@@ -91,7 +92,9 @@ describeForEachAutoAdapter(
       cy.contains("Emma Williams").click();
       cy.contains("Add Students").click();
 
+      cy.wait(1000);
       cy.get('[id="deleteButton_students.0"]').click();
+      cy.wait(1000);
 
       expectUpdateActionSubmissionVariables({
         course: {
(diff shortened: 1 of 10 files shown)


packages/react/spec/auto/storybook/form/AutoForm.stories.tsx
```diff
@@ -219,12 +213,5 @@ export const hasManyThrough: any = {
   name: "HasManyThrough fields",
   args: {
     action: api.hasManyThrough.baseModel.create,
-    children: (
-      <>
-        <AutoInput field="baseModelName" />
-        <AutoHasManyThroughInput field="baseModelHmtField" />
-        <AutoSubmit />
-      </>
-    ),
```
airhorns, line 228: why this change?

packages/react/spec/auto/storybook/form/AutoRelationshipForm.stories.tsx
```diff
@@ -213,16 +213,15 @@ const ExampleCourseCreateRelatedForm = (props: any) => {
...
               <AutoHasManyThroughJoinModelForm>
                 <AutoInput field="effectiveTo" />
                 <AutoInput field="effectiveFrom" />
+                <AutoInput field="attempt" />
               </AutoHasManyThroughJoinModelForm>
 
-              {/* Fields on the sibling model. No prefix */}
               <AutoInput field="firstName" />
               <AutoInput field="lastName" />
+              <AutoInput field="year" />
```
airhorns, line 224: Perhaps we should make these two cases two different tests so we know they aren't relying on each other to work?

airhorns (commented):
Makes sense to me but I think this is so intense that it definitely needs some tests!

packages/react/spec/auto/storybook/form/AutoForm.stories.tsx
MillanWangGadget: Originally, this storybook case was made before we made `AutoInput` work with HMT, so the only way to make it appear was with the FieldType specific input. Now that `AutoInput` works with HMT, The one-liner works to make it appear.

There is also a secondary benefit of proving that the implicit `joinModelHasMany` field does not appear as an individual input in the one-liner autoForm. That was a previous bug that I wanted to ensure did not regress

packages/react/cypress/component/auto/form/AutoHasManyThroughForm.cy.tsx
```diff
@@ -101,14 +101,20 @@ describeForEachAutoAdapter(
...
                 id: "50",
                 student: { update: { firstName: "Benjamin", id: "43", lastName: "Martin" } },
               },
             },
             // Created third
-            { create: { student: { _link: "10" } } },
+            {
+              create: {
+                effectiveFrom: null,
+                effectiveTo: null,
```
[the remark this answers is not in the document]
MillanWangGadget: Yes - These nulls indicate the those inputs were actually attached to the form properly even when the inputs themselves were untouched

packages/react/spec/auto/storybook/form/AutoRelationshipForm.stories.tsx
MillanWangGadget: This was a really good call. I ended up finding out that there also needed to be some work done on the display fields in the HMT forms to get them into the selection based on their full paths. I've fixed and pushed that now

MillanWangGadget:
> Makes sense to me but I think this is so intense that it definitely needs some tests!

Updated the tests so that they are now involving  number fields and validations with proper typed expects. I gotta run now  though so I can't come back for further feedback

airhorns:
Ok, we'll pick it up! Thanks Millan!

airhorns (approved):

PR #864, opened 2025-06-03
Add back in support for reexecution

garymardell:
Removes the override for `reexecuteOperation` so subscriptions can be re-established when inputs change. 

Related gadget PR: https://github.com/gadget-inc/gadget/pull/16380

## PR Checklist

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -1,12 +1,13 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import type { DefinitionNode, DirectiveNode, OperationDefinitionNode } from "@0no-co/graphql.web";
-import type { ClientOptions, RequestPolicy } from "@urql/core";
+import type { ClientOptions, Exchange, RequestPolicy } from "@urql/core";
 import { Client, cacheExchange, fetchExchange, subscriptionExchange } from "@urql/core";
 import type { ExecutionResult } from "graphql";
 import type { Sink, Client as SubscriptionClient, ClientOptions as SubscriptionClientOptions } from "graphql-ws";
 import { CloseCode, createClient as createSubscriptionClient } from "graphql-ws";
 import type { Maybe } from "graphql/jsutils/Maybe.js";
 import WebSocket from "isomorphic-ws";
+import { filter, merge, pipe, tap } from "wonka";
 import type { AuthenticationModeOptions, BrowserSessionAuthenticationModeOptions, Exchanges } from "./ClientOptions.js";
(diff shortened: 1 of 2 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -77,6 +78,32 @@ export enum AuthenticationMode {
 
 const objectForGlobals = typeof globalThis != "undefined" ? globalThis : typeof window != "undefined" ? window : undefined;
 
+const liveQueryExchange: Exchange = ({ forward }) => {
+  const executed = new Set<number>();
```
airhorns, line 82: `WeakSet` maybe? Otherwise this will leak if we have any bugs that make things never get torn down

packages/react/spec/liveQueries.spec.tsx
```diff
@@ -227,51 +222,4 @@ describe("live queries", () => {
...
-
-    // re-execute the live query operation
-    urqlClient.reexecuteOperation(findManyOperation!);
-
-    // let the event loop run to give urql a chance to re-execute the operation
-    await sleep(1000);
-
-    // make sure the operation was NOT re-executed
-    expect(operationCount).toBe(1);
-  });
```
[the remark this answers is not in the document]
garymardell: For sure, I can try to write a test. Just wanted some eyes on the approach as it seemed slightly convoluted before committing the time to writing one.

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -368,6 +395,7 @@ export class GadgetConnection {
     // apply urql's default caching behaviour when client side (but skip it server side)
     if (typeof window != "undefined") {
       exchanges.push(cacheExchange);
+      exchanges.push(liveQueryExchange);
```
scott-rc, line 398: We don't want this exchange on the server too?

packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -77,6 +78,32 @@ export enum AuthenticationMode {
 
 const objectForGlobals = typeof globalThis != "undefined" ? globalThis : typeof window != "undefined" ? window : undefined;
 
+const liveQueryExchange: Exchange = ({ forward }) => {
```
scott-rc, line 81: nit: can we move this into https://github.com/gadget-inc/js-clients/tree/connection-reexecute/packages/api-client-core/src/exchanges and add a comment explaining why it exists?

scott-rc (approved):
Sorry for the delay! This makes sense to me 👍

airhorns:
Superseded by #867

PR #868, opened 2025-06-22
Updated AutoTable to allow custom column header components

MillanWangGadget:
- UPDATE
  - A discord user asked if they could add custom react content to the column headers in AutoTable. We did not allow this in the types to follow the Polaris precedent, but since it works, I've updated the types to allow for it 

Example - Dynamically changing color text 
CleanShot 2025-06-22 at 16 56 28

### DIFF
--- a/packages/react/spec/auto/storybook/table/AutoTable.stories.tsx
+++ b/packages/react/spec/auto/storybook/table/AutoTable.stories.tsx
@@ -169,17 +169,31 @@ const CustomCheckboxCell = ({ record }: { record: any }) => {
   );
 };
 
+const ColorChangingText = (props: { text: string }) => {
+  const [color, setColor] = React.useState("#000000");
+
+  React.useEffect(() => {
+    const interval = setInterval(() => {
+      const randomColor = "#" + Math.floor(Math.random() * 16777215).toString(16);
+      setColor(randomColor);
+    }, 1000);
+
+    return () => clearInterval(interval);
+  }, []);
+  return <div style={{ color }}>{props.text}</div>;
+};
+
 export const CustomCellWithUseAction: any = {
   args: {
     model: api.autoTableTest,
     columns: [
       "bool",
       {
-        header: "Pass the whole function",
+        header: <ColorChangingText text="Pass the whole function" />,
         render: CustomCheckboxCell,
       },
       {
-        header: "JSX style",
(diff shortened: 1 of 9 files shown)


packages/react/spec/auto/storybook/table/AutoTable.stories.tsx
```diff
@@ -169,19 +169,51 @@ const CustomCheckboxCell = ({ record }: { record: any }) => {
...
-        header: "JSX style",
+        header: <ColorChangingText text="JSX style" />,
         render: (props: any) => <CustomCheckboxCell {...props} />,
       },
+
+      //
+      //
+      //
+      //
+      //
```
infiton, line 204: what are these for?

packages/react/src/use-table/helpers.tsx
```diff
@@ -156,7 +156,7 @@ export const getTableColumns = (spec: Pick<TableSpec, "fieldMetadataTree" | "tar
         identifier,
         render: targetColumn.render,
         header: targetColumn.header,
-        field: targetColumn.header,
+        // field: targetColumn.header,
```
infiton, line 159: 🔥 ?

packages/react/spec/auto/storybook/table/AutoTable.stories.tsx
MillanWangGadget: oops, I use these kinda like bookmarks so it's easier to see what areas I'm working on as I scroll through a file. They have been removed

packages/react/src/use-table/types.ts
```diff
@@ -29,20 +29,26 @@ export type TableColumn = {
...
-  field: string;
-  type: ColumnType;
+  header: ReactNode;
   /** parent relationship type if the parent field is a relationship */
   relationshipType?: RelationshipType;
   sortable: boolean;
   /** For controlling if the time is shown on DateTime cell renderers   */
   includeTime?: boolean;
-  /** Custom render function */
-  render?: CustomCellRenderer;
```
alanko0511, line 42: Just curious, whahappensen right now if I:
A. The `type` is `GadgetFieldType` and the `render` property exists, or...
B. The `type` is `CustomRenderer` and the `field` property exists

alanko0511 (approved):

infiton (approved):

PR #873, opened 2025-07-07
Introduce new public auth mode for single shop tenant

alanko0511:
This PR introduces a new authentication mode, `publicShopTenant`, that allows apps like Shopify theme app extensions to set up an unauthenticated role with a Shopify tenancy.

This allows use cases like fetching publicly accessible data for the specific shop without manually passing in the shop ID.

<details><summary>Click here to see an example usage</summary>

Example usage:
```liquid
<!-- In extensions/theme-extension/blocks/star_rating.liquid -->
<script src="https://shop-auth-mode--development.ggt.pub/api/client/web.min.js" defer="defer"></script>
<script>
document.addEventListener("DOMContentLoaded", function () {
  window.gadgetPublicClient = new Gadget({
    authenticationMode: {
      publicShopTenant: {
        shopId: {{ shop.id }},
      },
    },
  });
})
</script>
<script src="{{ 'getShopInfo.js' | asset_url }}" defer="defer"></script>
```

```js
// In extensions/theme-extension/assets/getShopInfo.js
const fetchCurrentShop = async () => {
  const shopInfoComponent = document.getElementById("shop-info");
  const shopInfo = await window.gadgetPublicClient.getCurrentShop();
  if (shopInfoComponent) {
    shopInfoComponent.innerHTML = JSON.stringify(shopInfo, null, 2);
  }
};

document.addEventListener("DOMContentLoaded", function() {
  fetchCurrentShop();
});
```

Then the `shopInfoComponent` should render the payload from the `getCurrentShop` global action.

</details> 

## PR Checklist

- [ ] Any user facing changes are documented (changelog section below)
- [ ] Any app Problems are added to the problem finders
- [ ] Any app data that needs a redeploy to take effect is listed in the `contentHash` functions
- [ ] Important or complicated production behaviour has traces and logs added

## Changelog

If this PR will affect change user-facing behaviour, add detailed information about the change here to prompt an AI generated changelog entry.

If no changelog is required, write `no-changelog-required` in square brackets `[]`.

### DIFF
--- a/packages/api-client-core/src/GadgetConnection.ts
+++ b/packages/api-client-core/src/GadgetConnection.ts
@@ -74,6 +74,7 @@ export enum AuthenticationMode {
   InternalAuthToken = "internal-auth-token",
   Anonymous = "anonymous",
   Custom = "custom",
+  PublicShopTenant = "public-shop-tenant",
 }
 
 const objectForGlobals = typeof globalThis != "undefined" ? globalThis : typeof window != "undefined" ? window : undefined;
@@ -173,6 +174,8 @@ export class GadgetConnection {
         this.authenticationMode = AuthenticationMode.InternalAuthToken;
       } else if (options.apiKey) {
         this.authenticationMode = AuthenticationMode.APIKey;
+      } else if (options.publicShopTenant) {
+        this.authenticationMode = AuthenticationMode.PublicShopTenant;
       } else if (options.custom) {
         this.authenticationMode = AuthenticationMode.Custom;
       }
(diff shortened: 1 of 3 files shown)


packages/api-client-core/src/GadgetConnection.ts
```diff
@@ -74,6 +74,7 @@ export enum AuthenticationMode {
   InternalAuthToken = "internal-auth-token",
   Anonymous = "anonymous",
   Custom = "custom",
+  PublicShopTenant = "public-shop-tenant",
```
infiton, line 77: let's make sure we agree on best name for this

PR #874, opened 2025-07-09
Fixup combined error

udnay:
Small change to how we format a combined error to iterate through the GrahpQL errors and append the network error.

## Changelog

[no-changelog-required]

### DIFF
--- a/packages/api-client-core/spec/support.spec.ts
+++ b/packages/api-client-core/spec/support.spec.ts
@@ -51,6 +51,22 @@ describe("support utilities", () => {
       ).toThrowErrorMatchingInlineSnapshot(`"[Network] foobar"`);
     });
 
+    test("throws the operation error if with the full network error as a string if there is no error message", () => {
+      expect(() =>
+        assertOperationSuccess(
+          {
+            operation: null as any,
+            data: null,
+            error: new CombinedError({ networkError: new Error() }),
+            stale: false,
+            hasNext: false,
+          },
+
+          ["foo", "bar"]
+        )
+      ).toThrow(`[Network] No message, error:`);
+    });
+
     test("throws an actual error object and not a string so that the user gets a stack message", () => {
       try {
         assertOperationSuccess(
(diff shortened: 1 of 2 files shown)


packages/api-client-core/src/support.ts
```diff
@@ -316,8 +316,12 @@ export const getNonNullableError = (response: Result & { fetching: boolean }, da
 
 export const assertOperationSuccess = (response: OperationResult<any>, dataPath: string[], throwOnEmptyData = false) => {
   if (response.error) {
-    if (response.error instanceof CombinedError && (response.error.networkError as any as Error[])?.length) {
-      response.error.message = (response.error.networkError as any as Error[]).map((error) => "[Network] " + error.message).join("\n");
+    if (response.error instanceof CombinedError && response.error.networkError) {
+      if (response.error.networkError?.message) {
+        response.error.message = `[Network] ${response.error.networkError.message}`;
+      } else {
+        response.error.message = `[Network] No message, stack: ${String(response.error.networkError.stack)}`;
```
airhorns, line 323: I think we should stringify the whole error error here instead of relying on just the stack. In JS you can throw anything, like `throw "foo"`, so the error may not have a message or a stack and just be a string. Since we dont exactly know what it is I think it might be wise to be defensive and just stringify the result right out

packages/api-client-core/spec/support.spec.ts
```diff
@@ -51,41 +51,38 @@ describe("support utilities", () => {
...
 
-    test("throws the operation error if there's multiple network errors on the operation", () => {
-      expect(() =>
+    test("throws an actual error object and not a string so that the user gets a stack message", () => {
+      try {
         assertOperationSuccess(
           {
             operation: null as any,
             data: null,
-            error: new CombinedError({ networkError: [new Error("foo"), new Error("foo")] as any }),
```
airhorns, line 78: hm this almost seems on purpose

packages/api-client-core/src/support.ts
udnay: Ah kk, my test was just returning `Error` with no stack when I did `string(response.networkError)` which is why I explicitly grabbed the stack. Would something like
`[Network] No message, error: string(response.error.networkError) \nstack: ${String(response.error.networkError.stack)}` work?

airhorns: yep

udnay:
Updated, good to merge on 🍏 ?

PR #875, opened 2025-07-09
Introduce `shopId` in `browserSession` to set shop tenant

alanko0511:
This PR introduces a new property in `browserSession`, `shopId`, that allows apps like Shopify theme app extensions to set up a session with a Shopify tenancy.

This allows use cases like fetching publicly accessible data for the specific shop without manually passing in the shop ID.

This PR also allows setting authentication modes at the top level, i.e. no need to declare the modes under the `authenticationMode`.

<details><summary>Click here to see an example usage</summary>

Example usage:
```liquid
<!-- In extensions/theme-extension/blocks/star_rating.liquid -->
<script src="https://shop-auth-mode--development.ggt.pub/api/client/web.min.js" defer="defer"></script>
<script>
document.addEventListener("DOMContentLoaded", function () {
  const client = new Gadget({
    browserSession: {
      shopId: {{ shop.id }},
    },
  });

  client.getCurrentShop().then((result) => {
    const shopInfoElement = document.getElementById("shop-info");
    shopInfoElement.innerHTML = JSON.stringify(result, null, 2);
  })
})
</script>

<pre id="shop-info"></pre>
```
Then the `shopInfoComponent` should render the payload from the `getCurrentShop` global action with the current shop's data.

</details> 

## PR Checklist

- [ ] Any user facing changes are documented (changelog section below)
- [ ] Any app Problems are added to the problem finders
- [ ] Any app data that needs a redeploy to take effect is listed in the `contentHash` functions
- [ ] Important or complicated production behaviour has traces and logs added

## Changelog

If this PR will affect change user-facing behaviour, add detailed information about the change here to prompt an AI generated changelog entry.

If no changelog is required, write `[no-changelog-required]` in square brackets `[]`.

### DIFF
--- a/packages/api-client-core/spec/GadgetConnection-suite.ts
+++ b/packages/api-client-core/spec/GadgetConnection-suite.ts
@@ -71,6 +71,38 @@ export const GadgetConnectionSharedSuite = (queryExtra = "") => {
   });
 
   describe("authorization", () => {
+    describe("when declaring authentication modes at the top level and under the `authenticationMode` key at the same time", () => {
+      it("should throw an error if same auth modes are declared", () => {
+        expect(
+          () =>
+            new GadgetConnection({
+              endpoint: "https://someapp.gadget.app/api/graphql",
+              authenticationMode: { browserSession: true },
+              browserSession: true,
+            } as any)
+        ).toThrowErrorMatchingInlineSnapshot(
+          `"Declaring authentication modes at the top level and under the \`authenticationMode\` key at the same time is not allowed."`
+        );
+      });
+
+      it("should throw an error if different auth modes are declared", () => {
+        expect(
+          () =>
+            new GadgetConnection({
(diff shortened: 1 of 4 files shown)


infiton (approved):
looks good; mind adding some type tests to make sure we have type errors if more than one option is passed

packages/api-client-core/spec/GadgetConnection-suite.ts
MillanWangGadget, line 1: Consider adding a test in here for when the shopId is passed in properly

MillanWangGadget (approved):
looks reasonable  👍

alanko0511:
> make sure we have type errors if more than one option is passed

I wasn't able to achieve this at the type level, so I didn't do it in this PR. I was able to do that in `GadgetConnectionOptions` but not in `ClientOptions`, the one that we expose to the user.

PR #880, opened 2025-07-29
Update `useSession` to pause before SSR hydration completes

MillanWangGadget:
- **UPDATE**
  - RRv7 apps have been getting bad hydration errors on the default Gadget web app template. This was caused by hooks that used `useSession` such as `useSignOut` being used before the hydration completes
  - This PR prevents those errors by pausing the session fetching until the hydration is complete. 

Prerelease 
```
"@gadgetinc/react": "https://codeload.github.com/gadget-inc/js-clients/tar.gz/@gadgetinc/react-v0.21.4-gitpkg-a041586f",
```

## 🎩 Instructions

<someone> please :tophat::

1. In an SSR app, go into authenticated areas and refresh repeatedly. Observe that there is sometimes a hydration error
2. Put this prerelease into your apps and go into authenticated areas. Refresh repeatedly to observe that there are not more hydration issues

### DIFF
--- a/packages/react/src/auth/useSession.ts
+++ b/packages/react/src/auth/useSession.ts
@@ -7,6 +7,7 @@ import type {
   LimitToKnownKeys,
   Select,
 } from "@gadgetinc/api-client-core";
+import { useEffect, useState } from "react";
 import { useApi } from "../GadgetProvider.js";
 import { useGet } from "../useGet.js";
 import type { OptionsType, ReadOperationOptions } from "../utils.js";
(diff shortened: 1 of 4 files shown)


packages/react/src/auth/useSession.ts
```diff
@@ -69,6 +75,7 @@ export function useSession<
 
     const opts: any = {
       suspense: true,
```
infiton, line 77: does this have anything to do with this suspense?

infiton: ok so after some reading I think the issue is that we have a `<Suspense>` boundary at the top level of the app and what happens on hydration is the `useSession` hook suspends which suspends up to the top level boundary; if before that suspense resolves some other state changes then we will get this hydration error

I think maybe the better solution is to move the suspense boundary closer to where the `useSession` is called

MillanWangGadget:
> can you think of a way to test this?

I'm gonna investigate this. I've been testing this by putting the prerelease into prod apps, and repeatedly refreshing the page 

> this is the one you're fixing right?

Yes. I've only observed that on prod web apps in logged in areas because of `useSignOut` which relies on `useUser -> useSession`

PR #886, opened 2025-08-21
add react 19 support

infiton:
Updates all of our react packages to support react 19; from the upgrade guide we hit:

1. https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument
2. https://react.dev/blog/2024/04/25/react-19-upgrade-guide#better-usereducer-typings

FYI @mitchcail 

## PR Checklist


packages/api-client-core/package.json
```diff
@@ -45,14 +45,14 @@
     "ws": "^8.18.2"
   },
   "devDependencies": {
-    "@types/node": "^16.11.7",
+    "@types/node": "^22.13.14",
```
scott-rc, line 48: If `api-client-core` starts using an API that's only available in node 22+ then apps that upgrade their packages but are still on framework versions less than 1.4 will start to break.

Our oldest supported framework version (1.0) uses node 20. Can we make lower this dependency to v20?

scott-rc (approved):

MillanWangGadget (approved):
- 🍏 Changes look reasonable. I did not see any effective differences in functionality 
- 🍏 cypress and storybook seem to work the same

PR #895, opened 2025-09-26
AutoTable - Auto disable search and error on search props when the model has search disabled

MillanWangGadget:
- UPDATE
  - in FWVersion 1.5, you can disable search ability in a model. 
  - When a search-disabled model is used in AutoTable, the search UI is automatically omitted, and errors are thrown if you pass in search based props like `searchable: true` or `searchValue: "something"`
  - The underlying `useTable` hook has been updated to only return the `search` when the model is searchable and has searchability enabled


airhorns (approved):
🤘

PR #900, opened 2025-10-06
adds the extracted @gadgetinc/core package

infiton:
This is the first PR in the series of deleting `api-client-core`

It introduces a new package `@gadgetinc/core` that contains the utility functions and classes that we use in both gadget monorepo, the generated JS clients and all other packages that consume it

the goal would be for this package to be very slowly changing and to be isolated enough that if an app had multiple versions floating around due to different packages depending on different versions they will be compatible

## PR Checklist


packages/core/src/GadgetRecord.ts
```diff
@@ -0,0 +1,230 @@
...
+export type RecordShape = Record<string, any> | null | undefined | void;
+
+const kFields = Symbol.for("g/fields");
+const kInstantiatedFields = Symbol.for("g/if");
+const kPersistedFields = Symbol.for("g/pf");
+const kFieldKeys = Symbol.for("g/fk");
+const kTouched = Symbol.for("g/t");
+
+/** Represents one record returned from a high level Gadget API call */
+export class GadgetRecord_<Shape extends RecordShape> {
```
airhorns, line 19: Does this sucker have to be in the core package? There's a bunch of stuff I want to do to it to change it such that it'd be great to be able to generate it instead

packages/core/src/InternalModelManager.ts
```diff
@@ -0,0 +1,21 @@
+import { AnyConnection } from "./AnyConnection.js";
```
airhorns, line 1: nit: perhaps this file should be called AnyInternalModelManager.ts

packages/core/src/GadgetRecordList.ts
```diff
@@ -0,0 +1,92 @@
...
+import { GadgetClientError, GadgetOperationError } from "./support.js";
+import type { PaginateOptions } from "./types.js";
+
+type PaginationConfig = {
+  pageInfo: { hasNextPage: boolean; hasPreviousPage: boolean; startCursor: string; endCursor: string };
+  options?: PaginateOptions;
+};
+
+/** Represents a list of objects returned from the API. Facilitates iterating and paginating. */
+export class GadgetRecordList<Shape extends RecordShape> extends Array<GadgetRecord<Shape>> {
```
airhorns, line 16: same must this be included in the package question

airhorns (commented):
THIS IS VERY EXCITING

packages/core/src/GadgetRecord.ts
infiton: I think after this first refactor though we are going to be in an ok state to maybe move it over; there's going to be this new thing once it's all said and done where the generated client gives the hooks the tools it needs to do things like build operations; so I think we could extract out a "here's how to build records"

